PR #17, opened 2026-01-31
feat: settings dashboard, UI fixes, and improved bulk delete

mbradley:
## Summary
- **Settings Dashboard**: Read-only view of relay config, NIP-11 metadata, NIP-86 status
- **Restored UI Improvements** (from feature/zendesk-sync merge):
  - Sticky controls on Reports card (CardHeader shrink-0, flex layout)
  - Consistent scroll-area height handling
  - Better user profile previews and thread context display
- **Improved Bulk Delete by Kind**:
  - Shows clickable kind breakdown of all event types the user has
  - Added Profile Metadata (kind 0) to dropdown
  - Click a kind chip to quickly select it
- **Bug Fixes**:
  - Fix npub input in ban dialog (was causing white screen)
  - Remove legacy environment option (deprecated)
  - Correct relay URLs for staging

## Test plan
- [x] Settings dashboard displays environment info, relay identity, NIP support
- [x] Ban dialog accepts npub input without crashing
- [x] Reports card controls stay sticky when scrolling
- [x] Bulk delete shows actual event kinds for the user
- [x] All tests pass

### DIFF
--- a/worker/src/index.ts
+++ b/worker/src/index.ts
@@ -277,9 +277,11 @@ async function handlePublish(
 
   // If this is a moderation/resolution label (kind 1985), sync to Zendesk
   if (body.kind === 1985 && body.tags) {
+    console.log('[handlePublish] Kind 1985 detected, checking for resolution label');
     const isResolutionLabel = body.tags.some(
       (tag: string[]) => tag[0] === 'L' && tag[1] === 'moderation/resolution'
     );
+    console.log('[handlePublish] isResolutionLabel:', isResolutionLabel);
 
     if (isResolutionLabel) {
       // Extract resolution status from 'l' tag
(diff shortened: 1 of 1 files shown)


rabble (approved):
## Overall

Good PR. Reversible moderation actions, profile linking, sticky action buttons, settings dashboard, environment variable refactoring — all solid improvements.

### Should fix before merge

1. **`profileUrl` can be `undefined`** — The try/catch in `EventCard` (and `PostCard`, `ThreadPost`) returns `undefined` on encode failure, but the `<a href={profileUrl}>` doesn't guard against it. Clicking navigates to the current page instead of being a no-op. Either fallback to `#` or wrap the link in a conditional.

2. **Debug `console.log` in frontend** — `BulkDeleteByKind.tsx` lines 103 and 110 have debug logging that should be removed before merge.

### Nice to have (follow-up OK)

- Unban/restore mutation logic is copy-pasted across 3 files — could extract into shared hooks
- Trust level badges ("Trusted") are based purely on report count, which could be misleading
- `scripts/bulk-delete-videos.mjs` has no authentication and defaults to production relay URL
- `divine.video` profile URL hardcoded in many places — one file uses `DIVINE_PROFILE_URL` constant but the rest don't

PR #47, opened 2026-04-07
feat: route profile URLs to njump.me for non-Divine users

mbradley:
## Summary

- Profile links in moderation UI now route to `njump.me/{npub}` when the user isn't indexed by Funnelcake (i.e., their profile doesn't exist on `relay.divine.video`). Previously all links went to `divine.video/profile/{npub}`, which is blank for non-Divine Nostr users.
- `useAuthor` hook now exposes `isFunnelcakeUser: boolean` tracking whether the Funnelcake REST path succeeded. Zero new network requests -- just reads the result already being fetched.
- Category A call sites updated: `UserIdentifier`, `UserDisplayName`, `ReporterCard`, `ReporterInline`, `ReportedBy`, `UserProfilePreview`. Category B (thread, events list, banned cards) deferred -- no regression, lower priority.

Surfaced by Aleysha's Apr 2 report batch -- moderators were clicking through to empty Divine profile pages for users who only exist on external relays.

## Test plan

- [ ] Deploy to staging, open a report from a non-Divine Nostr user, confirm profile link goes to njump.me
- [ ] Open a report from a known Divine user, confirm profile link still goes to divine.video
- [ ] Verify tooltip on ReporterInline shows "View profile on njump.me" vs "View profile on Divine" appropriately
- [ ] Existing tests: 71 pass, tsc clean


NotThatKindOfDrLiz (changes requested):
I like the direction here. The core `useAuthor` + `getProfileUrl` approach is clean, and routing non-Divine users to `njump.me` solves a real moderator problem.

I do want to request two fixes before merge.

First, `useAuthor` now implicitly depends on the current environment's `apiUrl`, but the query key in `src/hooks/useAuthor.ts` still only keys on `pubkey`:

```ts
queryKey: ['author', pubkey ?? ''],
```

That was fine before, but after this change callers that omit the override now fall back to `config.apiUrl`. If a moderator switches environments, TanStack Query can reuse a cached author result from the previous environment for the same pubkey, including the new `isFunnelcakeUser` flag. That creates exactly the kind of subtle misrouting this PR is trying to fix: profile links can point to the wrong destination for up to the cache lifetime. Please include `apiUrl` in the query key.

Second, the PR introduces new hardcoded production links for recent-content event URLs in `UserProfileCard` and `UserProfilePreview`:

```ts
https://divine.video/${...}
```

This repo's rule is that domains come from config, not application code, and this is more than stylistic: in staging or local, these links will kick moderators out to production. `ThreadContext` already has the same hardcoded pattern, but this PR adds more of it, so I don't think we should expand that further. Please route these through an environment-aware helper instead of hardcoding `divine.video`.

A couple of quick notes on the rest:
- The `ReporterInfo` issue looks fixed on the current head.
- The broader recent-content UI work is useful, but it does make the PR a bit wider than the title suggests.

Requesting changes for the `useAuthor` cache key and the hardcoded production event links. Once those are fixed, I'm comfortable taking another look.

NotThatKindOfDrLiz:
Picked this up and pushed the follow-up fix set on top of the branch (`4871928`).

What changed:
- `useAuthor` is now keyed by `apiUrl`, so environment switches do not reuse stale `isFunnelcakeUser` results from another environment.
- Public event links are now centralized through a helper instead of hardcoding new `divine.video` URLs in the UI. Production still routes to Divine; staging/local fall back to `njump` so moderators do not get kicked to production unexpectedly.
- Removed the two plan-doc files from the PR so the diff stays focused on the feature work.

Local verification:
- `npx eslint src/components/ThreadContext.tsx src/components/UserProfileCard.tsx src/components/UserProfilePreview.tsx src/hooks/useAuthor.ts src/lib/constants.ts`
- `npx vite build`

NotThatKindOfDrLiz (approved):
The follow-up fixes are in. The environment-sensitive caching and public-link routing issues are resolved, the PR diff is tighter, and the frontend checks I ran are clean.

PR #51, opened 2026-04-29
chore: add repo contribution guardrails

NotThatKindOfDrLiz:
## Summary
- add repo-local contributor and agent guardrails in AGENTS.md
- add semantic PR title enforcement plus PR and issue templates
- reinforce redaction of sensitive external names in public repo artifacts

## Motivation
- make repo-local contribution expectations explicit for humans and AI agents
- align this repo with the org standard without duplicating existing internal architecture docs

## Validation
- not run; docs and GitHub metadata changes only

### DIFF
--- a/.github/workflows/semantic_pr.yml
+++ b/.github/workflows/semantic_pr.yml
@@ -1,0 +1,12 @@
+name: Semantic PR
+
+on:
+  pull_request_target:
+    types: [opened, edited, synchronize, reopened]
+
+permissions:
+  pull-requests: read
+
+jobs:
+  semantic-pull-request:
+    uses: VeryGoodOpenSource/very_good_workflows/.github/workflows/semantic_pull_request.yml@v1
--- a/.github/workflows/semantic_pr.yml
+++ b/.github/workflows/semantic_pr.yml
@@ -9,4 +9,4 @@ permissions:
 
 jobs:
   semantic-pull-request:
-    uses: VeryGoodOpenSource/very_good_workflows/.github/workflows/semantic_pull_request.yml@v1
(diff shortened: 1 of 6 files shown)


.github/workflows/semantic_pr.yml
```diff
@@ -0,0 +1,12 @@
...
+on:
+  pull_request_target:
+    types: [opened, edited, synchronize, reopened]
+
+permissions:
+  pull-requests: read
+
+jobs:
+  semantic-pull-request:
+    uses: VeryGoodOpenSource/very_good_workflows/.github/workflows/semantic_pull_request.yml@v1
```
dcadenas: This reusable workflow is pinned to the floating `v1` tag. Combined with the `pull_request_target` trigger above, anything that gets pushed to `v1` in the future runs here against PR events, including PRs from forks. The `permissions: pull-requests: read` block keeps the blast radius small (no write tokens in scope), so this is a hardening suggestion rather than a block. GitHub's Actions security guidance recommends pinning third-party actions and reusable workflows by full commit SHA to defend against tag-movement:

```suggestion
    uses: VeryGoodOpenSource/very_good_workflows/.github/workflows/semantic_pull_request.yml@<commit-sha>  # v1
```

Replace `<commit-sha>` with the SHA that `v1` currently points to; the trailing comment keeps the human-readable version visible.

dcadenas (approved):
🛡️ guardrails look right — one hardening suggestion on the workflow pin inline.

.github/workflows/semantic_pr.yml
NotThatKindOfDrLiz: Pinned the reusable workflow to commit `c2e27845f65775943bc310b3ab89e5c5432b8d4d` and kept the `# v1` note for readability in c69619c.

PR #52, opened 2026-04-30
fix: add blob: to media-src CSP for proxied video previews

mbradley:
## Summary

Adds `blob:` to the `media-src` CSP directive in `index.html`. Proxied restricted media creates blob URLs that `media-src 'self' https:` was blocking. `img-src` and `connect-src` already allow `blob:` -- this aligns `media-src`.

## Motivation

Restricted video content loads through the admin media proxy (`/api/media-proxy/{sha256}`), which fetches from the media server's admin endpoint and creates a blob URL for the `<video>` element. CSP was blocking the blob URL, causing the AlertTriangle fallback instead of the video preview.

## Related Issue

- Relates to moderation media preview workflow

## Validation

- [x] `npm run test`
- [x] `npx tsc --noEmit`
- [x] `npx vite build`
- [x] No worker code changed

## Manual Test Plan / Notes

- [x] Playwright network capture: proxy requests return 200, no CSP console errors for blob URLs
- [x] Deployed to production Pages, opened NS-sexualContent report with age-restricted video -- video preview renders correctly through proxy chain
- [x] Non-restricted media still loads directly (verified on NS-spam report with external video)

## Visuals / API Examples

- [x] No visual change (CSP header only; video rendering was broken before, now works)

## Public Artifact Review

- [x] Titles, descriptions, branch names, and screenshots avoid partner, customer, brand, campaign, or other sensitive external names unless explicitly approved

### DIFF
--- a/index.html
+++ b/index.html
@@ -9,7 +9,7 @@
     <meta property="og:description" content="Manage your Nostr relay with ease using NIP-07 signing" />
     <title>Nostr Relay Manager</title>
     <link rel="manifest" href="/manifest.json" />
-    <meta http-equiv="content-security-policy" content="default-src 'none'; script-src 'self'; style-src 'self' 'unsafe-inline'; frame-src 'self' https:; font-src 'self'; base-uri 'self'; manifest-src 'self'; connect-src 'self' blob: https: wss:; img-src 'self' data: blob: https:; media-src 'self' https:">
+    <meta http-equiv="content-security-policy" content="default-src 'none'; script-src 'self'; style-src 'self' 'unsafe-inline'; frame-src 'self' https:; font-src 'self'; base-uri 'self'; manifest-src 'self'; connect-src 'self' blob: https: wss:; img-src 'self' data: blob: https:; media-src 'self' blob: https:">
   </head>
   <body>
     <div id="root"></div>


NotThatKindOfDrLiz:
Follow-up issue created from this PR review: #54 Investigate and remove stale `worker/public` frontend artifacts if no active deployment path depends on them.

NotThatKindOfDrLiz (changes requested):
The `index.html` CSP change looks correct: adding `blob:` to `media-src` matches how restricted media is loaded through `/api/media-proxy/{sha256}` and then rendered from a blob URL in the media preview code.

I am requesting changes because this PR currently includes unrelated `worker/public` rebuild output in a separate commit (`16df851`). Based on the repo’s current deployment path, the frontend is deployed from `dist` via Cloudflare Pages, while the worker is API-only. That makes the `worker/public` churn out of scope for this fix and it obscures review of the actual shipping change.

Please update this PR to:
1. Drop the `worker/public` rebuild commit so the diff is only the one-line CSP change in `index.html`.
2. Complete the two unchecked manual validations in the PR description:
- reload a report with restricted video content and confirm the preview renders
- verify non-restricted media still loads normally

Follow-up cleanup is tracked in #54, so that debt does not need to block this fix once the PR is narrowed and the manual checks are done.

mbradley:
@NotThatKindOfDrLiz Both items addressed:

1. Dropped the `worker/public` rebuild commit -- PR is now a single commit (the one-line CSP change in `index.html`).
2. Manual validations completed and checked off in the updated description. Verified in production: restricted video renders through proxy chain (NS-sexualContent report with age-restricted Vine import), non-restricted media loads directly (confirmed via network trace + curl).

Ready for re-review.

NotThatKindOfDrLiz (approved):
Rereviewed after the PR was narrowed to the single `index.html` CSP change and the manual validations were completed in production. The out-of-scope `worker/public` churn is gone, the `blob:` addition to `media-src` is the right minimal fix for proxied restricted video previews, and I don't see any remaining blockers.

PR #56, opened 2026-05-03
feat(auto-hide): configurable tiers, settings UI, media proxy fixes

mbradley:
## Summary

- **Auto-hide settings UI** — new `AutoHideSettings` component in the Settings tab for configuring auto-hide tiers (Immediate + Threshold), trusted clients, categories, and thresholds. Uses `getApiHeaders()` for CF Access auth.
- **ReportWatcher tier-aware auto-hide** — `processAutoHide` now reads tier config from DO storage, routes reports to immediate (single trusted-client report) or threshold (N unique reporters per category) processing paths. Threshold counting is category-scoped via exact `reason` match in D1.
- **Media proxy hardening** — fixes a dead-end state where proxy-fetched content that isn't renderable (e.g. non-video blob) caused a permanent black frame instead of falling back to the broken-media icon. Adds `key={displayUrl}` to force remount when proxy URL arrives.
- **QUARANTINE recognized as blocked** — `useMediaStatus` now treats `QUARANTINE` (moderation-service D1 action) as a blocked state, so the action bar correctly shows "Unblock Media" instead of "Block Media & Ban Event" for quarantined items.
- **Pending review filters** — auto-hide actions excluded from resolved targets so auto-hidden items appear in the pending review queue. Category filter cleared when toggling pending review on.
- **BLOSSOM_WEBHOOK_SECRET** binding added to prod and staging wrangler configs (Secrets Store).
- **Blob URL cleanup** on unmount for `InlineMediaPreview` (memory leak fix).

### Worker (Durable Object)
- `ReportWatcher.ts`: tier-aware `processAutoHide`, category-scoped `countUniqueReporters`, config GET/PUT endpoints, `isAlreadyAutoHidden` deduplication
- `ReportWatcher.test.ts`: 163 tests covering immediate/threshold tiers, trusted client gating, config validation, deduplication
- `index.ts`: forward request body to DO for PUT /config

### Frontend
- `AutoHideSettings.tsx`: new settings UI with chip lists for categories/clients, tier config, client-side validation
- `MediaPreview.tsx`: proxy blob recovery, blob URL cleanup, `key={displayUrl}` remount
- `useMediaStatus.ts`: QUARANTINE recognized as blocked
- `Reports.tsx`: auto-hide actions excluded from resolved, category filter cleared on pending review toggle
- `SettingsDashboard.tsx`: mounts `AutoHideSettings`

### Config
- `wrangler.prod.toml`, `wrangler.staging.toml`: BLOSSOM_WEBHOOK_SECRET secrets store binding

## Test plan

- [x] Deploy to staging (test instance), verify auto-hide settings load and save
- [x] Trigger threshold-tier auto-hide with 2 reports from different pubkeys, same category — confirm event is hidden
- [x] Verify single cross-category report does NOT trigger threshold (category scoping)
- [x] Confirm CSAM/immediate tier fires on single trusted-client report
- [ ] Open a report with quarantined media — verify "Unblock Media" button appears (not "Block Media & Ban Event")
- [ ] Verify media proxy shows broken icon (not black frame) when proxy content is unrenderable
- [ ] Verify pending review toggle shows auto-hidden items, not resolved items

### DIFF
--- a/shared/autohide.ts
+++ b/shared/autohide.ts
@@ -1,0 +1,40 @@
+export const AUTO_HIDE_ACTION = {
+  hidden: 'auto_hidden',
+  pending: 'auto_hide_pending',
+  skipped: 'auto_hide_skipped',
+  failed: 'auto_hide_failed',
+} as const;
+
+export const AUTO_HIDE_ACTIONS = [
+  AUTO_HIDE_ACTION.hidden,
+  AUTO_HIDE_ACTION.pending,
+  AUTO_HIDE_ACTION.skipped,
+  AUTO_HIDE_ACTION.failed,
+] as const;
+
+export type AutoHideAction = typeof AUTO_HIDE_ACTIONS[number];
+
+export const AUTO_HIDE_TIER_KINDS = ['immediate', 'threshold'] as const;
+export type AutoHideTierKind = typeof AUTO_HIDE_TIER_KINDS[number];
+
+export interface AutoHideTier {
+  kind: AutoHideTierKind;
+  name: string;
+  categories: string[];
+  threshold: number;
+  requireTrustedClient: boolean;
+}
+
+export interface AutoHideConfig {
+  enabled: boolean;
+  trustedClients: string[];
+  tiers: AutoHideTier[];
+}
+
+export function isImmediateAutoHideTier(tier: AutoHideTier): boolean {
+  return tier.kind === 'immediate';
+}
+
+export function isThresholdAutoHideTier(tier: AutoHideTier): boolean {
+  return tier.kind === 'threshold';
+}
--- a/src/components/AutoHideSettings.tsx
+++ b/src/components/AutoHideSettings.tsx
@@ -1,0 +1,289 @@
+import { useState, useEffect } from "react";
+import { useQuery, useMutation, useQueryClient } from "@tanstack/react-query";
+import { Card, CardContent, CardHeader, CardTitle } from "@/components/ui/card";
+import { Badge } from "@/components/ui/badge";
+import { Switch } from "@/components/ui/switch";
+import { Input } from "@/components/ui/input";
+import { Button } from "@/components/ui/button";
+import { Label } from "@/components/ui/label";
+import { Alert, AlertDescription } from "@/components/ui/alert";
+import { useAppContext } from "@/hooks/useAppContext";
+import { useToast } from "@/hooks/useToast";
+import { Shield, X, Plus, Save, AlertTriangle } from "lucide-react";
+
+interface AutoHideTier {
+  name: string;
+  categories: string[];
+  threshold: number;
+  requireTrustedClient: boolean;
+}
+
+interface AutoHideConfig {
+  enabled: boolean;
+  trustedClients: string[];
+  tiers: AutoHideTier[];
+}
+
+function ChipList({
+  items,
+  onRemove,
+  onAdd,
+  placeholder,
+}: {
+  items: string[];
+  onRemove: (item: string) => void;
+  onAdd: (item: string) => void;
+  placeholder: string;
+}) {
+  const [input, setInput] = useState("");
+
+  const handleAdd = () => {
+    const trimmed = input.trim();
+    if (trimmed && !items.includes(trimmed)) {
+      onAdd(trimmed);
+      setInput("");
+    }
+  };
+
+  return (
+    <div className="space-y-2">
+      <div className="flex flex-wrap gap-1.5">
+        {items.map((item) => (
+          <Badge key={item} variant="secondary" className="gap-1 pr-1">
+            {item}
+            <button
+              type="button"
+              onClick={() => onRemove(item)}
+              className="ml-1 rounded-full hover:bg-muted-foreground/20 p-0.5"
+            >
+              <X className="h-3 w-3" />
+            </button>
+          </Badge>
+        ))}
+      </div>
+      <div className="flex gap-2">
+        <Input
+          value={input}
+          onChange={(e) => setInput(e.target.value)}
+          onKeyDown={(e) => e.key === "Enter" && (e.preventDefault(), handleAdd())}
+          placeholder={placeholder}
+          className="h-8 text-sm"
+        />
+        <Button type="button" size="sm" variant="outline" onClick={handleAdd} className="h-8">
+          <Plus className="h-3 w-3" />
+        </Button>
+      </div>
+    </div>
+  );
+}
+
+function TierConfig({
+  tier,
+  onChange,
+  isImmediate,
+}: {
+  tier: AutoHideTier;
+  onChange: (tier: AutoHideTier) => void;
+  isImmediate: boolean;
+}) {
+  return (
+    <div className="space-y-3 rounded-lg border p-4">
+      <div className="flex items-center justify-between">
+        <h4 className="text-sm font-semibold">{tier.name}</h4>
+        <div className="flex items-center gap-2">
+          <Label htmlFor={`trusted-${tier.name}`} className="text-xs text-muted-foreground">
+            Require trusted client
+          </Label>
+          <Switch
+            id={`trusted-${tier.name}`}
+            checked={tier.requireTrustedClient}
+            onCheckedChange={(checked) =>
+              onChange({ ...tier, requireTrustedClient: checked })
+            }
+          />
+        </div>
+      </div>
+
+      <div className="space-y-1">
+        <Label className="text-xs text-muted-foreground">Categories</Label>
+        <ChipList
+          items={tier.categories}
+          onRemove={(cat) =>
+            onChange({ ...tier, categories: tier.categories.filter((c) => c !== cat) })
+          }
+          onAdd={(cat) => onChange({ ...tier, categories: [...tier.categories, cat] })}
+          placeholder="Add category..."
+        />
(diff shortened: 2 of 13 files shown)


src/components/MediaPreview.tsx
```diff
@@ -319,6 +324,14 @@ export function InlineMediaPreview({
   const [proxyUrls, setProxyUrls] = useState<Map<string, string>>(new Map());
   const apiUrl = useApiUrl();
 
+  useEffect(() => {
+    return () => {
+      proxyUrls.forEach(url => {
+        if (url.startsWith('blob:')) URL.revokeObjectURL(url);
+      });
+    };
+  }, [proxyUrls]);
```
NotThatKindOfDrLiz: This cleanup is still unsafe. Because the effect depends on `proxyUrls`, React runs the previous cleanup on every map change, which revokes blob URLs that may still be mounted in a `<video>`/`` from the previous render. That is a use-after-revoke bug, not just a leak fix. Please make this unmount-only or explicitly revoke only URLs that were removed from the map, and apply the same fix to the existing `MediaPreview` cleanup above as well.

worker/src/ReportWatcher.ts
```diff
@@ -223,6 +248,110 @@ export class ReportWatcher implements DurableObject {
...
+      if (!tier.name || typeof tier.name !== 'string') {
+        return 'Each tier must have a name';
+      }
+      if (!Array.isArray(tier.categories)) {
+        return `Tier "${tier.name}": categories must be an array`;
+      }
+      if (typeof tier.threshold !== 'number' || tier.threshold < 1) {
+        return `Tier "${tier.name}": threshold must be >= 1`;
+      }
+      if (tier.name !== 'Immediate' && tier.threshold < 2) {
```
NotThatKindOfDrLiz: We should not merge the tier model in this shape. Right now "immediate" is identified by display name here, by `threshold <= 1` in `processAutoHide`, and by `tier.name === "Immediate"` in the settings UI. Those are three different contracts for one concept, and `PUT /config` now accepts arbitrary data. Please add an explicit discriminator like `kind: 'immediate' | 'threshold'` and key validation, UI behavior, and runtime behavior off that single field.

src/components/Reports.tsx
```diff
@@ -479,8 +479,11 @@ export function Reports({ relayUrl, selectedReportId }: ReportsProps) {
     }
 
     // Add from moderation decisions (ban_user, delete_event, etc.)
+    // Exclude auto-hide actions — those targets belong in pendingReviewTargets, not resolved
+    const autoHideActions = ['auto_hidden', 'auto_hide_pending', 'auto_hide_skipped', 'auto_hide_failed'];
```
NotThatKindOfDrLiz: This introduces another local list of auto-hide action names in the UI. The worker/report-watcher logic is already defining this vocabulary implicitly, so this will drift over time. Please move these action ids into a shared constant/module and have both sides consume the same source of truth.

src/hooks/useMediaStatus.ts
```diff
@@ -26,7 +26,7 @@ export function useMediaStatus(hashes: string[]) {
           return {
             hash,
             status,
-            isBlocked: status?.action === 'PERMANENT_BAN',
+            isBlocked: status?.action === 'PERMANENT_BAN' || status?.action === 'QUARANTINE',
```
NotThatKindOfDrLiz: This QUARANTINE behavior change is reasonable on its own, but it's unrelated to the auto-hide config/settings feature in this PR. Please peel this into a separate PR together with the matching `ModerationAction` widening in `src/lib/adminApi.ts` so rollout and rollback stay isolated.

worker/wrangler.prod.toml
```diff
@@ -33,6 +33,12 @@ binding = "SERVICE_API_TOKEN"
 store_id = "ef490704b12a4feb91c7feb51229c364"
 secret_name = "SERVICE_API_TOKEN"
 
+# Blossom admin bypass token (proxies blocked media to moderators)
```
NotThatKindOfDrLiz: Same issue here: this secrets-store binding change is orthogonal infra work, not part of the auto-hide feature. Please move the Blossom secret binding changes into their own PR so they can be reviewed, deployed, and reverted independently from the app behavior changes.

NotThatKindOfDrLiz (changes requested):
Requesting changes. The branch is still mixing multiple independent stories, and there are a couple of blockers I don't want us to merge into `main` as tech debt. Please split out the unrelated QUARANTINE and Blossom binding changes (and optionally the media-proxy follow-ups if you want a tighter review surface). In the remaining auto-hide PR, the required fixes are: 1) correct the blob URL cleanup so we don't revoke URLs that are still rendered, 2) replace the current magic-string/threshold heuristics with an explicit tier discriminator so UI, validation, and runtime all key off the same invariant, and 3) centralize the auto-hide action names used by `Reports.tsx` instead of duplicating them locally. I left inline comments on the exact spots.

mbradley (commented):
all five addressed in `9e20544`:

**blob cleanup (MediaPreview.tsx)** — good catch, that was a use-after-revoke. switched to a ref so cleanup only fires on unmount.

**shared constants (Reports.tsx)** — moved auto-hide action names to `src/lib/constants.ts` as `AUTO_HIDE_ACTIONS`. worker has a matching constant with cross-reference comment. full shared package felt like overkill given the separate build targets.

**tier model (ReportWatcher.ts + AutoHideSettings.tsx)** — removed the name-based "Immediate" gate. `threshold === 1` is now the single discriminant for immediate behavior across validation, processing, and UI.

**QUARANTINE + secrets binding** — peeled into their own PRs:
- #62: QUARANTINE type + useMediaStatus
- #63: BLOSSOM_WEBHOOK_SECRET binding

NotThatKindOfDrLiz (commented):
Adding inline notes on the follow-up changes I took over so the PR history reflects the final cleanup work.

shared/autohide.ts
```diff
@@ -0,0 +1,40 @@
...
+export const AUTO_HIDE_ACTIONS = [
+  AUTO_HIDE_ACTION.hidden,
+  AUTO_HIDE_ACTION.pending,
+  AUTO_HIDE_ACTION.skipped,
+  AUTO_HIDE_ACTION.failed,
+] as const;
+
+export type AutoHideAction = typeof AUTO_HIDE_ACTIONS[number];
+
+export const AUTO_HIDE_TIER_KINDS = ['immediate', 'threshold'] as const;
```
NotThatKindOfDrLiz, line 17: Pulled the auto-hide tier and action contract into one shared module so the worker and frontend stop drifting on the same vocabulary. This is the real source of truth the earlier review was asking for, instead of a pair of keep-in-sync comments.

src/components/AutoHideSettings.tsx
```diff
@@ -0,0 +1,280 @@
...
+            onAdd={(c) => setDraft({ ...draft, trustedClients: [...draft.trustedClients, c] })}
+            placeholder="Add client name..."
+          />
+        </div>
+
+        {draft.tiers.map((tier, i) => (
+          <TierConfig
+            key={tier.name}
+            tier={tier}
+            isImmediate={isImmediateAutoHideTier(tier)}
```
NotThatKindOfDrLiz, line 253: Switched the settings UI to derive immediacy from `tier.kind` via the shared helper rather than `threshold === 1`. That keeps the frontend behavior aligned with worker validation and runtime enforcement.

worker/src/ReportWatcher.ts
```diff
@@ -223,6 +243,118 @@ export class ReportWatcher implements DurableObject {
...
+      if (!tier.name || typeof tier.name !== 'string') {
+        return 'Each tier must have a name';
+      }
+      if (!Array.isArray(tier.categories)) {
+        return `Tier "${tier.name}": categories must be an array`;
+      }
+      if (typeof tier.threshold !== 'number' || tier.threshold < 1) {
+        return `Tier "${tier.name}": threshold must be >= 1`;
+      }
+      if (!['immediate', 'threshold'].includes(tier.kind)) {
```
NotThatKindOfDrLiz: The worker validator now enforces the discriminator contract explicitly: only `immediate` and `threshold` are accepted, immediate tiers must stay at exactly 1, and threshold tiers must stay at 2+. This closes the gap that was leaving the branch red before takeover.

worker/src/ReportWatcher.test.ts
```diff
@@ -1026,4 +1022,365 @@ describe('ReportWatcher', () => {
...
+        body: JSON.stringify(badConfig),
+      });
+      const response = await watcher.fetch(request);
+
+      expect(response.status).toBe(400);
+      const body = await response.json() as { error: string };
+      expect(body.error).toContain('threshold');
+    });
+
+    it('should reject immediate tier with threshold other than 1', async () => {
```
NotThatKindOfDrLiz, line 1313: Added regression coverage for the two failure modes that mattered here: an immediate tier with the wrong threshold and an invalid tier kind. That makes the contract executable instead of just documented.

NotThatKindOfDrLiz (approved):
Approved after final rereview.

The remaining blockers are addressed:
- tier behavior now uses an explicit discriminator
- auto-hide action ids are centralized in a shared module
- legacy stored configs are migrated forward on load and persisted back

Checks are green and I do not see remaining merge blockers.

PR #58, opened 2026-05-03
ci: ReportWatcher health check with self-healing

mbradley:
## Summary
- Polls staging + production ReportWatcher every 15 min
- Auto-restarts via `/api/report-watcher/start` if unhealthy, re-checks after 30s
- Alerts `#divine-alerts` via Slack webhook only after 2+ consecutive post-recovery failures
- `continue-on-error: true` prevents workflow failure notifications to repo watchers

Closes divinevideo/support-trust-safety#130 -- Structured diagnostic output distinguishing auth errors (HTTP 401/403), timeouts (curl exit 28), DNS failures, connection refused, and JSON parse errors.

Closes divinevideo/support-trust-safety#131 -- Replaced actions/cache with GitHub Actions repo variables (`RW_FAILS_STAGING`, `RW_FAILS_PRODUCTION`). Simple integers, no cache accumulation, inspectable and resettable from Settings UI.

## Setup (before enabling)
Add three repo secrets (Settings > Secrets and variables > Actions):
- `CF_ACCESS_CLIENT_ID` -- Cloudflare Access service token (same as support-trust-safety)
- `CF_ACCESS_CLIENT_SECRET` -- same
- `SLACK_WEBHOOK_ALERTS` -- Incoming webhook for #divine-alerts (same value as `vmalertmanager-slack-webhook-production` in GCP Secret Manager, ask Sam/Ira)

Repo variables (`RW_FAILS_STAGING`, `RW_FAILS_PRODUCTION`) are created automatically on first run.

Workflow runs harmlessly without secrets (curl fails, Slack step skips with warning, workflow stays green).

## Context
Replaces the version in support-trust-safety. Lives here because ReportWatcher is a Cloudflare Durable Object outside the GKE observability perimeter -- no Prometheus, VMRule, or liveness probe can reach it.

## Test plan
- [ ] Merge and configure secrets
- [ ] Run manually from Actions tab to verify staging/prod checks
- [ ] Verify Slack message format in #divine-alerts
- [ ] Stop ReportWatcher manually, confirm auto-restart works
- [ ] Confirm no email notifications to repo watchers on failure

### DIFF
--- a/.github/workflows/report-watcher-health.yml
+++ b/.github/workflows/report-watcher-health.yml
@@ -1,0 +1,197 @@
+# ReportWatcher Health Check
+#
+# Polls staging and production ReportWatcher Durable Objects every 15 min.
+# If unhealthy, attempts auto-recovery via /start. If recovery fails,
+# alerts #divine-alerts via Slack webhook (same channel as GKE alerting).
+#
+# Lives in relay-manager (not iac-coreconfig) because ReportWatcher is a
+# Cloudflare Durable Object outside the GKE observability perimeter.
+# No Prometheus, no VMRule, no liveness probe can reach it.
+#
+# Setup:
+#   1. Repo secrets (Settings > Secrets and variables > Actions):
+#      - CF_ACCESS_CLIENT_ID    — Cloudflare Access service token
+#      - CF_ACCESS_CLIENT_SECRET — Cloudflare Access service token
+#      - SLACK_WEBHOOK_ALERTS   — Incoming webhook for #divine-alerts
+#        (same value as vmalertmanager-slack-webhook-{env} in GCP Secret Manager)
+#   2. To test manually: Actions tab > "ReportWatcher Health Check" > Run workflow
+
+name: ReportWatcher Health Check
+
+on:
+  schedule:
+    - cron: '*/15 * * * *'
+  workflow_dispatch:
+
+jobs:
+  health-check:
+    runs-on: ubuntu-latest
+    # Never mark the workflow as failed — that emails all repo watchers.
+    # Alerting goes through Slack only.
+    continue-on-error: true
+
+    strategy:
+      matrix:
+        env:
+          - name: staging
+            url: https://api-relay-staging.divine.video
+          - name: production
+            url: https://api-relay-prod.divine.video
+
+    steps:
+      - name: Check health
+        id: check
+        env:
+          CF_ACCESS_CLIENT_ID: ${{ secrets.CF_ACCESS_CLIENT_ID }}
+          CF_ACCESS_CLIENT_SECRET: ${{ secrets.CF_ACCESS_CLIENT_SECRET }}
+          API_URL: ${{ matrix.env.url }}
+        run: |
+          RESPONSE=$(curl -sf --max-time 10 \
+            -H "CF-Access-Client-Id: $CF_ACCESS_CLIENT_ID" \
+            -H "CF-Access-Client-Secret: $CF_ACCESS_CLIENT_SECRET" \
+            "$API_URL/api/report-watcher/status" 2>&1) || {
+            echo "healthy=false" >> "$GITHUB_OUTPUT"
+            echo "payload=HTTP request failed" >> "$GITHUB_OUTPUT"
+            exit 0
+          }
+          CONNECTED=$(echo "$RESPONSE" | jq -r '.status.connected')
+          RUNNING=$(echo "$RESPONSE" | jq -r '.status.running')
+          if [ "$CONNECTED" = "true" ] && [ "$RUNNING" = "true" ]; then
+            echo "healthy=true" >> "$GITHUB_OUTPUT"
+          else
+            echo "healthy=false" >> "$GITHUB_OUTPUT"
+          fi
+          echo "payload=$(echo "$RESPONSE" | jq -c '.status')" >> "$GITHUB_OUTPUT"
+
+      # --- Self-healing: try to restart before alerting ---
+
+      - name: Attempt auto-recovery
+        id: recover
+        if: steps.check.outputs.healthy == 'false'
+        env:
+          CF_ACCESS_CLIENT_ID: ${{ secrets.CF_ACCESS_CLIENT_ID }}
+          CF_ACCESS_CLIENT_SECRET: ${{ secrets.CF_ACCESS_CLIENT_SECRET }}
+          API_URL: ${{ matrix.env.url }}
+        run: |
+          echo "ReportWatcher unhealthy on ${{ matrix.env.name }}, attempting restart..."
+          curl -sf --max-time 15 -X POST \
+            -H "CF-Access-Client-Id: $CF_ACCESS_CLIENT_ID" \
+            -H "CF-Access-Client-Secret: $CF_ACCESS_CLIENT_SECRET" \
+            "$API_URL/api/report-watcher/start" || true
+
+          echo "Waiting 30s for reconnection..."
+          sleep 30
+
+          RESPONSE=$(curl -sf --max-time 10 \
+            -H "CF-Access-Client-Id: $CF_ACCESS_CLIENT_ID" \
+            -H "CF-Access-Client-Secret: $CF_ACCESS_CLIENT_SECRET" \
+            "$API_URL/api/report-watcher/status" 2>&1) || {
+            echo "recovered=false" >> "$GITHUB_OUTPUT"
+            echo "payload=HTTP request failed after restart" >> "$GITHUB_OUTPUT"
+            exit 0
+          }
+          CONNECTED=$(echo "$RESPONSE" | jq -r '.status.connected')
+          RUNNING=$(echo "$RESPONSE" | jq -r '.status.running')
+          if [ "$CONNECTED" = "true" ] && [ "$RUNNING" = "true" ]; then
+            echo "recovered=true" >> "$GITHUB_OUTPUT"
+            echo "ReportWatcher recovered after restart"
+          else
+            echo "recovered=false" >> "$GITHUB_OUTPUT"
+            echo "payload=$(echo "$RESPONSE" | jq -c '.status')" >> "$GITHUB_OUTPUT"
+            echo "ReportWatcher still unhealthy after restart"
+          fi
+
+      # --- Failure counting: only alert on 2+ consecutive post-recovery failures ---
+
+      - name: Restore failure count
+        if: steps.check.outputs.healthy == 'false' && steps.recover.outputs.recovered != 'true'
+        uses: actions/cache/restore@v4
+        with:
+          path: .health-state
+          key: rw-health-${{ matrix.env.name }}-${{ github.run_id }}
+          restore-keys: rw-health-${{ matrix.env.name }}-
+
+      - name: Update failure count
+        id: counts
+        if: always()
+        env:
+          HEALTHY: ${{ steps.check.outputs.healthy }}
+          RECOVERED: ${{ steps.recover.outputs.recovered }}
+          ENV_NAME: ${{ matrix.env.name }}
+        run: |
+          mkdir -p .health-state
+
+          FAILS=$(cat ".health-state/$ENV_NAME-fails" 2>/dev/null || echo 0)
+
+          if [ "$HEALTHY" = "true" ] || [ "$RECOVERED" = "true" ]; then
+            FAILS=0
+          else
+            FAILS=$((FAILS + 1))
+          fi
+
+          echo "$FAILS" > ".health-state/$ENV_NAME-fails"
+          echo "fails=$FAILS" >> "$GITHUB_OUTPUT"
+
+          if [ "$FAILS" -ge 2 ]; then
+            echo "alert=true" >> "$GITHUB_OUTPUT"
+          else
+            echo "alert=false" >> "$GITHUB_OUTPUT"
+          fi
+
+          # Track if we auto-recovered (worth a quiet log, not an alert)
+          if [ "$HEALTHY" = "false" ] && [ "$RECOVERED" = "true" ]; then
+            echo "self_healed=true" >> "$GITHUB_OUTPUT"
+          else
+            echo "self_healed=false" >> "$GITHUB_OUTPUT"
+          fi
+
+      - name: Save failure count
+        if: always()
+        uses: actions/cache/save@v4
+        with:
+          path: .health-state
+          key: rw-health-${{ matrix.env.name }}-${{ github.run_id }}
+
+      # --- Alert to Slack only when auto-recovery failed twice in a row ---
+
+      - name: Alert to Slack
+        if: steps.counts.outputs.alert == 'true'
+        env:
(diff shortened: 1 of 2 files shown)


mbradley:
**Merge / enable note (for reviewers):** happy to have this reviewed and approved, but please leave the merge to me. Merging activates the scheduled workflow immediately, and it needs its repo secrets configured before or at merge, or it runs blind (failing every 15 min with no real monitoring). Those are:

- `CF_ACCESS_CLIENT_ID`, `CF_ACCESS_CLIENT_SECRET` (new here; the sibling minor-review health check probes unauthenticated and does not use these)
- `SLACK_WEBHOOK_ALERTS` (likely already present, since the sibling check uses the same secret)
- `RW_HEALTH_VARS_TOKEN`, a fine-grained PAT with Variables: read/write

I'll set those and merge, then run the test-plan verification.

NotThatKindOfDrLiz (approved):
Takeover fixes are complete at 96c5b16. CF Access 302 responses are now classified as auth failures and skip the restart nudge; the job has a 10-minute timeout; curl preserves transport diagnostics; response-derived output crosses into shell through env; and the workflow documents scheduled-monitor limits. The title now reflects monitoring and alerting rather than claiming self-healing.\n\nVerified with actionlint (only the disclosed SC2129 style warning), npm test (74 files, 757 passed, 1 skipped, production build), explicit frontend typecheck, targeted worker tests (30 passed), and the full worker suite (689 passed). CI is green after rerunning one unrelated worker timing flake.

NotThatKindOfDrLiz:
@mbradley I took over the verified review fixes in commit 96c5b16 and pushed them to your branch. The workflow now distinguishes CF Access 302 auth failures from ReportWatcher outages, skips pointless restart attempts for auth failures, caps each job at 10 minutes, retains curl transport diagnostics, and avoids interpolating response-derived output into shell source. I also corrected the title and documented GitHub's scheduled-workflow limits.\n\nLocal validation and the complete GitHub test workflow are green. Please review the takeover commit; the PR is ready for you to merge when you’re comfortable. @rabble and @divinevideo/reviewers have been requested for the normal human review.

realmeylisdev:
Read through this one carefully — the idea is right and the header comments are unusually good, but I think the state machine has a systematic bias I'd want fixed before it starts paging people. All line references are to `.github/workflows/report-watcher-health.yml`.

The short version: **the monitor's most likely long-lived states are the ones where it stays green and says nothing**, while several paths that *do* page fire for things that aren't ReportWatcher outages.

## It can be silently green while the watcher is down

- **L228 — no PAT means no alerts, and no complaint about it.** When `RW_HEALTH_VARS_TOKEN` is absent the step writes `alert=false`, never sets `persist_broken`, and exits 0, so `Alert to Slack`, `Alert monitor failure` and `Fail the run` (L362) are all skipped. That is the repo's state the moment this merges, and it's indistinguishable from the PAT later being revoked. If prod loses its relay socket for a week, that's ~670 green checkmarks and zero Slack posts. It also makes the monitor-failure copy at L342 unreachable for the case it names first.
- **L241 — the counter read has no retry and can't fail loudly.** `gh variable get ... 2>/dev/null || echo "0"` maps 404, 401, 403, 5xx and network failure alike to `0`, then persists the erased count. The 3x retry went on the write half, but that's the side whose failure *over*-counts; a failed read *resets*. One 502 on the GET during a real outage takes `FAILS` from 1 back to 1, `[ 1 -ge 2 ]` is false, nobody is paged, and nothing in the log or the variable records that the count was lost. Worth retrying the read too, and distinguishing "variable not found" (legitimately 0) from "read failed" (don't reset).
- **L244 — any recovered check zeroes the counter, so flapping never reaches the threshold.** There's no second counter or rate window; `self_healed` only emits a `::notice::` on a green run. Your own header (L9-14) says the common unhealthy state is a dropped socket the DO heals itself — and `HEALTH_CHECK_INTERVAL_MS = 30000` in `worker/src/ReportWatcher.ts` means that heal lands inside your 30s re-check window. A watcher that drops and reconnects on every check records `healthy=false, recovered=true, FAILS=0` on all 96 runs/day/env: permanently green, while kind-1984 reports are missed on every drop. Same hole for a 50% duty cycle — `FAILS` goes 1,0,1,0 forever.
- **L148 — `connected && running` can't see a wedged socket.** The REQ `['REQ', id, {kinds:[1984], limit:0}]` is sent exactly once from the `open` listener (`ReportWatcher.ts:688-694`) and nothing re-sends or verifies it, and `alarm()` (`:1482-1484`) only checks `readyState !== OPEN`. If the relay restarts or drops the REQ without closing the TCP socket, `getStatus()` keeps reporting connected/running while auto-hide is dead. You already capture `lastEventAt` and `eventsProcessed` into `payload` at L153 and never test them — on prod, where reports arrive continuously, a frozen `lastEventAt` is the only thing separating a live watcher from a wedged one.
- **L267 — the persist-failure path is inverted.** While the PAT is broken the read at L241 always yields 0, so `FAILS` can never exceed 1 and the *outage* alert can never fire — the channel only ever hears "effectively unmonitored" and never learns that the same run observed `connected:false`. The converse also bites: a PAT scoped Variables-read-only with the counter already at 1 sets `alert=true` *and* `persist_broken=true` in one run, so one incident produces a "ReportWatcher unhealthy" page immediately followed by a "the monitor is broken, check PAT scopes" page pointing on-call somewhere else.

## It pages for things that aren't outages, and one path changes prod

- **L174 — the nudge restarts a watcher someone deliberately stopped.** `/start` and `/stop` have no UI (the only frontend caller hits `/config`), so a stopped watcher is a legitimate steady state. `handleStop` sets `running = false`, so `/status` reports `running:false, connected:false`, which this job reads as unhealthy and answers with `POST /start`. With `running` false, `handleStart` takes the non-no-op branch (`ReportWatcher.ts:246-267`): `running = true`, `persistState()`, `connect()`, `scheduleHealthCheck()` — restoring the kind-1984 subscription that auto-bans events, with `AUTO_HIDE_ENABLED="true"` in `wrangler.prod.toml`. The re-check then sets `recovered=true`, resets the counter, and logs a `::notice::`, so a moderator who stopped the watcher because auto-hide was misfiring gets it silently restarted within 15 minutes. The "harmless" note at L158 holds only for the already-running branch. Related: `RESTART_CODE` (L177) is captured and echoed but never inspected, and that curl discards stderr and its exit code, so a `/start` that 404s after a route rename reads identically to one that worked — the exact regression class the sibling workflow exists to catch.
- **L216 — `always()` also runs on cancel and timeout.** Per the Actions docs, `always()` returns true even when cancelled, and `timeout-minutes` *is* GitHub cancelling the job; `!cancelled()` is the documented alternative. L290, L325 and L362 carry the same guard. So a 10-minute timeout during a slow `gh` call, or a human clicking Cancel, leaves both outputs empty, L244 treats empty as not-healthy, and `FAILS` increments and persists — secrets stay available through the 5-minute cancellation window and the 3-retry write fits inside it, so the inflated count lands. Two aborts, or one abort plus one genuine transient, page for a runner hiccup.
- **L309 — a CF Access failure is paged as a ReportWatcher outage.** On 302/401/403 the recover step is skipped entirely (L164), yet the copy says "a restart nudge did not bring it back" and L311 says "(counting the post-nudge re-checks)" — neither happened. When the service token expires both legs 302, `RECOVERED` is empty, and 30 minutes later on-call is told prod is down and a restart failed, when the DO is fine. It's worse in the other direction: while the token is broken you can't observe the DO at all, so a real outage starting in that window is indistinguishable. `transport_error` (L113) and `parse_error` (L139) have the same problem — all three mean "could not observe the target", not "the target is down", and you already have a separate monitor-broken alert shape for exactly that.
- **L181 — the 30s re-check is shorter than the DO's own backoff.** `MAX_RECONNECT_DELAY_MS = 60000`, and `/start` on an already-running watcher returns "Already running" without touching `reconnectDelay` or forcing a connect. Once backoff pins at 60s, `recovered=false` is the default outcome rather than a signal: a watcher whose next alarm is at T+150s gets counted as failed at T+122s and reconnects unaided 28 seconds later. That inverts the rule stated at L12-14.
- **L247 — manual runs and re-runs mutate live alerting state.** No `github.event_name == 'schedule'` guard and no dry-run input, so the test procedure at L47 and the "Re-run failed jobs" button both read-modify-write the counter. Clicking Run workflow to look at an in-progress outage can land between two reconnect attempts, read `connected=true`, and push the due page out another 15 minutes; re-running a red run double-counts one blip.

## When it does page, the page is weaker than what you already computed

- **L199 — the recover step's jq calls are unguarded**, unlike L137/L142. Under Actions' default `bash --noprofile --norc -e -o pipefail` a non-JSON body makes jq exit 5 and errexit kills the step before `recovered=` or `payload=` is written (I reproduced this with those exact flags). A 200 carrying an HTML interstitial or a truncated body passes the `!= "200"` guard at L192, and the job goes red on the *first* unhealthy check — defeating the ">=2 consecutive" design at L358-360 — with the failure email saying only "Process completed with exit code 5".
- **L153 — both `parse_error` guards are unreachable for the realistic cases, and `payload` can be empty.** `jq -r '.status.connected'` prints `null` and exits 0 on valid JSON without `.status`, and prints nothing and exits 0 on an empty body; only a hard parse error exits non-zero, which L137 already caught — so the L142 block can never emit its message. A 200 with a zero-length body yields `payload=` empty, and since L295's `||` treats empty as falsy on both sides, the page renders ``*Status:* ``` `` — an alert with no diagnostic content, from the step whose comment says it captures the body for diagnostics. Envelope drift instead pages `*Status:* `null`` while the DO is healthy. L153 also wraps the substitution in `echo`, discarding jq's status, unlike L137/L142.
- **L295 — the coarser diagnosis wins.** `steps.recover.outputs.payload || steps.check.outputs.payload` prefers recover, which runs on every unhealthy case except CF Access — the normal alerting path. So `transport_error: curl exit 6: curl: (6) Could not resolve host: ...` from L101-113 is discarded in favour of `transport_error after restart` (L188), which has no exit code and no message because that curl uses `-s` plus `2>/dev/null` and never captures `$?`. The human gets strictly less than the workflow knew 45 seconds earlier.
- **L290 — no dedup, backoff, or all-clear.** `FAILS` only grows while unhealthy, so `alert == 'true'` holds on every subsequent run: the >=2 threshold suppresses the first page, not the repeats. Prod sitting at `running:true, connected:false` over a weekend is ~380 identical messages plus 380 red runs, in a channel shared with GKE alertmanager, and the practical response is muting the channel. Recovery resets `FAILS` silently, so there's no resolve signal either. Alerting on crossings (2, 4, 8, 16…) plus one recovery post gives the same detection latency with far fewer messages. Neither payload carries the probed URL or a first-failure timestamp — the sibling puts the full `METHOD URL -> code` in its message, and a responder can't copy-paste a curl from this one.
- **L342 — the monitor-failure page names a cause it can't have and omits the ones it does.** `persist_broken` is only reachable below L238, i.e. only when `GH_TOKEN` is non-empty, so "missing" is impossible on that path; meanwhile org PAT-policy revocation and repository-access scope — the causes that actually get you here — aren't mentioned, and the retry loop at L260-266 swallows `gh`'s stderr, so the alert carries no evidence. The responder checks Settings, sees the token present and unexpired, and is out of leads.

## Lifecycle

- **L47 — none of this can be exercised before merge.** `workflow_dispatch` only surfaces a Run workflow button for files already on the default branch, and `schedule` only fires there too — you document that for `schedule` at L20-21, then promise manual testing at L47, and the test plan depends on it. So the first-ever execution is post-merge and unattended, against both environments at once, including the state-changing POST at L174 aimed at production. I'd rewrite the setup section to say the first run *is* the test, and merge with someone watching the first two ticks.
- **L71 — `cancel-in-progress: false` doesn't queue.** Per the docs, any existing *pending* run in the group is cancelled by default; holding more than one pending run needs the `queue:` property, which isn't declared. Queue wait isn't charged against `timeout-minutes`, so under runner contention the :18 run can be cancelled outright by the :33 run — that interval is neither probed nor counted, the "consecutive" semantics break, and the only trace is a grey run. `queue: max` is the documented fix. (I checked separately that `matrix` in job-level `concurrency` is fine, and actionlint accepts it.)
- **L55 — worth asking whether this belongs in the worker at all.** `worker/src/index.ts:792-805` already fetches this exact DO `/status` every 5 minutes on `crons = ["*/5 * * * *"]` and throws away everything but a `console.log`, and the same handler at :827-844 already posts to Slack via `sendDbUnavailableAlert(env.SLACK_WEBHOOK_URL, ...)`. An `if (!status.connected || !status.running)` there is roughly ten lines, runs at 5-minute resolution instead of 15, and removes every availability dependency this file adds — 60-day auto-disable on a public repo, best-effort cron delivery, an expiring PAT, CF Access tokens, the default-branch requirement — plus the whole counter apparatus, since `getStatus()` already exports `connectedAt`, `lastEventAt` and `reconnectAttempts`. Your call, but it's a big difference in moving parts for the same alert.

## Smaller things

- **L5** — "it fails the run so GitHub emails the watchers" isn't how this works: for a scheduled workflow GitHub notifies only the single user who last modified the cron line, subject to their own settings. The sibling states it correctly at L58-61.
- **L61** — the PR body says `permissions: contents: read`; the file says `permissions: {}`. The body also calls the actionlint nit pre-existing — actionlint is clean on `main` and reports SC2129 on this file, so it's introduced here. And the template's Visuals / API Examples section is omitted while the PR adds two Block Kit payloads nobody can see rendered; a screenshot of both messages would help, since three of the findings above are about that copy.
- **L95** — `(#130)` is a cross-repo reference written as a bare number, so here it resolves to a closed, unrelated issue ("Age Review: share enforcement response contract types"). Needs the fully-qualified `owner/repo#130` form. It's misleading in context because the sibling's bare `#174`/`#108` *are* correct local refs.
- **L289** — the two Slack steps (L289, L324) are the same 25-line body, the recover step is a degraded copy of `Check health`, the `self_healed` if/else is written twice (L232-236, L280-284), and the 302/401/403 triple lives in both bash (L124) and YAML (L164) — that last pair is the one that will drift. About 130 lines duplicate the sibling and would fit a `workflow_call` reusable workflow, which would also collapse the 2-job matrix (192 job starts/day where the sibling does more probes in one). Minor: the 10s retry budget is shorter than the ~60s secondary-rate-limit cooldown it names, `gh variable set` rewrites 0 over 0 on every healthy run, and the nudge's `if:` excludes only auth codes, so the POST + 30s sleep + re-check still runs against a host that just returned `http_code=0`.

## Things I checked that are fine

Both probe hostnames are correct and live (302 behind CF Access on staging and prod — no repeat of the #108 host mismatch); `matrix` is legal in job-level `concurrency`; the cron really is every 15 minutes and doesn't collide with the sibling's; the counter arithmetic is safe against empty, negative and non-numeric stored values; and no path leaks `CF_ACCESS_CLIENT_SECRET`, the Slack webhook, or the PAT into a log, annotation, step output or Slack body.

---

I haven't pushed anything — this touches CI, so it stays your call. Happy to put the mechanical half on a branch for you to take or leave (jq guards, retrying the read, `!cancelled()`, `queue: max`, the `#130` ref, and the L5/L342 copy), which would leave you the four design decisions: what "healthy" means, how flapping is accounted for, alert dedup, and whether the nudge should touch prod at all.

PR #62, opened 2026-05-12
fix: recognize QUARANTINE as blocked state in media status

mbradley:
## Summary

- Adds `QUARANTINE` to the `ModerationAction` union type in `adminApi.ts`
- Treats QUARANTINE as blocked in `useMediaStatus` so quarantined items show correct action buttons

Split out from #56 per review feedback — isolated for independent rollout/rollback.

## Test plan

- [x] tsc passes
- [ ] Verify quarantined media shows blocked UI (ban icon, not play button)

### DIFF
--- a/src/components/HiveAIReport.tsx
+++ b/src/components/HiveAIReport.tsx
@@ -1,7 +1,7 @@
 // ABOUTME: Displays Hive AI moderation results from the divine-moderation-service
 // ABOUTME: Shows confidence scores, classification categories, timestamps, and re-check ability
 
-import { getApiHeaders } from "@/lib/adminApi";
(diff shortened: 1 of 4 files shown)


NotThatKindOfDrLiz (approved):
Reviewed the follow-up changes on this branch. The QUARANTINE UI fix is correct, and the request/status action types are now split cleanly so the frontend no longer widens the write API contract. Checks are green.

NotThatKindOfDrLiz (approved):

NotThatKindOfDrLiz (approved):

NotThatKindOfDrLiz (approved):

PR #63, opened 2026-05-12
fix: add BLOSSOM_WEBHOOK_SECRET secrets store binding to worker configs

mbradley:
## Summary

- Adds `BLOSSOM_WEBHOOK_SECRET` secrets store binding to both staging and prod wrangler configs
- Media proxy was returning 500 because the handler code existed but the secret was never wired
- Already hotfixed to staging and prod; this PR aligns the config files

Split out from #56 per review feedback — infra change isolated for independent review.

## Test plan

- [x] Already deployed and verified on both environments
- [ ] Confirm media proxy returns content after next deploy from this config


NotThatKindOfDrLiz (approved):
Reviewed the config fix and follow-up cleanup. The secret binding change is correct, and the adjacent docs/local setup drift is now resolved.

PR #64, opened 2026-05-13
feat(#61): child safety report category labels

DSanich:
## Summary

- Added `CATEGORY_LABELS` mappings for new child-safety categories and aliases so reports render human-readable labels instead of raw strings.
- Added support for both camelCase and kebab-case variants used by clients:
  - `NS-childSafety`, `childSafety`, `child-safety`, `NS-child-safety` -> `Child Safety`
  - `NS-underageUser`, `underageUser`, `underage-user`, `NS-underage-user` -> `Under 16`
- Updated high-priority category handling to include child-safety aliases for moderator-safety treatment.
- Extended constants tests to cover new aliases and high-priority expectations.

## Motivation

- Mobile and web now emit split child-safety report categories with different wire-format aliases.
- Without these mappings, moderator UI shows raw category strings, which reduces readability and slows triage.
- This keeps relay-manager display behavior aligned with the category split introduced in companion mobile/web work.

## Related Issue

- Closes #61

## Validation

- [x] `npm run test`
- [x] `npx tsc --noEmit`
- [x] `npx vite build`
- [x] `cd worker && npx vitest run` if worker code changed (N/A: worker code was not changed)
- [ ] I could not run some validation locally, and I explained why below

## Manual Test Plan / Notes

- Open the reports list and verify child-safety/underage reports display as `Child Safety` / `Under 16` instead of raw keys.
- Verify both camelCase and kebab-case aliases resolve to the same label in the UI.
- Verify child-safety categories are treated as high-priority in report list/detail highlighting.
- Confirm no worker behavior changes are included in this PR.

## Visuals / API Examples

- [ ] No visual change
- [x] Screenshots, sample payloads, or logs attached if needed

## Public Artifact Review

- [x] Titles, descriptions, branch names, and screenshots avoid partner, customer, brand, campaign, or other sensitive external names unless explicitly approved

### DIFF
--- a/src/lib/constants.ts
+++ b/src/lib/constants.ts
@@ -12,6 +12,10 @@ export const CATEGORY_LABELS: Record<string, string> = {
   'NS-spam': 'Spam',
   'NS-harassment': 'Harassment',
   'NS-violence': 'Violence',
+  'NS-childSafety': 'Child Safety',
+  'NS-child-safety': 'Child Safety',
+  'NS-underageUser': 'Under 16',
+  'NS-underage-user': 'Under 16',
   'NS-sexualContent': 'Sexual Content',
   'NS-sexual-content': 'Sexual Content',
   'NS-copyright': 'Copyright',
@@ -44,6 +48,10 @@ export const CATEGORY_LABELS: Record<string, string> = {
   'ai-generated': 'AI Generated',
   'NS-ai-generated': 'AI Generated',
   'violence': 'Violence',
+  'childSafety': 'Child Safety',
+  'child-safety': 'Child Safety',
+  'underageUser': 'Under 16',
+  'underage-user': 'Under 16',
   'sexual-content': 'Sexual Content',
   'sexualContent': 'Sexual Content',
   'false-info': 'Misinformation',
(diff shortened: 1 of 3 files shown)


src/lib/constants.ts
```diff
@@ -57,6 +65,7 @@ export type ResolutionStatus = typeof RESOLUTION_STATUSES[number];
 // Categories where media should be hidden by default for moderator safety
 export const HIGH_PRIORITY_CATEGORIES = [
```
dcadenas, line 71: Include every accepted child-safety and underage alias in the shared high-priority policy, and use that shared policy in `Reports.tsx` instead of a local copy.
Right now `CATEGORY_LABELS` accepts `childSafety`, `child-safety`, `NS-underageUser`, `NS-underage-user`, `underageUser`, and `underage-user`, but `HIGH_PRIORITY_CATEGORIES` only adds `NS-childSafety` and `NS-child-safety`.
A report with one of the missing aliases gets the friendly label while `HIGH_PRIORITY_CATEGORIES.includes(category)` stays false, so moderator-safety handling can miss media hiding, priority sorting, and badges.
Add tests that cover high-priority membership for each accepted child-safety and underage alias.

PR #65, opened 2026-05-15
fix: HLS.js fallback for video playback in media preview

mbradley:
## Summary

- some uploaded videos fail to play in Chrome's native `<video>` decoder. divine-web already handles this with HLS.js fallback to transcoded segments — the admin UI didn't have it.
- adds `HlsVideo` component: tries native MP4 first, falls back to HLS.js (transcoded segments via `VITE_CDN_DOMAIN`), then to the authenticated media proxy for gated content
- distinguishes "content not available" (404/deleted) from "failed to load" (other errors) in the media preview UI
- fixes the `showByDefault` race condition where media would flash-show then collapse as report priority data loaded (tri-state `hasHighPriorityReports`)
- adds `blob:` to `media-src` CSP in worker fallback HTML to match dev CSP
- shows full sha256 hashes in media preview (no truncation)

## Test plan

- [x] open a report with video content at relay.admin.divine.video — videos that previously showed as black rectangles with inactive controls should now play via HLS
- [x] open a report with image content — images still load normally
- [x] open a report with a high-priority category — media should start hidden
- [x] open a report with a non-high-priority category — media should start visible
- [x] verify no CORS errors in browser console on report list or report detail pages
- [x] verify deleted/unavailable media shows "Content not available" (ImageOff icon) instead of generic warning triangle

Verified on staging worker and test frontend deployment.

### DIFF
--- a/src/components/MediaPreview.tsx
+++ b/src/components/MediaPreview.tsx
@@ -5,10 +5,11 @@ import { useState, useMemo, useEffect, useRef } from "react";
 import { Card, CardContent, CardHeader, CardTitle } from "@/components/ui/card";
 import { Button } from "@/components/ui/button";
 import { Badge } from "@/components/ui/badge";
-import { Image, Video, Eye, EyeOff, AlertTriangle, ExternalLink } from "lucide-react";
+import { Image, Video, Eye, EyeOff, AlertTriangle, ExternalLink, ImageOff } from "lucide-react";
 import type { NostrEvent } from "@nostrify/nostrify";
 import { useApiUrl } from "@/hooks/useAdminApi";
 import { getApiHeaders } from "@/lib/adminApi";
+import Hls from "hls.js";
 
 interface MediaItem {
   url: string;
@@ -132,6 +133,61 @@ function extractMediaItems(content: string, tags: string[][]): MediaItem[] {
   return Array.from(items.values());
 }
 
+function getHlsUrl(sha256?: string): string | null {
+  return sha256 ? `https://media.divine.video/${sha256}/hls/master.m3u8` : null;
+}
+
+// Native <video> with HLS.js fallback for C2PA-signed videos that Chrome can't decode natively.
+// Mirrors divine-web's approach: try native MP4 first, fall back to transcoded HLS segments.
+function HlsVideo({
+  src,
+  hlsSrc,
+  onError,
+  ...props
+}: {
+  src: string;
+  hlsSrc: string | null;
+  onError: () => void;
+} & Omit<React.VideoHTMLAttributes<HTMLVideoElement>, 'onError' | 'src'>) {
+  const videoRef = useRef<HTMLVideoElement>(null);
+  const hlsRef = useRef<Hls | null>(null);
+  const triedHlsRef = useRef(false);
+
+  useEffect(() => {
+    return () => { hlsRef.current?.destroy(); };
+  }, []);
+
+  const handleError = () => {
+    const video = videoRef.current;
+    if (hlsSrc && !triedHlsRef.current && video) {
+      triedHlsRef.current = true;
+      if (Hls.isSupported()) {
+        const hls = new Hls();
+        hlsRef.current = hls;
+        hls.loadSource(hlsSrc);
+        hls.attachMedia(video);
+        hls.on(Hls.Events.ERROR, (_, data) => {
+          if (data.fatal) {
+            hls.destroy();
+            hlsRef.current = null;
+            onError();
+          }
+        });
+      } else if (video.canPlayType('application/vnd.apple.mpegurl')) {
+        video.src = hlsSrc;
+      } else {
+        onError();
+      }
+    } else {
+      onError();
+    }
+  };
+
+  return (
+    <video ref={videoRef} src={src} onError={handleError} {...props} />
+  );
+}
+
 export function MediaPreview({
   event,
   content: propContent,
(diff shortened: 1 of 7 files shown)


worker/src/index.test.ts
```diff
@@ -27,6 +27,7 @@ describe('relay manager cors', () => {
     expect(response.headers.get('Access-Control-Allow-Origin')).toBe('https://app.divine.video');
     expect(response.headers.get('Access-Control-Allow-Methods')).toBe('GET, POST, PUT, DELETE, OPTIONS');
     expect(response.headers.get('Access-Control-Allow-Headers')).toBe('Content-Type, Authorization, X-Requested-With, Range, X-Admin-Key, CF-Access-Client-Id, CF-Access-Client-Secret');
+    expect(response.headers.get('Access-Control-Allow-Credentials')).toBe('true');
```
NotThatKindOfDrLiz: This assertion became stale in the second commit. `fix: remove global CORS credentials` removed `Access-Control-Allow-Credentials` from `buildCorsHeaders`, but these two expectations stayed behind, so `cd worker && npx vitest run` fails on the committed PR state. Please drop the leftover credential assertions here and in the matching preview-origin test below so the documented worker suite stays green.

src/components/MediaPreview.tsx
```diff
@@ -5,10 +5,11 @@ import { useState, useMemo, useEffect, useRef } from "react";
 import { Card, CardContent, CardHeader, CardTitle } from "@/components/ui/card";
 import { Button } from "@/components/ui/button";
 import { Badge } from "@/components/ui/badge";
-import { Image, Video, Eye, EyeOff, AlertTriangle, ExternalLink } from "lucide-react";
+import { Image, Video, Eye, EyeOff, AlertTriangle, ExternalLink, ImageOff } from "lucide-react";
 import type { NostrEvent } from "@nostrify/nostrify";
 import { useApiUrl } from "@/hooks/useAdminApi";
 import { getApiHeaders } from "@/lib/adminApi";
```
NotThatKindOfDrLiz, line 11: This eager import pulls `hls.js` into the main app bundle even though it is only used after native video playback fails. I built both revisions locally: `origin/main` produces a ~964 kB minified main chunk (~270 kB gzip), while this branch jumps to ~1491 kB (~433 kB gzip). That is a large regression for a rare fallback path. Please move this to a dynamic `import('hls.js')` inside the fallback branch so the cost is only paid when the fallback is actually needed.

src/components/MediaPreview.tsx
```diff
@@ -141,18 +199,14 @@ export function MediaPreview({
...
   proxyUrlsRef.current = proxyUrls;
   const apiUrl = useApiUrl();
 
-  // Collapse media if showByDefault changes to false (e.g., high-priority reports loaded)
-  // This handles the race condition where reports load after initial render
   useEffect(() => {
-    if (!showByDefault) {
-      setShowMedia(false);
-    }
+    setShowMedia(showByDefault);
```
NotThatKindOfDrLiz: This fixes the initial flash/collapse race, but it also means any later `showByDefault` transition from `false` to `true` will overwrite a moderator's manual toggle. In `ReportDetail`, `showByDefault` now starts false while reports are loading and then flips to true when `hasHighPriorityReports === false`, so this effect will reopen media even if the user explicitly hid it. Please preserve explicit user intent here and only apply the async default without clobbering a manual toggle.

NotThatKindOfDrLiz (changes requested):
Requesting changes before merge. I re-ran the relevant validation locally and found three issues that should be fixed before this lands: the committed PR state leaves the worker test suite broken after the CORS header revert, `hls.js` is being pulled into the main bundle for a rare fallback path, and the new `showByDefault` sync can override a moderator's manual toggle when reports finish loading.

NotThatKindOfDrLiz (approved):
Re-ran validation against commit 0104a9e1990a7ed6b4725d9be400459abda2422d. The earlier regressions called out in review are fixed: hls.js is split into a separate async chunk, the manual media toggle is preserved after report data loads, and the stale worker CORS assertions were removed. Validation run: npm test, npx vite build, and worker Vitest suite all passed locally.

NotThatKindOfDrLiz (changes requested):
## Summary

Three blocking issues prevent merge:

1. **Stale worker test assertions** — incomplete revert leaves broken assertions after header removal
2. **Bundle regression from eager hls.js import** — ~530 kB increase for a rare fallback path
3. **Manual media toggle override** — loading state can clobber moderator's explicit choice to show/hide

See inline comments for details and fix guidance.

NotThatKindOfDrLiz (changes requested):
Three blocking issues prevent merge:

**1. Stale worker test assertions (worker/src/index.test.ts:30, :63)**
Commit 1 added CORS credential assertions; commit 2 removed the header but left the tests. Result: worker suite fails.
→ Remove the stale assertions that expect 'Access-Control-Allow-Credentials: true'

**2. Bundle regression: eager hls.js import (src/components/MediaPreview.tsx:11)**
hls.js is only needed when native playback fails, but it's imported at the top level (~530 kB minified). This violates the 'no new tech debt' bar.
→ Change to lazy import: dynamic import('hls.js') inside the fallback error handler (around line 162)

**3. Manual toggle override (src/components/MediaPreview.tsx:208-209)**
showByDefault sync unconditionally does setShowMedia(showByDefault). When loading completes and hasHighPriorityReports flips from undefined→false, this overwrites an explicit moderator toggle.
→ Only apply showByDefault when showMedia is uninitialized (useRef to track first load), not on every hasHighPriorityReports change

After fixes, run: npm run test && npx tsc --noEmit && cd worker && npx vitest run && npx vite build

PR #67, opened 2026-05-15
feat: age review clock — 15-day under-16 moderation workflow

mbradley:
## Summary

Backend and admin UI for the 15-day age review clock. This is what divine-mobile PR #3531 will call to check moderation status and submit parent contact info.

- **D1 table + migration** for `age_review_cases` with state machine, clock pause/resume, and deadline tracking
- **ReportWatcher hook** auto-creates a case when an `NS-underageUser` report arrives (in-memory Set guard against DO WebSocket handler race conditions)
- **Admin API** for case listing (server-side filtering), detail, and moderator actions (state transitions, age band changes, clock control, resolution notes)
- **Mobile-facing `/v1/` endpoints** behind NIP-98 auth: `GET /v1/account/moderation-status` and `POST /v1/minor-review-cases/:caseId/parent-contact` (matches #3531 contract)
- **Cron deadline checker** with Slack alerts for approaching/expired cases, auto-close on expiry
- **Admin UI tab** with split-pane queue, urgency indicators, moderator action panel

What this provides: case tracking, clock management, status reporting to the mobile app, and moderator workflow. Posting restriction is already possible via relay ban on case denial. Keycast integration (Track C, separate PR) adds account-level suspension, covering login and key delegation beyond just relay access.

## Commits

1. `feat: support X-Admin-Key header for local dev`
2. `docs: add local dev setup script and documentation`
3. `feat: age review clock — 15-day under-16 moderation workflow`

## Test plan

**Already verified locally:**
- [x] 123 worker tests passing (`worker/` vitest), including 23 new age-review tests
- [x] 80 frontend tests passing (root vitest)
- [x] `tsc --noEmit` clean
- [x] `eslint` clean (0 errors across all changed files)
- [x] divine-plugin review-before-commit passed (ID integrity, logging discipline, content security)

**Staging verification (pre-prod):**
- [ ] Deploy worker to staging, run `wrangler d1 migrations apply`
- [ ] Send a test `NS-underageUser` report, verify case auto-created
- [ ] Open admin UI, verify Age Review tab loads with case list and detail
- [ ] Test moderator actions: state transitions, clock pause/resume, clear/deny
- [ ] Call `/v1/account/moderation-status` with NIP-98 signed request, verify response shape matches #3531 contract
- [ ] Set a test case deadline to past, trigger cron, verify auto-close and Slack alert

### DIFF
--- a/src/components/AgeReviewDetail.tsx
+++ b/src/components/AgeReviewDetail.tsx
@@ -1,4 +1,4 @@
-import { useState } from "react";
+import { useState, useEffect } from "react";
 import { useMutation, useQueryClient } from "@tanstack/react-query";
 import { useAdminApi } from "@/hooks/useAdminApi";
 import { Badge } from "@/components/ui/badge";
@@ -85,6 +85,10 @@ export function AgeReviewDetail({ caseData: c }: Props) {
   const [resolutionNote, setResolutionNote] = useState(c.resolution_note ?? "");
   const [copiedField, setCopiedField] = useState<string | null>(null);
 
+  useEffect(() => {
+    setResolutionNote(c.resolution_note ?? "");
+  }, [c.id, c.resolution_note]);
+
   const isTerminal = TERMINAL_STATES.includes(c.state);
   const daysRemaining = getDaysRemaining(c);
 
@@ -147,7 +151,7 @@ export function AgeReviewDetail({ caseData: c }: Props) {
               {copiedField === 'pubkey' ? <Check className="h-3 w-3" /> : <Copy className="h-3 w-3" />}
             </button>
             <a
-              href={`/users/${npub}`}
+              href={`/users/${c.pubkey}`}
               className="text-muted-foreground hover:text-foreground"
               title="View in Users tab"
             >
--- a/src/components/AgeReviewDetail.tsx
+++ b/src/components/AgeReviewDetail.tsx
@@ -1,0 +1,360 @@
+import { useState } from "react";
+import { useMutation, useQueryClient } from "@tanstack/react-query";
+import { useAdminApi } from "@/hooks/useAdminApi";
+import { Badge } from "@/components/ui/badge";
+import { Button } from "@/components/ui/button";
+import { Textarea } from "@/components/ui/textarea";
+import { Separator } from "@/components/ui/separator";
+import { ScrollArea } from "@/components/ui/scroll-area";
+import {
+  Select,
+  SelectContent,
+  SelectItem,
+  SelectTrigger,
+  SelectValue,
+} from "@/components/ui/select";
+import {
+  Clock,
+  Play,
+  Pause,
+  CheckCircle,
+  XCircle,
+  AlertTriangle,
+  Loader2,
+  Copy,
+  Check,
+  ExternalLink,
+} from "lucide-react";
+import { nip19 } from "nostr-tools";
+import {
+  AGE_BANDS,
+  TERMINAL_STATES,
+  type AgeReviewCase,
+  type AgeReviewState,
+  type AgeBand,
+} from "../../shared/age-review";
+
+const STATE_LABELS: Record<AgeReviewState, string> = {
+  open_reported: "Open — Reported",
+  under_moderator_review: "Under Moderator Review",
+  restricted_pending_user_response: "Restricted — Pending User Response",
+  restricted_pending_parental_consent: "Restricted — Pending Parental Consent",
+  restricted_pending_support_email: "Restricted — Pending Support Email",
+  submitted_for_review: "Submitted for Review",
+  needs_follow_up: "Needs Follow-up",
+  cleared: "Cleared",
+  denied_closed: "Denied — Closed",
+};
+
+const BAND_LABELS: Record<AgeBand, string> = {
+  under_13: "Under 13",
+  age_13_15: "Age 13–15",
+  age_16_plus_claimed: "16+ (Claimed)",
+};
+
+function stateColor(state: AgeReviewState): string {
+  if (state === 'cleared') return 'bg-green-100 text-green-800 dark:bg-green-900/30 dark:text-green-400';
+  if (state === 'denied_closed') return 'bg-red-100 text-red-800 dark:bg-red-900/30 dark:text-red-400';
+  if (state.startsWith('restricted_')) return 'bg-orange-100 text-orange-800 dark:bg-orange-900/30 dark:text-orange-400';
+  if (state === 'needs_follow_up') return 'bg-yellow-100 text-yellow-800 dark:bg-yellow-900/30 dark:text-yellow-400';
+  return 'bg-blue-100 text-blue-800 dark:bg-blue-900/30 dark:text-blue-400';
+}
+
+function bandColor(band: AgeBand): string {
+  if (band === 'under_13') return 'bg-red-100 text-red-800 dark:bg-red-900/30 dark:text-red-400';
+  if (band === 'age_13_15') return 'bg-orange-100 text-orange-800 dark:bg-orange-900/30 dark:text-orange-400';
+  return 'bg-yellow-100 text-yellow-800 dark:bg-yellow-900/30 dark:text-yellow-400';
+}
+
+function getDaysRemaining(c: AgeReviewCase): number | null {
+  if (c.clock_paused && c.remaining_days_when_paused != null) {
+    return c.remaining_days_when_paused;
+  }
+  if (!c.deadline_at) return null;
+  const ms = new Date(c.deadline_at).getTime() - Date.now();
+  return ms / (24 * 60 * 60 * 1000);
+}
+
+interface Props {
+  caseData: AgeReviewCase;
+}
+
+export function AgeReviewDetail({ caseData: c }: Props) {
+  const api = useAdminApi();
+  const queryClient = useQueryClient();
+  const [resolutionNote, setResolutionNote] = useState(c.resolution_note ?? "");
+  const [copiedField, setCopiedField] = useState<string | null>(null);
+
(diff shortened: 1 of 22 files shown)


src/components/AgeReviewDetail.tsx
```diff
@@ -0,0 +1,360 @@
...
+            <code className="text-xs bg-muted px-1.5 py-0.5 rounded font-mono break-all">{npub}</code>
+            <button
+              onClick={() => copyToClipboard(c.pubkey, 'pubkey')}
+              className="text-muted-foreground hover:text-foreground"
+              title="Copy hex pubkey"
+            >
+              {copiedField === 'pubkey' ? <Check className="h-3 w-3" /> : <Copy className="h-3 w-3" />}
+            </button>
+            <a
+              href={`/users/${npub}`}
```
NotThatKindOfDrLiz: This link is building `/users/${npub}`, but the selected-user flow in `UserManagement` treats the route param as a hex pubkey and uses it directly for status checks and ban/allow mutations. Following this link can therefore show the wrong status and send an npub into NIP-86 calls. Pass the hex pubkey in the route, or make the Users route normalize npub inputs before any comparisons or mutations.

src/components/AgeReviewDetail.tsx
```diff
@@ -0,0 +1,360 @@
...
+}
+
+interface Props {
+  caseData: AgeReviewCase;
+}
+
+export function AgeReviewDetail({ caseData: c }: Props) {
+  const api = useAdminApi();
+  const queryClient = useQueryClient();
+  const [resolutionNote, setResolutionNote] = useState(c.resolution_note ?? "");
```
NotThatKindOfDrLiz, line 95: This state is initialized from props once and never resynced when `caseData` changes. If a moderator types a note on case A and then selects case B, the textarea can still contain case A text and submit it for case B. This needs a prop-to-state sync or a different state ownership model so notes cannot bleed across cases.

worker/src/age-review.ts
```diff
@@ -0,0 +1,368 @@
...
+  if (TERMINAL_STATES.includes(existing.state as AgeReviewState)) {
+    return json({ success: false, error: 'Cannot modify a closed case' }, 400, corsHeaders);
+  }
+
+  const body = await request.json() as Record<string, unknown>;
+  const updates: string[] = [];
+  const binds: unknown[] = [];
+
+  // State transition
+  if (body.state && typeof body.state === 'string') {
```
NotThatKindOfDrLiz, line 97: The PR description calls this a state machine, but the server currently accepts any enum value here with no transition validation. That makes the API the source of new tech debt: any admin client or script can move a case into impossible states, and the mobile contract depends on these states being coherent. Please enforce allowed transitions server-side.

worker/src/age-review.ts
```diff
@@ -0,0 +1,368 @@
...
+    SELECT * FROM age_review_cases
+    WHERE state NOT IN (${TERMINAL_STATES.map(() => '?').join(',')})
+      AND clock_paused = 0
+      AND deadline_at IS NOT NULL
+      AND deadline_at < datetime('now', '+2 days')
+      AND deadline_at > datetime('now')
+    ORDER BY deadline_at ASC
+  `).bind(...TERMINAL_STATES).all<AgeReviewCase>();
+
+  if (approaching.results.length > 0 && env.SLACK_WEBHOOK_URL) {
```
NotThatKindOfDrLiz, line 333: This will send the same approaching-deadline Slack alert on every cron run until the case expires or changes state. Without a dedupe marker like `last_alerted_at` or a separate alert log, this will create repeated operational noise for the same case.

NotThatKindOfDrLiz (changes requested):
Requesting changes before merge. The current implementation has a few correctness issues that can break moderator workflows or create avoidable operational noise:

1. The Age Review detail deep link passes an npub into the Users route, but the selected-user workflow expects a hex pubkey and uses it directly in status checks and moderation actions.
2. The resolution note textarea does not resync when switching between cases, so a note typed on one case can be submitted for another.
3. The admin update endpoint accepts arbitrary enum-to-enum state changes even though this feature is described as a state machine; transition rules need to be enforced server-side.
4. Deadline alerts have no dedupe, so approaching cases will page Slack on every cron run until expiry.

src/components/AgeReviewDetail.tsx
mbradley: Fixed — deep link now passes hex pubkey (`c.pubkey`) instead of npub.

src/components/AgeReviewDetail.tsx
mbradley: Fixed — added `useEffect` that resyncs `resolutionNote` from props when `c.id` or `c.resolution_note` changes. The second dep handles same-case server refreshes (e.g., after TanStack Query invalidation).

worker/src/age-review.ts
mbradley: Fixed — added a `VALID_TRANSITIONS` map in the shared module (`shared/age-review.ts`) and server-side enforcement in `handleUpdateAgeReviewCase`. Invalid transitions return 400 with a descriptive error (`Cannot transition from 'X' to 'Y'`). New test covers the rejection path.

The map encodes the reachable transitions from the UI and mobile endpoints. Terminal states have empty arrays (already blocked by the existing closed-case check above, but the map is explicit).

worker/src/age-review.ts
mbradley: Fixed — added `last_alerted_at` column (migration 0007) and a 12-hour dedupe filter in the approaching-deadline query. The column is stamped only after a successful Slack delivery (`sendSlackAlert` now returns `boolean`), so a failed webhook doesn't burn the dedupe window. Two new tests cover both paths.

Expired-case alerts don't need dedupe since the case transitions to `denied_closed` on the same cron run.

worker/src/age-review.ts
```diff
@@ -0,0 +1,387 @@
...
+    const deadline = activeCase.deadline_at ? new Date(activeCase.deadline_at) : null;
+    const remainingDays = deadline ? (deadline.getTime() - now.getTime()) / (24 * 60 * 60 * 1000) : DEADLINE_DAYS;
+
+    await env.DB.prepare(`
+      UPDATE age_review_cases
+      SET parent_contact_email = ?,
+          clock_paused = 1,
+          clock_paused_at = ?,
+          remaining_days_when_paused = ?,
+          state = CASE
```
NotThatKindOfDrLiz, line 301: This endpoint still bypasses `VALID_TRANSITIONS`. A client can hit `POST /v1/minor-review-cases/:id/parent-contact` while the case is still `open_reported`, `under_moderator_review`, or `needs_follow_up`, and this raw SQL will jump it straight into `restricted_pending_parental_consent` / `restricted_pending_support_email` without ever passing through the server-side state-machine guard added above. Please route this through the same transition validation and add a regression test for the mobile path.

worker/src/ReportWatcher.ts
```diff
@@ -865,6 +883,64 @@ export class ReportWatcher implements DurableObject {
...
+    if (existing) {
+      console.log(`[ReportWatcher] Active age review case ${existing.id} already exists for ${reportedPubkey.slice(0, 8)}..., skipping`);
+      return;
+    }
+
+    const caseId = crypto.randomUUID();
+    const band = 'age_13_15' as const;
+    const deadline = new Date(Date.now() + DEADLINE_DAYS * 24 * 60 * 60 * 1000).toISOString();
+
+    await this.env.DB.prepare(`
```
NotThatKindOfDrLiz, line 918: This creates an `open_reported` case from a single `NS-underageUser` pubkey report with no trusted-client or moderator-review gate. Combined with `handleGetModerationStatus()` returning `restrictedMinorReview` for every non-terminal case, one unsolicited report is enough to surface a restricted-state payload to mobile before a human looks at it. If mobile actually enforces on that status, this is a real abuse/regression risk; if it does not, that contract needs to be explicit and covered by tests before merge.

NotThatKindOfDrLiz (changes requested):
Requesting changes before merge. Two blocking issues remain: the mobile parent-contact path still bypasses the shared age-review state machine, and the current report-to-restriction flow appears to let a single unsolicited underage report surface `restrictedMinorReview` before moderator review. I don't think this is safe to merge until both are resolved or the mobile contract is made explicit and tested.

worker/src/age-review.ts
mbradley: Fixed — `handleParentContact` now validates the transition against `VALID_TRANSITIONS` before modifying the case. If the current state doesn't allow advancing to `restricted_pending_parental_consent` / `restricted_pending_support_email`, it returns 400. This means the endpoint only works from `restricted_pending_user_response` (for 13–15) as intended — calling it from `open_reported`, `under_moderator_review`, or `needs_follow_up` is now rejected. New test covers the `open_reported` rejection path.

worker/src/ReportWatcher.ts
mbradley: Valid concern. Fixed — `handleGetModerationStatus` now only returns `restrictedMinorReview` for moderator-reviewed states (`under_moderator_review`, `restricted_pending_*`, `submitted_for_review`, `needs_follow_up`). Cases in `open_reported` return `active` to mobile, so a single unsolicited report can't restrict the user before a human confirms. New test verifies `open_reported` returns `active`.

The trusted-client gate on case creation is a separate question — right now any `NS-underageUser` report creates a case. That's acceptable with this fix since `open_reported` is now inert from mobile's perspective, but if we want to gate creation too (e.g., require trusted client tag like auto-hide does), that can be a follow-up.

NotThatKindOfDrLiz (approved):
Reviewed the updated head including the follow-up under-13 restriction fix. The remaining workflow gap is closed, the regression test is in place, and the validation checks passed on the updated branch.

PR #70, opened 2026-05-18
feat: Zendesk integration for age review parent contact

mbradley:
## Summary

Closes #68. Fast follow on PR #67 (age review clock).

- **Outbound**: Auto-creates Zendesk ticket (parent as requester) when parent email is submitted via mobile endpoint. Ticket ID stored on the case for bidirectional linking.
- **Inbound**: New webhook endpoint `POST /api/zendesk/age-review-reply` -- detects parent replies and advances case to `submitted_for_review`.
- **Resolution sync**: When moderator clears/denies a case with a linked ticket, adds internal note and auto-solves the Zendesk ticket.
- All Zendesk interactions are non-critical side effects (fire-and-forget with logging).
- **ReportWatcher log cleanup**: Expands truncated Nostr IDs (pubkeys, event IDs) to full 64-char hex in log lines. Truncated IDs are useless for debugging -- caught by our own code review as a violation of the project's "do not truncate Nostr IDs" guardrail (CLAUDE.md). Included here since the age-review case creation logs were already being touched.

### Zendesk-side config needed after merge

- Trigger: on tickets tagged `age-review` where requester replied, fire webhook to `POST /api/zendesk/age-review-reply` with `{ "ticket_id": {{ticket.id}} }`
- **Secret**: `ZENDESK_AGE_REVIEW_WEBHOOK_SECRET` -- separate from the existing `ZENDESK_WEBHOOK_SECRET`. Must be configured in the Cloudflare Secrets Store and matched in the Zendesk webhook trigger.

### Migration

`0008_age_review_zendesk.sql` -- adds `zendesk_ticket_id` column to `age_review_cases`. `ensureSchema()` also self-heals existing tables via `ALTER TABLE ADD COLUMN`.

## Test plan

- [x] 7 new tests: ticket creation (success, failure, missing creds), webhook handler (transition, no match, non-pending state, missing ticket_id)
- [x] All 32 age-review tests pass
- [x] `tsc --noEmit` clean (frontend + worker, excluding pre-existing worker issues)
- [ ] Deploy to staging and verify ticket creation with test Zendesk instance
- [ ] Configure Zendesk trigger and verify inbound webhook flow

### DIFF
--- a/worker/src/ReportWatcher.ts
+++ b/worker/src/ReportWatcher.ts
@@ -609,10 +609,10 @@ export class ReportWatcher implements DurableObject {
 
     console.log(`[ReportWatcher] Report received:`, {
       reportId: event.id,
-      reporter: event.pubkey.slice(0, 8) + '...',
+      reporter: event.pubkey,
       category,
       targetType,
-      targetId: targetId.slice(0, 8) + '...',
+      targetId,
       content: event.content.slice(0, 50) + (event.content.length > 50 ? '...' : ''),
     });
 
@@ -682,12 +682,12 @@ export class ReportWatcher implements DurableObject {
     }
 
     if (await this.hasHumanResolution(targetEventId)) {
-      console.log(`[ReportWatcher] Event ${targetEventId.slice(0, 8)}... has human resolution, skipping auto-hide`);
+      console.log(`[ReportWatcher] Event ${targetEventId} has human resolution, skipping auto-hide`);
       return;
     }
 
     if (await this.isAlreadyAutoHidden(targetEventId)) {
-      console.log(`[ReportWatcher] Event ${targetEventId.slice(0, 8)}... already auto-hidden, skipping`);
+      console.log(`[ReportWatcher] Event ${targetEventId} already auto-hidden, skipping`);
       return;
     }
 
@@ -723,10 +723,10 @@ export class ReportWatcher implements DurableObject {
     const count = await this.countUniqueReporters(targetEventId, category);
 
     if (count >= tier.threshold) {
-      console.log(`[ReportWatcher] Threshold met for ${targetEventId.slice(0, 8)}... (${count}/${tier.threshold})`);
+      console.log(`[ReportWatcher] Threshold met for ${targetEventId} (${count}/${tier.threshold})`);
       await this.executeAutoHide(event, category, targetEventId, tier.name);
     } else {
-      console.log(`[ReportWatcher] Below threshold for ${targetEventId.slice(0, 8)}... (${count}/${tier.threshold})`);
+      console.log(`[ReportWatcher] Below threshold for ${targetEventId} (${count}/${tier.threshold})`);
     }
   }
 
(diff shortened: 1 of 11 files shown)


worker/src/db.ts
```diff
@@ -68,6 +68,7 @@ export async function ensureSchema(db: D1Database): Promise<void> {
       moderator_pubkey TEXT,
       resolution_note TEXT,
       last_alerted_at TEXT,
+      zendesk_ticket_id INTEGER,
```
NotThatKindOfDrLiz, line 71: CREATE TABLE IF NOT EXISTS only fixes fresh DBs. Existing staging/prod tables created before 0008 will not pick up this new column unless the migration is applied in perfect lockstep. Please mirror the existing reporter_pubkey pattern here with a best-effort ALTER TABLE age_review_cases ADD COLUMN zendesk_ticket_id INTEGER inside ensureSchema() so the webhook lookup and ticket-link write self-heal on older databases too.

worker/src/index.ts
```diff
@@ -51,6 +52,7 @@ interface Env {
   ZENDESK_PREAUTH_SECRET?: string;  // HMAC secret for pre-auth token generation/verification
   ZENDESK_WEBHOOK_SECRET?: string;  // For /api/zendesk/webhook
   ZENDESK_PARSE_REPORT_SECRET?: string;  // For /api/zendesk/parse-report
+  ZENDESK_AGE_REVIEW_WEBHOOK_SECRET?: string;  // For /api/zendesk/age-review-reply
```
NotThatKindOfDrLiz, line 55: This new secret name is correct in code, but the rollout/docs need to match it. The current PR description still says to reuse ZENDESK_WEBHOOK_SECRET; if someone configures the trigger from that text, this endpoint will just 401. Please update the PR/deploy docs to call out ZENDESK_AGE_REVIEW_WEBHOOK_SECRET explicitly before merge.

worker/src/ReportWatcher.ts
```diff
@@ -609,10 +609,10 @@ export class ReportWatcher implements DurableObject {
 
     console.log(`[ReportWatcher] Report received:`, {
       reportId: event.id,
-      reporter: event.pubkey.slice(0, 8) + '...',
+      reporter: event.pubkey,
```
NotThatKindOfDrLiz, line 612: This log expansion is unrelated to the age-review Zendesk change. I’m not objecting to full pubkey/event-id logging here on its own, but it’s scope creep in a PR that otherwise stays tightly focused. Please either split this into a separate change or call it out explicitly in the PR description so it gets conscious review.

NotThatKindOfDrLiz (changes requested):
Requesting changes before merge. The blocking issue is schema safety on existing D1 databases: this adds the new column to fresh schemas and ships a migration, but ensureSchema() still does not self-heal existing age_review_cases tables, so deploy ordering can break parent-contact writes and the inbound webhook lookup. I left inline comments on the specific spots. I also called out the webhook secret naming/docs mismatch and the unrelated ReportWatcher log-scope expansion so those are reviewed intentionally.

NotThatKindOfDrLiz (approved):
Re-reviewed at b226a02. The original blocking schema-safety issue is fixed, and the remaining closeout gap I called out is now covered by worker entrypoint tests for /api/zendesk/age-review-reply (valid signed request and invalid signature path). Validation on the updated branch is clean enough to merge: root test command passed, worker vitest passed (142/142), frontend tsc passed, and the existing unrelated worker tsc failures remain outside this PR.

PR #72, opened 2026-05-18
fix: resolve SERVICE_API_TOKEN from Secrets Store before use

mbradley:
## Summary

- `SERVICE_API_TOKEN` changed from a plain string secret to a Secrets Store binding, but code calling `.get()` clobbered, so moderation-service received `Bearer [object Object]` and returned 401
- Applies the same `typeof`/`.get()` resolution pattern used by `NOSTR_NSEC`, `CF_ACCESS_CLIENT_ID`, and `BLOSSOM_WEBHOOK_SECRET`
- Fixes both call sites: `handleModerateMedia` and `notifyModerationService`

## Test plan

- [ ] Deploy to staging, attempt age-restrict action from admin UI (currently 401)
- [ ] Verify DM notifications still fire on ban actions

### DIFF
--- a/worker/src/index.test.ts
+++ b/worker/src/index.test.ts
@@ -1,5 +1,5 @@
-import { describe, expect, it } from 'vitest';
-import worker from './index';
+import { describe, expect, it, vi } from 'vitest';
+import worker, { notifyModerationService } from './index';
 
 const env = {
   ALLOWED_ORIGINS: 'https://app.divine.video,https://*.openvine-app.pages.dev',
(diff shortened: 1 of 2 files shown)


NotThatKindOfDrLiz (approved):
Reviewed the follow-up cleanup on top of the hotfix. The branch now fixes the Secrets Store regression, centralizes secret resolution in the touched auth paths, keeps the non-critical DM failure behavior intact, and remains covered by the worker test suite.

PR #75, opened 2026-05-20
feat: enrich age review UI with profile data

mbradley:
## Summary

- Replace raw hex pubkeys in Age Review list and detail views with `UserIdentifier` / `UserDisplayName` components
- Moderators now see display names, NIP-05, and avatars for flagged users and reporters
- Removes ~40 lines of duplicated clipboard/npub logic from `AgeReviewDetail`
- Adds `useAuthor` mock to `AgeReviewDetail` tests

Closes #73


NotThatKindOfDrLiz (approved):
Final pass complete on c87d918. The follow-up removes the list-row clipboard regression, keeps the moderator hex-copy workflow explicit, adds regression coverage for both behaviors, and leaves the branch green on GitHub Actions.

PR #76, opened 2026-05-20
feat: create Zendesk ticket on age review restriction

mbradley:
## Summary

- Create an internal Zendesk ticket when a moderator restricts an account (state -> `restricted_*`), giving T&S visibility before any parent email arrives
- When a parent email comes in later, update the existing ticket with the parent as requester and send the outreach email (no duplicate tickets)
- Set `ZENDESK_FIELD_AGE_REVIEW_DEADLINE` date custom field on all age review tickets (dormant until field is created in Zendesk admin and ID is set in wrangler config)
- Add `ZENDESK_FIELD_CATEGORY` / `ZENDESK_FIELD_ISSUE` to parent-contact tickets for solve consistency
- Add `ZENDESK_FIELD_AGE_REVIEW_DEADLINE` to top-level `Env` interface

All Zendesk operations are non-critical side effects wrapped in try/catch.

Closes #74

### DIFF
--- a/worker/src/age-review.test.ts
+++ b/worker/src/age-review.test.ts
@@ -714,7 +714,7 @@ describe('handleParentContact Zendesk integration', () => {
     vi.unstubAllGlobals();
   });
 
-  it('skips ticket creation when case already has a zendesk_ticket_id', async () => {
+  it('updates existing Zendesk ticket when case already has a zendesk_ticket_id', async () => {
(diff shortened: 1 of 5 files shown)


NotThatKindOfDrLiz (approved):
Reviewed the updated PR at 0e4f666. The stale Zendesk ticket response bug is fixed, the new restricted-state ticket path is now covered, and the added Zendesk duplication has been consolidated. Checks are green.

PR #80, opened 2026-05-21
feat: wire age review actions to Keycast account suspension (#79)

### LINKED ISSUE #79: task: wire age review actions to Keycast account suspension
## Context

Keycast PR #235 (account suspension) merged May 20. Keycast now exposes `GET/PUT /api/admin/users/:pubkey/status` with service-token auth. Status values: `active`, `suspended`, `banned`. Suspended users cannot sign events or decrypt DMs via the signer.

Relay-manager's age review system currently restricts accounts by age-restricting media (Blossom) and optionally banning the pubkey (Funnelcake). This is effective but imprecise: media restriction hides existing content, but the user can still post new events via any client that doesn't use Keycast signing. With Keycast suspension, the restriction is authoritative.

## What

When a moderator transitions an age review case to a restricted state (`restricted_pending_user_response`, `restricted_pending_support_email`):
- Call `PUT /api/admin/users/:pubkey/status` with `{ status: "suspended", suspended_reason: "age_review" }`

When a moderator clears the case (`clear`):
- Call `PUT /api/admin/users/:pubkey/status` with `{ status: "active" }`

When a case expires (cron auto-close with deny outcome):
- Keep suspended (no unsuspend call)

## Prerequisites

- `KEYCAST_SERVICE_TOKEN` secret set on relay-manager staging + production
- `KEYCAST_URL` env var (probably `https://login.divine.video`)

## Acceptance criteria

- Restricting an account in age review suspends the user in Keycast
- Clearing a case unsuspends the user in Keycast
- Keycast call failure does not block the age review state transition (graceful degradation -- log the error, continue)
- Integration test covering the suspend/unsuspend calls

## Related

- Keycast PR divinevideo/keycast#235 (merged)
- Relay-manager PR #78 (deletion alignment, in review)
- Under-16 coordination doc: `docs/moderation/under-16-system-coordination.md`

mbradley:
## Summary

- Adds `keycast-client.ts` module with `suspendUser`, `unsuspendUser`, `banUser` functions calling Keycast's `PUT /admin/users/:pubkey/status` endpoint
- Wires into age review state transitions: suspend on restrict, unsuspend on clear, ban on deny
- Wires into cron auto-close: ban expired cases
- Graceful degradation: missing config or Keycast failure never blocks a state transition
- Adds `KEYCAST_URL` to staging and production wrangler configs (`KEYCAST_SERVICE_TOKEN` added separately via `wrangler secret`)

Closes #79

## Prerequisites

- Keycast PR divinevideo/keycast#235 (merged May 20)
- `KEYCAST_SERVICE_TOKEN` must be set on staging/production via `wrangler secret put` before Keycast calls fire (calls silently return "not configured" until then)

## Test plan

- [ ] 15 new tests (10 unit, 5 integration), 160 total passing
- [ ] Verify on staging: restrict an age review case, confirm Keycast status changes to suspended
- [ ] Verify on staging: clear a case, confirm Keycast status returns to active
- [ ] Verify on staging: deny a case, confirm Keycast status changes to banned
- [ ] Verify graceful degradation: deploy without `KEYCAST_SERVICE_TOKEN` set, confirm age review transitions still work


NotThatKindOfDrLiz (approved):
Final pass looks clean. The unsafe test wrangler file is removed, the Keycast/account-status transitions are now gated correctly, and the new regression coverage closes the previously unsafe clear-path behavior. Local validation and GitHub checks are green on aa989a1.

worker/src/age-review.ts
```diff
@@ -188,8 +190,18 @@ export async function handleUpdateAgeReviewCase(
   let updated = await env.DB.prepare('SELECT * FROM age_review_cases WHERE id = ?')
     .bind(caseId).first<AgeReviewCase>();
 
+  const requestedState = typeof body.state === 'string'
+    ? body.state as AgeReviewState
+    : undefined;
+  const enteredRestrictedState = requestedState !== undefined
+    && isAccountRestrictedAgeReviewState(requestedState)
+    && !isAccountRestrictedAgeReviewState(existing.state);
+  const clearedRestrictedState = requestedState === 'cleared'
```
dcadenas: Make the clear transition unsuspend accounts that already passed through a restricted age-review state, not only accounts whose immediately previous state is restricted.
A normal path is `restricted_pending_*` -> `submitted_for_review` or `needs_follow_up` -> `cleared`, and the middle transition intentionally leaves Keycast suspended.
With the current `clearedRestrictedState` check, the final clear skips `unsuspendUser` and leaves the account suspended after the case is cleared.
Add a regression test for `restricted_pending_*` -> `submitted_for_review` -> `cleared`, and cover `needs_follow_up` if that path can also follow a suspension.

dcadenas (changes requested):
The Keycast account state must be fixed before merge. Clearing an age review case can currently leave the account suspended after the case moves through review or follow-up states.

worker/src/age-review.ts
mbradley: Fixed in b4b9f3e. Replaced `clearedRestrictedState` (which required the immediately prior state to be restricted) with an unconditional check on `requestedState === 'cleared'`. Both `unsuspendUser` and `un-age-restrict-all` are idempotent, so they fire on every clear regardless of prior state.

Added regression tests for `restricted_pending_*` → `submitted_for_review` → `cleared` and `needs_follow_up` → `cleared`.

dcadenas (approved):
✅ Keycast age actions land cleanly.

PR #82, opened 2026-05-21
fix: dedup report categories, surface Keycast failures, audit partial deletes

mbradley:
## Summary

- **Dedup getReportCategory and HIGH_PRIORITY_CATEGORIES** (closes #81): the local copy in Reports.tsx was missing `NS-underageUser` and `NS-childSafety` (6 entries vs 8 in constants), so child-safety reports weren't getting priority treatment in the queue list. Now imports from `@/lib/constants`.

- **Surface failed Keycast enforcement in age review** (pre-existing): when a moderator restricts/clears/denies an age review case, the worker calls Keycast to suspend/unsuspend/ban the account and returns `keycastUpdated` in the response. The UI was ignoring this flag — showing success even when account enforcement failed silently. Now shows a destructive toast so the moderator knows to retry or escalate.

- **Audit partial deletes in combined event+media deletion** (pre-existing): the "Delete Event & Media" action ran four sequential steps with a single audit log at the end. If media deletion failed after the event was already destroyed, the error toast hid what succeeded and the audit trail was incomplete. Now logs each irreversible side effect as it completes and surfaces partial success to the moderator.

## Test plan

- [x] `npx tsc --noEmit` clean
- [x] All 99 tests pass (10 suites including EventActions and AgeReviewDetail)
- [ ] Staging: verify child-safety reports appear with priority badge in queue
- [ ] Staging: trigger age review restrict with Keycast unconfigured, verify warning toast appears
- [ ] Staging: verify combined delete surfaces partial success if media step fails

### DIFF
--- a/src/components/AgeReviewDetail.tsx
+++ b/src/components/AgeReviewDetail.tsx
@@ -1,6 +1,7 @@
-import { useState, useEffect } from "react";
+import { useState, useEffect, useRef } from "react";
 import { useQuery, useMutation, useQueryClient } from "@tanstack/react-query";
 import { useAdminApi } from "@/hooks/useAdminApi";
+import { useToast } from "@/hooks/useToast";
 import { DeleteConfirmDialog } from "@/components/DeleteConfirmDialog";
 import { UserActions } from "@/components/UserActions";
 import { Badge } from "@/components/ui/badge";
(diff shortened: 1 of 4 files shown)


dcadenas (changes requested):
Two moderation paths need changes before merge: age-review enforcement failures must not look successful, and combined delete must preserve audit state across partial failure. Both are normal moderator actions, so the PR needs to make these failure modes visible and recoverable.

---

- **Surface failed account enforcement** ([src/components/AgeReviewDetail.tsx:100](https://github.com/divinevideo/divine-relay-manager/blob/070b21948b8eb2398e3a3cac76975abd4775b08b/src/components/AgeReviewDetail.tsx#L100))
  Treat `keycastUpdated: false` as a failed moderation action here.
  The worker can persist the case change and still return 200 when Keycast suspension or ban fails, so the UI currently tells the moderator the action worked while account enforcement did not happen.
  Show a retryable error, or change the API contract so required account enforcement failures fail the state transition.
- **Preserve audit state for partial deletes** ([src/components/EventActions.tsx:176](https://github.com/divinevideo/divine-relay-manager/blob/070b21948b8eb2398e3a3cac76975abd4775b08b/src/components/EventActions.tsx#L176))
  Preserve an audit record as each irreversible side effect succeeds in this combined delete flow.
  The event delete and deletion-request publish can both complete before media deletion or the final decision log fails, which leaves the mutation in a failed state after moderation work has already happened.
  Log completed side effects immediately, or return a partial-success result that refreshes the UI and makes the audit trail explicit.

dcadenas (approved):
✅ safer moderation flow, ship it

PR #84, opened 2026-05-22
cleanup: remove unused Create Label button from report detail

mbradley:
## Summary

- Removes the "Create Label" button and inline `LabelPublisherInline` form from the report detail view
- The button had a UX bug: clicking it hid the button and rendered the form in a different scroll zone above the action bar, making it appear like nothing happened
- Manual labeling was never part of the active moderation workflow — the AI moderation service auto-publishes kind 1985 labels, and moderators act via direct RPC actions (ban, suspend, delete)
- The standalone `LabelPublisher` component and Labels tab are retained

## Test plan

- [ ] Verify report detail view renders without "Create Label" button
- [ ] Verify no TypeScript errors (`npx tsc --noEmit`)
- [ ] Verify Labels tab still works independently


dcadenas (approved):
🧹 unused label button is gone

PR #86, opened 2026-05-22
config: wire Zendesk credentials via Secrets Store bindings

mbradley:
## Summary

Provisions Cloudflare Secrets Store bindings for Zendesk API credentials (`ZENDESK_SUBDOMAIN`, `ZENDESK_API_TOKEN`, `ZENDESK_EMAIL`, `ZENDESK_AGE_REVIEW_WEBHOOK_SECRET`) in staging and prod wrangler configs. Updates env interfaces and credential resolution to handle `SecretStoreSecret` objects.

Without these bindings, all Zendesk ticket creation and sync for age review cases silently no-ops because `getZendeskClientConfig()` returns null when the credentials are absent.

**Changes:**
- Add `secrets_store_secrets` bindings to `wrangler.staging.toml` and `wrangler.prod.toml`
- Add shared `resolveZendeskCreds` helper in `zendesk-sync.ts` handling both `string` and `SecretStoreSecret` types
- Update `age-review.ts` to delegate credential resolution to `resolveZendeskCreds`
- Fix fire-and-forget `updateZendeskTicket` call to use awaited try/catch
- Add protective comments on plain-string Zendesk secrets passed directly to crypto APIs

No behavioral change when credentials are absent -- existing no-op guards remain intact.

## Test plan

- [x] `cd worker && npx vitest run` passes (183 tests)
- [x] Deploy to prod, restrict age review case -- Zendesk ticket #2441 created
- [x] Clear case -- Zendesk ticket solved via `syncAgeReviewTicketResolution`
- [x] E2E verified with `invited.divine.video` test account

### DIFF
--- a/worker/src/age-review.ts
+++ b/worker/src/age-review.ts
@@ -11,14 +11,16 @@ import {
   defaultResolutionForBand,
 } from '../../shared/age-review';
 import { handleBulkModerate, type BulkModerateEnv } from './bulk-moderate';
+import { resolveZendeskCreds } from './zendesk-sync';
 import type { BulkAction, BulkModerateResult } from '../../shared/bulk-moderation';
 import { suspendUser, unsuspendUser, banUser, type KeycastEnv } from './keycast-client';
+import type { SecretStoreSecret } from './nip86';
 
 interface AgeReviewEnv extends BulkModerateEnv, KeycastEnv {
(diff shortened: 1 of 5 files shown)


dcadenas (approved):
🔐 secrets wiring lands cleanly

PR #87, opened 2026-05-22
fix(age-review): correct Keycast path and Zendesk Issue field

mbradley:
## Summary

- **Keycast API path**: `/admin/users/:pubkey/status` -> `/api/admin/users/:pubkey/status`. The wrong path hit SvelteKit's catch-all and returned 405, so Keycast suspend/unsuspend calls during age review silently failed.
- **Zendesk Issue field**: `age_review` -> `content_report_under_16`. `age_review` is not a valid dropdown option in the Zendesk field configuration, causing the custom field to be silently dropped on ticket creation and 422 on ticket resolution.

Both discovered during E2E verification of the under-16 age review flow.

Relates to divinevideo/support-trust-safety#145

## Test plan

- [x] `cd worker && npx vitest run` passes (183 tests, assertions updated)
- [x] Deploy to prod, restrict age review case -> `keycastUpdated: true` (was silently failing)
- [x] Clear case -> `keycastUpdated: true` (unsuspend works)
- [x] Zendesk ticket #2441 created with correct Issue field value
- [x] E2E verified with `invited.divine.video` and `u16.divine.video` test accounts

### DIFF
--- a/worker/src/age-review.test.ts
+++ b/worker/src/age-review.test.ts
@@ -326,7 +326,7 @@ describe('handleUpdateAgeReviewCase', () => {
     expect(payload.ticket.tags).toEqual(['age-review', 'age-band-age_16_plus_claimed', 'internal']);
     expect(payload.ticket.custom_fields).toEqual([
       { id: 1001, value: 'trust___safety' },
-      { id: 1002, value: 'age_review' },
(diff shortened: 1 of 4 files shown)


dcadenas (approved):
✅ Keycast and Zendesk fixes land cleanly

PR #88, opened 2026-05-22
fix(moderate): route ban/unban through NIP-86 and correct method names

mbradley:
## Summary

- **ban_pubkey** published a kind 10000 event via WebSocket -- the relay accepted it but never processed it as a management command. Bans from the admin UI silently failed. Refactored to route through `handleRelayRpc` (NIP-86 HTTP POST with NIP-98 auth).
- **allow_pubkey** sent `allowpubkey` to the relay. Funnelcake accepts this silently but has no implementation. Changed to `unbanpubkey`.
- **allowevent** -> `unbanevent` across all frontend components and `nip86.ts`. Same issue -- Funnelcake's methods are `unban*`, not `allow*`.
- Removed duplicate DM notification from `ban_pubkey` (handleRelayRpc already sends one).
- Restored `markHumanReviewed` on `allow_pubkey` (dropped during refactor).
- Removed ~40 lines of dead code (unreachable `finalizeEvent`/`publishToRelay` block).
- Fixed pre-existing `bulk-moderate.test.ts` failure from vitest 4.x breaking change.

Combined effect: ban/unban user and restore event buttons in the admin UI were completely non-functional. Both now work.

Fixes divinevideo/support-trust-safety#147
See also divinevideo/divine-funnelcake#450 (Funnelcake should reject unknown methods)

## Review feedback addressed

- **Restore media actions filtered by per-hash status** -- Unblock only sends `SAFE` to blocked hashes; Remove Restriction only clears restricted hashes. `EventActions` now receives `mediaHashStatuses` with per-hash blocked/restricted flags instead of operating on all hashes blindly.
- **Profile-card delete audit row** -- `onDeleteEvent` in `ReportDetail` now calls `logDecision` with `targetType: 'event'`, action, reason, and `reportId` after `deleteEvent` succeeds.
- **Kind 5 deletion via WebSocket** -- `publishKind5Deletion` now opens a WebSocket to `RELAY_URL`, sends `["EVENT", ...]`, and waits for the relay's `["OK", ...]` response with a 10s timeout. Replaces the HTTP POST that standard relays don't accept for event ingestion.

## Files changed

| File | Change |
|------|--------|
| `worker/src/index.ts` | `ban_pubkey`: WebSocket -> `handleRelayRpc`. `allow_pubkey`: `allowpubkey` -> `unbanpubkey`. Dead code removed. |
| `worker/src/nip86.ts` | `allowEvent()`: `allowevent` -> `unbanevent`. `publishKind5Deletion`: HTTP POST -> WebSocket with OK wait. |
| `worker/src/nip86.test.ts` | Assertions updated. Kind 5 tests rewritten for WebSocket mock (OK, reject, error, timeout). |
| `worker/src/bulk-moderate.test.ts` | Arrow function -> function expression for vitest 4.x |
| `src/components/EventActions.tsx` | `allowevent` -> `unbanevent`. Restore mutations filter by per-hash status. Separate `removeRestrictionMutation`. |
| `src/components/EventModeration.tsx` | `allowevent` -> `unbanevent` |
| `src/components/ReportDetail.tsx` | `allowevent` -> `unbanevent`. Profile-card delete now logs audit decision with reportId. Passes `mediaHashStatuses` to EventActions. |
| `src/components/EventsList.tsx` | `allowevent` -> `unbanevent` |
| `src/components/EventDetail.tsx` | `allowevent` -> `unbanevent` |
| `src/components/UserManagement.tsx` | `allowpubkey` -> `unbanpubkey` |

## Test plan

- [x] `cd worker && npx vitest run` passes (184 tests)
- [x] `grep -rn "allowevent\|allowpubkey" src/` returns zero hits
- [ ] Deploy to staging, verify ban/unban and restore flows
- [ ] Verify kind-5 deletion publishes and relay returns OK
- [ ] Verify audit log row appears for profile-card deletes

### DIFF
--- a/src/components/EventActions.tsx
+++ b/src/components/EventActions.tsx
@@ -6,10 +6,17 @@ import { useAdminApi } from '@/hooks/useAdminApi';
 import { DeleteConfirmDialog } from './DeleteConfirmDialog';
 import { ShieldX, Undo2, Video, ShieldAlert, Unlock, Trash2 } from 'lucide-react';
 
+export interface MediaHashStatus {
+  hash: string;
+  isBlocked: boolean;
+  isRestricted: boolean;
+}
+
 interface EventActionsProps {
   eventId: string;
   pubkey: string;
   mediaHashes?: string[];
+  mediaHashStatuses?: MediaHashStatus[];
   isEventBanned?: boolean;
   hasBlockedMedia?: boolean;
   hasRestrictedMedia?: boolean;
@@ -20,6 +27,7 @@ export function EventActions({
   eventId,
   pubkey,
   mediaHashes = [],
+  mediaHashStatuses = [],
   isEventBanned = false,
   hasBlockedMedia = false,
   hasRestrictedMedia = false,
@@ -131,14 +139,16 @@ export function EventActions({
 
   const unblockMediaMutation = useMutation({
     mutationFn: async () => {
-      for (const hash of mediaHashes) {
+      const blockedHashes = mediaHashStatuses.filter(s => s.isBlocked).map(s => s.hash);
+      if (blockedHashes.length === 0) return;
+      for (const hash of blockedHashes) {
         await api.moderateMedia(hash, 'SAFE', 'Unblocked by moderator');
       }
       await api.logDecision({
         targetType: 'event',
         targetId: eventId,
         action: 'unblock_media',
-        reason: `Unblocked/unrestricted ${mediaHashes.length} media file(s)`,
+        reason: `Unblocked ${blockedHashes.length} media file(s)`,
       });
     },
     onSuccess: () => {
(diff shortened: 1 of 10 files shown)


dcadenas (changes requested):
Several moderation actions can now do the wrong thing or skip required audit and relay confirmation. Please fix the inline issues before merge so restore, delete, and deletion-publish behavior stay safe.

---

- **Restore media actions affect every hash on the event** ([src/components/EventActions.tsx:134](https://github.com/divinevideo/divine-relay-manager/blob/c39b473cefc583edf6ee1e5d5145f56ab8e238bf/src/components/EventActions.tsx#L134))
  Filter restore actions to the hashes that actually match the selected state before sending `SAFE`.
  Right now this path loops over every media hash passed into `EventActions`, while `ReportDetail` passes all hashes and only aggregate flags.
  On an event with a restricted hash and a blocked hash, Remove Restriction can also unblock the blocked hash.
  Pass per-hash status or derive filtered lists so Unblock only clears blocked hashes and Remove Restriction only clears restricted hashes.
- **Profile delete path skips the moderation decision audit row** ([src/components/ReportDetail.tsx:794](https://github.com/divinevideo/divine-relay-manager/blob/c39b473cefc583edf6ee1e5d5145f56ab8e238bf/src/components/ReportDetail.tsx#L794))
  Log the moderation decision before refreshing after this delete path.
  This callback now calls `deleteEvent` directly from the reported-user card, but the replaced mutation also wrote `moderation_decisions` with the report id.
  That lets a normal report-review delete remove relay content without the audit row tied to the report.
  Await `logDecision` with targetType `event`, action `delete_event`, the deleted event id, reason, and `reportId: report?.id` after `deleteEvent` succeeds.
- **Kind 5 deletion publishing uses the wrong relay transport** ([worker/src/nip86.ts:220](https://github.com/divinevideo/divine-relay-manager/blob/c39b473cefc583edf6ee1e5d5145f56ab8e238bf/worker/src/nip86.ts#L220))
  Publish kind-5 deletion events through the relay event transport that waits for acceptance.
  This helper builds a normal Nostr `["EVENT", event]` frame but posts it with `fetch()` to an HTTP URL derived from `RELAY_URL`.
  Standard Nostr relays accept `EVENT` frames over WebSocket and return `OK`, so permanent delete can ban the event but fail to publish the NIP-09 deletion request.
  Use the existing WebSocket publish path or a documented HTTP ingestion endpoint with a test against that exact endpoint.

dcadenas (approved):
✅ cleaner moderation routing, ship it

PR #89, opened 2026-05-23
fix(nip86): use allowevent instead of unbanevent for event unbans

mbradley:
## Summary

- PR #88 changed event unban RPC calls from `allowevent` to `unbanevent`, but Funnelcake does not implement `unbanevent` — the method returns `{"error": "Unknown method: unbanevent"}`, silently breaking all event restore/unban actions in the UI and worker.
- Funnelcake's method naming is asymmetric: `unbanpubkey` exists and correctly removes from the banned set, but the event equivalent is `allowevent` (not `unbanevent`). `allowpubkey` is a separate operation that adds to an allowlist without unbanning.
- This fix reverts the event method name to `allowevent` in `nip86.ts` and all 5 frontend components that call the relay directly.

**Affected actions (all broken since PR #88 merge):**
- Restore auto-hidden content (ReportDetail)
- Undo delete (ReportDetail toast action)
- Restore event (EventDetail)
- Allow event (EventActions)
- Allow event (EventModeration)
- Allow event in bulk (EventsList)

## Method name reference

| Operation | Correct method | What PR #88 used | Funnelcake behavior |
|-----------|---------------|-----------------|-------------------|
| Ban event | `banevent` | `banevent` | OK |
| Unban event | `allowevent` | `unbanevent` | **"Unknown method"** error |
| Ban pubkey | `banpubkey` | `banpubkey` | OK |
| Unban pubkey | `unbanpubkey` | `unbanpubkey` | OK |

The asymmetry exists because Funnelcake implements `allowevent` (which removes from the ban cache + ClickHouse banned set) but never added `unbanevent` as an alias. For pubkeys, `unbanpubkey` exists separately from `allowpubkey` (which adds to a whitelist).

## Also included: lint fix inherited from #90

Rebasing this branch onto `main` after #90 merged pulled in 10 `@typescript-eslint/no-explicit-any` eslint errors in `worker/src/age-review.test.ts`. They slipped through #90's review and turned `main`'s CI red, which this branch then inherited. Rather than open a separate cleanup PR, the fix is folded in here: `{ DB: db } as any` becomes `{ DB: db as unknown as D1Database }`, matching the typing idiom already used throughout that test file.

This is a test-only change with no runtime effect. The compiled worker bundle is byte-identical before and after (verified by comparing `wrangler deploy --dry-run` output hashes), so it does not touch the unban fix or the deployed artifact.

Note: `main` itself stays red until this PR merges, since the fix lives on this branch rather than a standalone PR to `main`.

## Test plan

- [x] `npx tsc --noEmit` — clean
- [x] `cd worker && npx vitest run` — 209 tests pass (nip86.test.ts updated)
- [x] eslint — clean (resolves the failures inherited from #90)
- [x] `grep -rn unbanevent src/ worker/src/` — zero hits
- [x] Deployed to staging and production; banned then unbanned an event end to end — unban succeeds via `allowevent`

### DIFF
--- a/src/components/EventActions.tsx
+++ b/src/components/EventActions.tsx
@@ -39,7 +39,7 @@ export function EventActions({
 
   const banEventMutation = useMutation({
     mutationFn: async () => {
-      await api.callRelayRpc('banevent', [eventId, 'Banned by moderator']);
+      await api.banEvent(eventId, 'Banned by moderator');
       await api.logDecision({
         targetType: 'event',
         targetId: eventId,
@@ -58,7 +58,7 @@ export function EventActions({
 
   const restoreEventMutation = useMutation({
     mutationFn: async () => {
-      await api.callRelayRpc('allowevent', [eventId]);
+      await api.allowEvent(eventId);
       await api.logDecision({
         targetType: 'event',
         targetId: eventId,
--- a/src/components/EventActions.tsx
+++ b/src/components/EventActions.tsx
@@ -58,7 +58,7 @@ export function EventActions({
 
   const restoreEventMutation = useMutation({
     mutationFn: async () => {
-      await api.callRelayRpc('unbanevent', [eventId]);
+      await api.callRelayRpc('allowevent', [eventId]);
       await api.logDecision({
         targetType: 'event',
         targetId: eventId,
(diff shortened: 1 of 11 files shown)


NotThatKindOfDrLiz (approved):
Reviewed the allowevent fix and the follow-up helper cleanup. Verified against Funnelcake method definitions, checked the PR scope, and confirmed CI is green.

PR #90, opened 2026-05-27
feat: approved minor onboarding via claim link

mbradley:
## Summary

- Worker endpoint `POST /api/age-review/create-minor-account` proxies to Keycast, creates D1 audit record with `state=cleared` and `created_via=minor_onboarding`
- ReportWatcher auto-clears age review cases for users with `verified_minor=true` in Keycast (fail-open if Keycast unreachable)
- Admin UI: CreateMinorAccountDialog with username/display name/Zendesk ticket ID form, claim link display on success
- D1 migration + ensureSchema for `created_via` and `claim_link_url` columns
- "Minor Onboarding" and "Auto-cleared" badges in case detail view

Relates to divinevideo/support-trust-safety#148
Part of divinevideo/support-trust-safety#138

Depends on divinevideo/keycast#244 being deployed first (new `create-minor-account` endpoint + `verified_minor` flag on status response).

## Test plan

- [ ] Create minor account via admin UI with username + optional display name + optional Zendesk ticket ID
- [ ] Success state shows pubkey and copyable claim link
- [ ] Invalid username (special chars, too long) shows validation error
- [ ] Duplicate username shows 409 conflict error
- [ ] Created case appears in age review list with "Cleared" state
- [ ] Case detail shows "Minor Onboarding" badge
- [ ] Submit an age report for a verified minor; case auto-clears with "Auto-cleared" badge
- [ ] Submit an age report for a non-verified user; case creates normally with deadline
- [ ] Keycast unreachable during report: case still creates normally (fail-open)

## Local verification

- [x] tsc, eslint, vitest all pass locally
- [x] CreateMinorAccountDialog renders in Age Review tab header
- [x] Dialog opens with username (required), display name (optional), zendesk ticket ID (optional) fields
- [x] Submit button disabled when username is empty
- [x] Vite build succeeds (CI green)

### DIFF
--- a/worker/src/ReportWatcher.ts
+++ b/worker/src/ReportWatcher.ts
@@ -17,11 +17,12 @@ import {
   TERMINAL_STATES,
   defaultResolutionForBand,
 } from '../../shared/age-review';
+import { getUserStatus, type KeycastEnv } from './keycast-client';
 
 /**
  * Extended environment for ReportWatcher DO
  */
-export interface ReportWatcherEnv extends Nip86Env {
+export interface ReportWatcherEnv extends Nip86Env, KeycastEnv {
   DB?: D1Database;
   AUTO_HIDE_ENABLED?: string;
   TRUSTED_CLIENTS?: string;
@@ -909,6 +910,33 @@ export class ReportWatcher implements DurableObject {
     if (existing) {
       console.log(`[ReportWatcher] Active age review case ${existing.id} already exists for ${reportedPubkey}, skipping`);
       return;
+    }
+
+    // Auto-clear: if user is a previously verified minor, create case as immediately cleared
+    try {
+      const keycastStatus = await getUserStatus(reportedPubkey, this.env);
+      if (keycastStatus.success && keycastStatus.verified_minor) {
+        const caseId = crypto.randomUUID();
+        await this.env.DB.prepare(`
+          INSERT INTO age_review_cases
+          (id, pubkey, reporter_pubkey, report_id, suspected_age_band, state, allowed_resolution, resolution_note, created_via)
+          VALUES (?, ?, ?, ?, 'age_13_15', 'cleared', 'parent_video_or_email', 'Auto-cleared: previously verified minor', 'report')
+        `).bind(caseId, reportedPubkey, event.pubkey, event.id).run();
+
+        await this.logDecision({
+          targetType: 'pubkey',
+          targetId: reportedPubkey,
+          action: AGE_REVIEW_ACTION.caseCreated,
+          reason: `Under-16 report auto-cleared: verified minor (case ${caseId})`,
+          reportId: event.id,
+          reporterPubkey: event.pubkey,
+        });
+
+        console.log(`[ReportWatcher] Age review case ${caseId} auto-cleared for verified minor ${reportedPubkey}`);
+        return;
+      }
+    } catch (err) {
+      console.warn(`[ReportWatcher] Keycast verified_minor check failed for ${reportedPubkey}, proceeding with normal case:`, err);
     }
 
     const caseId = crypto.randomUUID();
--- a/worker/src/ReportWatcher.ts
+++ b/worker/src/ReportWatcher.ts
@@ -458,9 +458,10 @@ export class ReportWatcher implements DurableObject {
       });
 
       this.ws.addEventListener('message', (event) => {
-        this.handleMessage(event.data as string).catch(error => {
+        const work = this.handleMessage(event.data as string).catch(error => {
           console.error('[ReportWatcher] Message handler failed:', error);
         });
+        this.state.waitUntil(work);
       });
 
       this.ws.addEventListener('close', (event) => {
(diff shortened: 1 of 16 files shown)


worker/src/keycast-client.ts
```diff
@@ -74,3 +74,100 @@ export async function unsuspendUser(pubkey: string, env: KeycastEnv): Promise<Ke
...
+      return { success: false, error: `${res.status}: ${text}` };
+    }
+    const data = await res.json() as Record<string, unknown>;
+    return {
+      success: true,
+      pubkey: data.pubkey as string,
+      status: data.status as string,
+      suspended_reason: data.suspended_reason as string | undefined,
+      suspended_at: data.suspended_at as string | undefined,
+      verified_minor: data.verified_minor as boolean | undefined,
```
dcadenas: Validate `verified_minor` before using it to auto-clear an age-review case.
`getUserStatus` currently passes untyped Keycast JSON through to `ReportWatcher`, where any truthy value clears the report.
A response like `"false"` or `1` would skip normal review for an underage-user report.
Decode the response at runtime and only treat `verified_minor === true` as approved; malformed or missing values should fall through to the normal open case path, with tests for `"false"`, `1`, missing, and `true`.

worker/src/age-review.ts
```diff
@@ -285,6 +285,64 @@ export async function handleUpdateAgeReviewCase(
...
+
+  // Create a cleared audit record in D1. If this fails, the Keycast account
+  // exists but has no audit trail -- log enough context to recover manually.
+  const caseId = crypto.randomUUID();
+  try {
+    await env.DB.prepare(`
+      INSERT INTO age_review_cases
+      (id, pubkey, suspected_age_band, state, allowed_resolution, resolution_note, created_via, claim_link_url, zendesk_ticket_id)
+      VALUES (?, ?, 'age_13_15', 'cleared', 'parent_video_or_email', 'Approved via parental consent (minor onboarding)', 'minor_onboarding', ?, ?)
+    `).bind(caseId, result.pubkey, result.claim_url, body.zendesk_ticket_id ?? null).run();
```
dcadenas, line 340: Make the local audit write part of a successful minor-account response.
This path can create the Keycast account and one-time claim link, then swallow a D1 insert failure and still return `{ success: true, claim_url, case_id }`.
That leaves operators without the durable moderation record for an already-created minor account, and the recovery log includes the claim URL.
Reject malformed `zendesk_ticket_id` before calling Keycast, and if the D1 insert fails after Keycast succeeds, return a recoverable error or persist a recovery record without logging the claim URL.

dcadenas (changes requested):
This PR needs changes before merge because the new minor-onboarding path can clear or create safety records without strong checks around external data and durable persistence. Please make the Keycast status check strict, make the audit write part of a successful claim-link response, and attach underage-report work to the Durable Object lifecycle before this merges.

---

- **Attach age-review side effects to the Durable Object lifecycle** ([worker/src/ReportWatcher.ts:633](https://github.com/divinevideo/divine-relay-manager/blob/b9743d92c1f82fd63e7511fb6513e2331e5decec/worker/src/ReportWatcher.ts#L633))
  Attach `createAgeReviewCase` to the Durable Object work before `handleReportEvent` returns.
  The `NS-underageUser` branch starts Keycast and D1 work as an untracked promise, so the WebSocket handler can finish before the safety case is durably scheduled or written.
  Await the async path, register it with the Durable Object lifetime, or persist a small queue item before returning.
  Keep the pending-pubkey guard behavior and add a test that underage reports are not considered processed until the insert path is scheduled or complete.

worker/src/ReportWatcher.ts
```diff
@@ -457,7 +458,9 @@ export class ReportWatcher implements DurableObject {
       });
 
       this.ws.addEventListener('message', (event) => {
-        this.handleMessage(event.data as string);
+        this.handleMessage(event.data as string).catch(error => {
```
dcadenas: This listener kicks off the age-review work and then returns, so nothing keeps the Durable Object alive until that work finishes.

The awaits you added downstream are correct. The `NS-underageUser` branch in `handleReportEvent` awaits `createAgeReviewCase` (ReportWatcher.ts:634), so the Keycast and D1 writes are properly sequenced. The problem is one level up. This callback calls `this.handleMessage(...).catch(...)` and returns immediately, so the whole chain runs in a detached promise. If the object is reset, hibernated, or evicted before that promise settles, the case creation is lost and the underage-user report ends up with no age-review case.

Attach the work to the object's lifecycle before the callback returns. The lightweight fix is to pass the promise to `this.state.waitUntil(...)` so the runtime waits for it. The durable fix, which also survives hibernation and eviction, is to persist the report to storage or an alarm here and run the Keycast and D1 work from there.

Keep the `pendingAgeReviewPubkeys` guard, or swap it for a durable idempotency key if you move to a queue. Add a regression test that proves an `NS-underageUser` message is not treated as fully processed until the case-creation path is attached to the lifecycle or durably queued.

dcadenas (changes requested):
The relay `message` listener starts the underage-user case creation in a detached promise, so that work can be lost if the Durable Object is reset, hibernated, or evicted before it finishes. That would leave an underage-user report with no age-review case, so it needs to be fixed before this can merge. Attach the work to the object's lifecycle or a durable queue before the listener returns, and add a regression test that pins the behavior. The rest of the change looks solid.

dcadenas (approved):
🎟️ claim-link onboarding for approved minors reads clean, ship it.

PR #91, opened 2026-06-02
perf: use REST API for bulk-moderate video hashes

mbradley:
## Summary

- Replaces WebSocket REQ with Funnelcake REST API (`/api/users/{pubkey}/videos`) for fetching video sha256 hashes in bulk-moderate actions
- Fixes silent data loss where age-restrict-all / un-age-restrict-all missed videos due to Funnelcake's addressable event dedup bug
- For delete-all, fetches events (WebSocket) and media hashes (REST) in parallel for correctness and speed

## Motivation

Funnelcake's `relay_events_by_kind_time` materialized view deduplicates addressable events by `(pubkey, kind)` instead of `(pubkey, kind, d_tag)`, returning only 1 video per kind instead of all videos. The REST API routes through the correct `events_deduped` view.

Workaround for divinevideo/divine-funnelcake#471.

Secondary goal: the REST API returns `sha256` directly in the response, eliminating tag parsing and reducing latency vs. a WebSocket REQ round-trip.

## Test plan

- [x] 18 bulk-moderate tests pass (4 new: REST API unit tests)
- [x] Full worker suite: 191 tests pass
- [x] `npx tsc --noEmit` clean
- [ ] Deploy to staging, test age-restrict-all on a multi-video profile
- [ ] Verify all video hashes are sent to moderation service (not just 1)

Closes #107.

---

## Update — rebuilt on current main (force-pushed)

The branch was 20 commits behind main and its old form reverted two landed fixes (#120 `age-restrict-all` → QUARANTINE withhold, and #121 event pagination). I reset it onto current main and re-applied **only** the REST media-hash swap on top, so it preserves both:

- `age-restrict-all` stays `QUARANTINE` (withhold), never `AGE_RESTRICTED` — guarded by a test.
- `queryRelayEvents` stays paginated; the partial-set failure is surfaced for delete-all.

Review fixes also folded in: `queryUserMediaHashes` now routes through the shared `deriveFunnelcakeApiUrl` (honors `FUNNELCAKE_API_URL`, strips trailing slash) and bounds the fetch with a 10s `AbortSignal`; added fail-closed tests (a videos-REST 500 throws, with no moderation calls and no events banned); removed the now-dead WS-based `extractMediaHashes`.

Worker suite green (231); tsc + eslint clean. The funnel-returns-all-videos leg remains a staging smoke test (local funnel `/api/videos` 500s for an unrelated funnelcake-side reason). Sequencing: #134 → this → #135 (async).

---

## Update 2 — paginate the videos endpoint (deploy-review fix)

The funnelcake `/api/users/{pubkey}/videos` endpoint defaults to `limit=25` (max 100). The first rebuild fetched it without `?limit`, so bulk delete/age-restrict would silently action only the first 25 videos and leave the rest live on a withhold path. Now pages at `limit=100` with `offset` until a short page (safety bound that throws rather than under-enforcing), and the test mock honors `limit`/`offset` (with a 250-video / 3-page regression test) so the gap can't hide.

### DIFF
--- a/worker/src/bulk-moderate.ts
+++ b/worker/src/bulk-moderate.ts
@@ -1,7 +1,7 @@
 import { getAdminPubkey, banEvent, publishKind5Deletion, type Nip86Env } from './nip86';
 import { syncZendeskAfterAction, type ZendeskSyncEnv } from './zendesk-sync';
 import { VALID_BULK_ACTIONS, type BulkAction, type BulkModerateResult } from '../../shared/bulk-moderation';
-import { extractMediaHashes as extractSharedMediaHashes } from '../../shared/media-hashes';
+import { deriveFunnelcakeApiUrl } from './funnelcake-proxy';
 
 const BULK_ACTION_CONCURRENCY = 5;
 // Page through ALL of an author's events via `until` cursoring instead of
@@ -16,6 +16,8 @@ export interface BulkModerateEnv extends Nip86Env, ZendeskSyncEnv {
   MODERATION_API?: Fetcher;
   MODERATION_ADMIN_URL?: string;
   SERVICE_API_TOKEN?: string | { get(): Promise<string> };
+  // Explicit Funnelcake REST API URL; derived from RELAY_URL when unset.
+  FUNNELCAKE_API_URL?: string;
 }
 
 interface RelayEventSummary {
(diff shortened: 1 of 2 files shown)


worker/src/bulk-moderate.ts
```diff
@@ -242,13 +258,56 @@ export async function queryRelayEvents(
...
+      throw new Error(`Video query failed: ${res.status}`);
+    }
+    const videos = await res.json() as Array<{ sha256?: string }>;
+    for (const v of videos) {
+      if (v.sha256 && SHA256_HEX.test(v.sha256)) {
+        hashes.add(v.sha256.toLowerCase());
+      }
+    }
+    // A short page (fewer than the page size) means we've reached the end.
+    if (videos.length < VIDEO_PAGE_SIZE) {
```
dcadenas: This stops paging as soon as a page has fewer than 100 rows, which assumes the server always returns exactly 100 per full page.
If the endpoint ever caps or clamps the page size below 100, the very first page already looks short and the loop returns after page 0 with only that page enumerated.
The 10000-video guard below won't catch this, because the function returns early instead of reaching the bound, so the fail-closed protection is bypassed and the action withholds or deletes only the first page.
Terminate on a zero-row page, or compare against the page size the server actually returned, so a server cap below 100 cannot be read as end-of-data.

worker/src/bulk-moderate.ts
```diff
@@ -242,13 +258,56 @@ export async function queryRelayEvents(
...
+    const res = await fetch(
+      `${baseUrl}/api/users/${pubkey}/videos?limit=${VIDEO_PAGE_SIZE}&offset=${offset}`,
+      { signal: AbortSignal.timeout(VIDEO_QUERY_TIMEOUT_MS) },
+    );
+    if (!res.ok) {
+      throw new Error(`Video query failed: ${res.status}`);
+    }
+    const videos = await res.json() as Array<{ sha256?: string }>;
+    for (const v of videos) {
+      if (v.sha256 && SHA256_HEX.test(v.sha256)) {
```
dcadenas, line 316: This keeps only rows whose `sha256` is present and valid hex and silently drops the rest, which fails open on this path.
If the videos endpoint renames the field, returns a non-string, or otherwise drifts, a 200 full of video rows yields an empty hash set and the bulk action reports success having actioned nothing.
Pick a deliberate posture when a page returns rows but no valid hashes: surface a failure rather than treating zero hashes as a clean success.
This is the case most likely to bite, since the moderator would see age-restrict-all succeed while none of the account's media was withheld.

worker/src/bulk-moderate.ts
```diff
@@ -242,13 +258,56 @@ export async function queryRelayEvents(
...
+  for (let page = 0; page < VIDEO_MAX_PAGES; page++) {
+    const offset = page * VIDEO_PAGE_SIZE;
+    const res = await fetch(
+      `${baseUrl}/api/users/${pubkey}/videos?limit=${VIDEO_PAGE_SIZE}&offset=${offset}`,
+      { signal: AbortSignal.timeout(VIDEO_QUERY_TIMEOUT_MS) },
+    );
+    if (!res.ok) {
+      throw new Error(`Video query failed: ${res.status}`);
+    }
+    const videos = await res.json() as Array<{ sha256?: string }>;
```
dcadenas: This casts the body straight to an array and then iterates it, so a non-array 200 body throws 'is not iterable'.
A pagination wrapper like `{videos: [...], total}`, a `null`, or an error envelope would all trip it, and the error surfaces as a generic 500 from the route handler.
That does fail closed, but it breaks every bulk-moderate request and the cast hides the shape mismatch from the type checker.
Validate that the parsed body is an array and throw a clear error if it is not, instead of asserting the shape with a cast.

worker/src/bulk-moderate.ts
```diff
@@ -242,13 +258,56 @@ export async function queryRelayEvents(
...
+    // A short page (fewer than the page size) means we've reached the end.
+    if (videos.length < VIDEO_PAGE_SIZE) {
+      return Array.from(hashes);
+    }
   }
-  return Array.from(hashes);
+
+  // Hit the page bound: fail closed rather than silently enforcing over a partial
+  // set on a withhold path.
+  throw new Error(`More than ${VIDEO_MAX_PAGES * VIDEO_PAGE_SIZE} videos for ${pubkey}; narrow the scope or add deeper pagination`);
```
dcadenas: This throws when the loop exits without ever seeing a short page, but an account whose video count exactly fills the last allowed page also exits that way with every hash already fetched.
At exactly 10000 videos, page index 99 returns a full 100, the loop condition goes false, and this throws even though enumeration actually completed.
So a moderator can never bulk-action an account at exactly 10000 videos; it 500s every time.
Give the bound one page of headroom, or confirm completion with a zero-row fetch before throwing, while keeping the genuine over-limit throw for counts beyond that.

dcadenas (dismissed):
The new REST enumeration in `queryUserMediaHashes` decides it has fetched every video by trusting the response shape, and a few of those assumptions fail open on a child-safety path.
A page shorter than 100 is read as the end of the list, rows with a missing or invalid `sha256` are dropped without any signal, and a non-array body throws an opaque 500.
The exact-10000-video case also fails closed and can never be actioned at all.
Please harden the page-termination, row-validation, and page-bound logic before merge so a quiet change in the videos response can't leave a flagged account's media live while the moderator sees success.

dcadenas (approved):
✅ paged REST enumeration with a hard bound and clean termination, ship it.

PR #97, opened 2026-06-03
feat(age-review): claim link display + expiry

mbradley:
Closes #92. Addresses #93 Part A (expiry persistence + display). #93 Part B (link regeneration) stays open pending a Keycast capability spike.

## What this does

Lets a moderator retrieve and act on an approved-minor onboarding claim link after the creation dialog is closed, and see when it expires.

## Changes

- **`worker/src/db.ts`** — added `claim_link_expires_at TEXT` to `age_review_cases` (added to `CREATE TABLE` plus an idempotent `ALTER` in `ensureSchema`, matching the `created_via`/`claim_link_url` pattern) + mirror migration `worker/migrations/0010_claim_link_expiry.sql`.
- **`worker/src/age-review.ts`** — `handleCreateMinorAccount` now persists `result.expires_at` into `claim_link_expires_at` (stores `null` when Keycast omits it).
- **`shared/age-review.ts`** — added `claim_link_expires_at: string | null` to `AgeReviewCase`.
- **`src/components/AgeReviewDetail.tsx`** — in the `created_via === 'minor_onboarding'` block, renders the claim link (via `CopyableId`) and the expiry date, with an "Expired" badge when the expiry is in the past.
- **`src/components/CopyableId.tsx`** — added a semantic `'url'` type (used for the claim link; previously reused `'hex'`).

## Tests

- `worker/src/age-review.test.ts` — persists `claim_link_expires_at`; stores `null` when absent.
- `src/components/AgeReviewDetail.test.tsx` — renders claim link + expiry; shows Expired badge when past; shows `N/A` when no expiry.

## Notes

- Also normalizes the pre-existing `{ DB: db } as any` casts in `age-review.test.ts` to `{ DB: db as unknown as D1Database }`. This is the inherited PR #90 lint fix that keeps `main` red; it is **identical to the same fix in #89 and #98**, so these merge in any order without conflict.
- The exposed worker-`tsc` typing gap in `age-review.test.ts` (not gated in CI) is tracked in #99, to be cleaned up once these land.

## Merge order

Independent of the suspend-path PR (#98) — disjoint files. Any order. CI green.

### DIFF
--- a/shared/age-review.ts
+++ b/shared/age-review.ts
@@ -93,6 +93,7 @@ export interface AgeReviewCase {
   zendesk_ticket_id: number | null;
   created_via: string | null;
   claim_link_url: string | null;
+  claim_link_expires_at: string | null;
   created_at: string;
   updated_at: string;
 }
--- a/src/components/AgeReviewDetail.test.tsx
+++ b/src/components/AgeReviewDetail.test.tsx
@@ -154,4 +154,26 @@ describe('AgeReviewDetail', () => {
     expect(screen.getByText(/Expires: N\/A/)).toBeInTheDocument();
     expect(screen.queryByText('Expired')).not.toBeInTheDocument();
   });
+
+  it('does not render the claim link for a non-minor-onboarding terminal case', () => {
+    renderDetail(makeCase({
+      state: 'cleared',
+      created_via: 'report',
+      resolution_note: 'Cleared after review',
+      claim_link_url: 'https://login.test/claim/xyz',
+    }));
+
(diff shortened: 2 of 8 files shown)


NotThatKindOfDrLiz (approved):
Approved. I reviewed the schema, persistence path, shared type, UI rendering guards, and tests. The expiry is persisted null-safely, existing reads carry the new column through, and the claim link UI is guarded to minor-onboarding cases with a stored link. Verified focused frontend and worker tests plus frontend typecheck/build. No blocking issues found.

PR #101, opened 2026-06-04
chore(worker): gate worker typecheck + tests in CI

mbradley:
Closes #99.

## Context

Follow-on to the under-16 age-review work (#97/#98). Those PRs surfaced that the
Cloudflare Worker — which hosts the age-review backend — was never type-checked or
tested in CI. The ~64 errors this gate uncovered were almost entirely in
`age-review.test.ts`, and the age-review worker tests weren't running in CI at all.
So this PR hardens the under-16 review flow's backend (and the rest of the worker)
against silent type/test regressions. It does not change the feature, which shipped
in #97/#98 — it gates it.

## What this does

Type-checks and runs the Cloudflare Worker in CI. Previously CI only ran the root
suite, which **excludes `worker/**`** from both `tsc` (root checks `tsconfig.app.json`
only) and vitest. So the worker's TypeScript was never gated and its test suite never
ran in CI.

Turning the gate on surfaced ~64 worker `tsc` errors, almost all from one pattern:
test env-object literals (`{ DB: ... }`, `{}`) passed where handlers expect
`AgeReviewEnv`, which requires `NOSTR_NSEC`/`RELAY_URL`. These were masked before by
`{ DB: db } as any` until the eslint `no-explicit-any` rule forced the cast to change.

## Changes

- **`worker/src/age-review.ts`** — export `AgeReviewEnv`.
- **`worker/src/age-review.test.ts`** — add a typed `makeEnv()` helper and route every handler env arg through it (removes scattered per-call casts); add the minor-onboarding fields (`created_via`, `claim_link_url`, `claim_link_expires_at`) to the `makeCase` mock.
- **`worker/src/index.ts`** — `ApiResponse` gains `events?: object[]` for the relay-query endpoints (`/api/reports`, `/api/resolution-labels`) that legitimately return arrays.
- **`worker/src/bulk-moderate.ts`** — type the `queryRelayEvents` settle helper instead of `unknown`.
- **`worker/src/ReportWatcher.test.ts`**, **`worker/src/funnelcake-proxy.test.ts`** — small test-typing fixes.
- **`worker/package.json`** — add `typecheck` and reuse the `test:run` script.
- **`.github/workflows/test.yml`** — three new steps (`Worker install` / `Worker typecheck` / `Worker tests`, each `working-directory: worker`). Install uses `--legacy-peer-deps` due to a known peer-dep conflict in the worker lockfile.

No `as any` introduced.

## Verification (local, clean install)
- `worker` `tsc`: 0 errors (sanity-checked that tsc is live by injecting a deliberate error, which it caught)
- `worker` vitest: pass
- root `tsc -p tsconfig.app.json`, `eslint`: clean

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -36,7 +36,7 @@ jobs:
       # Split into separate steps so a typecheck vs test failure is obvious.
       - name: Worker install
         working-directory: worker
-        run: npm install --legacy-peer-deps
+        run: npm ci --legacy-peer-deps
 
       - name: Worker typecheck
         working-directory: worker
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -28,3 +28,14 @@ jobs:
           
       - name: Run tests
         run: npm run test
+
+      # The worker has its own package/lockfile and was previously not
+      # type-checked or tested in CI (root vitest excludes worker/**).
+      # --legacy-peer-deps: worker lockfile has a known peer-dep conflict
+      # (@cloudflare/vitest-pool-workers wants vitest 2-3.x, lockfile pins 4.x).
+      - name: Worker typecheck and tests
+        run: |
+          cd worker
+          npm install --legacy-peer-deps
+          npm run typecheck
+          npx vitest run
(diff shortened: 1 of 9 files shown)


NotThatKindOfDrLiz (approved):
Approved after final pass. Follow-up commit makes the worker typecheck gate self-contained by declaring TypeScript in the worker package and switches worker CI install to npm ci against the lockfile. Local validation and GitHub checks are green.

PR #102, opened 2026-06-09
fix(moderate): wire ban/unban into Keycast for enforcement parity (#100)

mbradley:
Closes #100.

## Problem

`banpubkey` enforced at the relay (Funnelcake hides content + blocks posting) and sent an `ACCOUNT_BANNED` DM, but never updated **Keycast** — so a banned user's Keycast account stayed `active` and they could still sign in (landing in a broken/empty state) instead of hitting a clear gate at login. Same login-side gap #94/#98 closed for suspend.

## Changes

- `banpubkey` → `banUser` (Keycast `status: banned`); **add `unbanpubkey`** → `unsuspendUser` (Keycast `active`) so a reinstated user can log in again. `unbanpubkey` sends no DM (restore-on-unban DM tracked in #96).
- Extract `enforceKeycastState()` and route ban/suspend/unsuspend/unban through it — collapses four near-identical non-critical `ctx.waitUntil` blocks (each logs+swallows failure, warns when no `ctx`).
- `handleModerate` `allow_pubkey` now passes `ctx`, so unban reaches Keycast via that path too.
- Tests (TDD): ban → Keycast `banned`, unban → Keycast `active`, on both the relay-rpc and `/api/moderate` paths. Also added a scoped `afterEach` so a failing test can't leak its global `fetch` spy and cascade into later tests.

## Notes

- Ban remains **destructive at the relay** (Funnelcake purge is irreversible); this only adds the Keycast login gate. Suspend stays the reversible tool for cases that may be restored.
- Reason value reuses `'moderation'`; the Keycast `status` (`banned` vs `suspended`) carries the distinction, so no new `KeycastReason` enum value.

## Verification (local, clean install)
- worker `tsc`: 0 errors · worker vitest: 219 pass · root `tsc -p tsconfig.app.json` + `eslint`: clean

### DIFF
--- a/worker/src/index.test.ts
+++ b/worker/src/index.test.ts
@@ -1,4 +1,4 @@
-import { describe, expect, it, vi } from 'vitest';
+import { afterEach, describe, expect, it, vi } from 'vitest';
 import worker from './index';
 
 const env = {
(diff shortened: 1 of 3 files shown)


dcadenas (approved):
🔒 bans now close the login gate, not just the relay.

PR #104, opened 2026-06-09
feat: accept a dedicated Secrets Store admin key for moderation-service -> relay-rpc (#170)

mbradley:
## Why

Paired with divinevideo/divine-moderation-service#170. The moderation dashboard's per-card **Ban User** calls this worker's `POST /api/relay-rpc` over the public host `api-relay-prod.divine.video`, which returns **HTTP 522** (the CF edge can't reach the worker — reproduced live). The fix on the moderation-service side is to call this worker via a **service binding** (worker-to-worker, by name), which bypasses the edge and DNS — but also bypasses CF Access. So this worker needs to authenticate that caller another way.

## What

`verifyAdminAccess` now **additively** accepts a dedicated shared key on the `X-Admin-Key` header:

- New `MOD_RELAY_ADMIN_KEY` binding — a Cloudflare Secrets Store secret (`MODERATION_TO_RELAY_ADMIN_KEY`, store `ef490704…`), bound in `worker/wrangler.prod.toml`.
- Accepted **in addition to** the existing `ADMIN_API_KEY` and CF Access paths — those are unchanged, so no other caller is affected.
- Resolved via the existing `resolveSecret()` helper, so `verifyAdminAccess` is now `async` (its single caller awaits it).

A dedicated secret (rather than reusing `ADMIN_API_KEY`) keeps this connection isolated, so it can be rotated without disrupting other admin-key callers.

## Deploy notes

- The secret already exists in the store. This worker's deploy already binds 9 Secrets Store secrets, so its deploy token has the needed Secrets Store access.
- The moderation-service side (divine-moderation-service#170 / its PR) sends this key. **Order:** this PR can land first safely (it only widens what's accepted); the moderation-service side must not deploy until this is live.

## Review / tests

- Independent code review: no issues (auth bypass via null/empty key, missing header, and the CF Access path all verified safe; only caller awaited; config matches the established pattern).
- `tsc` clean; full suite **219 tests pass**, including new tests proving the new key is accepted and a wrong key is rejected (401).

Relates to divinevideo/divine-moderation-service#170.

### DIFF
--- a/worker/src/index.test.ts
+++ b/worker/src/index.test.ts
@@ -395,3 +395,57 @@ describe('relay-rpc account-state side effects', () => {
     fetchSpy.mockRestore();
   });
 });
+
+describe('admin access via MOD_RELAY_ADMIN_KEY (Secrets Store shared key)', () => {
+  function makeEnvWithModKey(modKeyValue: string) {
+    return {
+      ALLOWED_ORIGINS: 'https://app.divine.video',
+      RELAY_URL: 'wss://relay.divine.video',
+      ADMIN_API_KEY: 'test-admin-key',
+      MOD_RELAY_ADMIN_KEY: { get: vi.fn(async () => modKeyValue) },
+      MODERATION_ADMIN_URL: 'https://moderation-api.divine.video',
+      SERVICE_API_TOKEN: 'svc-token',
+      MODERATION_API: {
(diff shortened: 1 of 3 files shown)


NotThatKindOfDrLiz (approved):
Final pass looks good. Follow-up commit scopes MOD_RELAY_ADMIN_KEY to POST /api/relay-rpc, updates the comments/config wording to match that contract, and adds coverage proving the dedicated key is not accepted on another admin endpoint. Local validation and GitHub checks are green.

PR #106, opened 2026-06-09
fix(staging): bind MOD_RELAY_ADMIN_KEY to match prod (#170 parity)

mbradley:
Restores prod/staging parity for the moderation-service -> relay-rpc admin key.

#104 added the `MOD_RELAY_ADMIN_KEY` Secrets Store binding (#170) to `wrangler.prod.toml` only; `wrangler.staging.toml` was missing it. Without it, the moderation-service ban path can't reach the staging worker's `/api/relay-rpc` via the shared admin key. This mirrors the prod binding exactly — same shared Secrets Store secret (`MODERATION_TO_RELAY_ADMIN_KEY`).

Validated with `wrangler deploy --config wrangler.staging.toml --dry-run`: the `MOD_RELAY_ADMIN_KEY` binding resolves.

Config-only: one binding block.

no reviewer requested on purpose 😉 seeing how fast the review bot beats a human to it.

### DIFF
--- a/worker/wrangler.staging.toml
+++ b/worker/wrangler.staging.toml
@@ -16,6 +16,14 @@ binding = "NOSTR_NSEC"
 store_id = "ef490704b12a4feb91c7feb51229c364"
 secret_name = "NOSTR_NSEC"
 
+# Shared X-Admin-Key for the divine-moderation-service -> /api/relay-rpc connection (#170).
+# verifyAdminAccess accepts this only for POST /api/relay-rpc, so the moderation-service
+# ban path can reach the worker via service binding without CF Access.
+[[secrets_store_secrets]]
+binding = "MOD_RELAY_ADMIN_KEY"
+store_id = "ef490704b12a4feb91c7feb51229c364"
+secret_name = "MODERATION_TO_RELAY_ADMIN_KEY"
+
 # CF Access credentials for realness proxy (shared across workers)
 [[secrets_store_secrets]]
 binding = "CF_ACCESS_CLIENT_ID"


worker/wrangler.staging.toml
```diff
@@ -16,6 +16,14 @@ binding = "NOSTR_NSEC"
 store_id = "ef490704b12a4feb91c7feb51229c364"
 secret_name = "NOSTR_NSEC"
 
+# Shared X-Admin-Key for the divine-moderation-service -> /api/relay-rpc connection (#170).
+# verifyAdminAccess accepts this only for POST /api/relay-rpc, so the moderation-service
+# ban path can reach the worker via service binding without CF Access.
+[[secrets_store_secrets]]
+binding = "MOD_RELAY_ADMIN_KEY"
```
dcadenas, line 23: With this key bound, the automated moderation-service ban path is live on staging, and every staging `banpubkey`, `suspendpubkey`, or `unsuspendpubkey` now ends in a DM request for the real user.
The chain is `notifyAccountState` (worker/src/index.ts:947, 954, 958) into `notifyModerationService`, which posts to `/api/v1/notify` through the `MODERATION_API` service binding, and staging's binding targets the same `divine-moderation-service` as prod (worker/wrangler.staging.toml:121-122).
The payload carries no environment field, so the DM service gets no signal to tell staging bans from prod ones:

https://github.com/divinevideo/divine-relay-manager/blob/c5b58bbf23755f598c2cb61ac9fd09dfc0227040/worker/src/index.ts#L189-L195

Pubkeys are global, so a user whose pubkey shows up in staging content can get a production-looking ban notice for an action that only touched the staging relay.
Non-blocking, but worth settling at the documented single chokepoint `notifyModerationService` (worker/src/index.ts:158): skip the post when `env.ENVIRONMENT !== 'production'` (already set in both tomls) and log the suppression, or add an `environment` field to the payload so the service can filter, or leave a note at the call site if staging-origin DMs are intentional for testing.

dcadenas (approved):
🔑 staging picks up the same mod-relay key wiring as prod.
One non-blocking suggestion on DM side effects inline.

PR #119, opened 2026-06-23
Age Review: case, clock, and enforcement hardening

mbradley:
Part of #110.

Hardens the Age Review case state machine, the 15 day clock, the deadline cron, and
the restrict/clear/deny actions so that an action's effect is verified and its status
reported accurately.

In Divine's apps a restricted account already appears blank (the clients honor the
account's moderation status and labels). Part of this PR closes the remaining path
where the same events were still reachable by querying the relay directly, outside
the apps.

What changed:
- Deadline auto-close acts only on cases that were actually restricted and are
  awaiting a response, not cases that were only reported or already responded (#111).
- Deadlines are evaluated as timestamps, so expiry is determined correctly (#112).
- Case updates and the cron use an optimistic version check (new version column,
  migration 0011), so concurrent actions cannot overwrite each other or double-fire (#113).
- Each enforcement step's real outcome is reported; a failed required step returns a
  non-success result with per-step detail instead of a blanket success (#114).
- Restriction applies the relay's reversible pubkey suspension (hides existing events,
  blocks new writes) and clear reverses it (#115).
- Denial and expiry remove the user's events at the relay at the pubkey level (#116).

Validation (local):
- Default suite: `npm test` (node, self-contained).
- Real D1 (Miniflare): `npm run test:d1` -- before/after coverage of the cron scope,
  deadline comparison, and the version check.
- Live relay (local funnelcake + ClickHouse): `npm run test:e2e` -- publishes an event,
  suspends/bans the pubkey, confirms the relay no longer serves it.
The d1 and e2e suites are opt-in (excluded from the default run) so CI is not gated on
Miniflare or a running relay.

Closes #111, closes #112, closes #113, closes #114, closes #115, closes #116.

### DIFF
--- a/shared/age-review.ts
+++ b/shared/age-review.ts
@@ -96,4 +96,5 @@ export interface AgeReviewCase {
   claim_link_expires_at: string | null;
   created_at: string;
   updated_at: string;
+  version: number;
 }
--- a/src/components/AgeReviewDetail.test.tsx
+++ b/src/components/AgeReviewDetail.test.tsx
@@ -8,6 +8,7 @@ import type { AgeReviewCase, AgeBand, AgeReviewState } from '../../shared/age-re
 const updateAgeReviewCase = vi.fn().mockResolvedValue({ success: true });
 const getAgeReviewConfig = vi.fn().mockResolvedValue({ auto_delete_on_deny: false });
 const writeText = vi.fn().mockResolvedValue(undefined);
+const toast = vi.fn();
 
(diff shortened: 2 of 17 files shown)


NotThatKindOfDrLiz (approved):
Approved after final pass. The PR now surfaces incomplete enforcement in the moderator UI, sends expected_version for case updates, has corrected retry wording, and all required checks are green. Leaving merge to maintainers.

PR #120, opened 2026-06-23
Age Review: reversible withhold for restricted media

mbradley:
Part of #110.

Restricting an account's media used an age-gate state that still serves the bytes to
any signed-in request made directly to the media server (the apps already hide it).
This uses the reversible withhold state instead.

What changed:
- Restriction sends the withhold action, which maps (via the moderation service) to the
  media server's Restricted state: not served to anyone but the owner, and reversible.
- Clear restores the media; denial still deletes.

Companion: a comment correction in divine-moderation-service (separate small PR).

Validation (local): the emitted media action is asserted for restrict and clear; the
media server's own tests confirm the withhold state returns not-found to non-owners
while the age-gate state does not.

Closes #117.

### DIFF
--- a/docs/api-webhooks.md
+++ b/docs/api-webhooks.md
@@ -99,7 +99,7 @@ Moderate media content (images/videos) by SHA-256 hash.
 ```json
 {
   "sha256": "abc123...",
-  "action": "PERMANENT_BAN" | "SAFE" | "AGE_RESTRICTED" | "REVIEW",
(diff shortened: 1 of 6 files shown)


NotThatKindOfDrLiz (approved):
Approved after verifying the reversible withhold change and adding the related media action contract cleanup. Local validation: npx vitest run src/lib/adminApi.test.ts; cd worker && npx vitest run src/bulk-moderate.test.ts; npx tsc -p tsconfig.app.json --noEmit; cd worker && npx tsc -p tsconfig.json --noEmit.

PR #121, opened 2026-06-23
Age Review: page through all events when actioning

mbradley:
Part of #110.

Bulk actioning queried an author's events in a single page and stopped once an account
had more than one page, without surfacing the stop, so a high-volume account could be
actioned over none of its content while the case still showed restricted.

What changed:
- Page through all of the author's events with a created_at cursor (dedup across
  boundaries), with a per-page timeout and a logged (not silent) page bound.

Validation (local): a test that pages through 1200 events across three pages and
collects all of them.

Closes #118.

### DIFF
--- a/worker/src/bulk-moderate.test.ts
+++ b/worker/src/bulk-moderate.test.ts
@@ -155,7 +155,7 @@ function mockPaginatedRelay(all: Array<{ id: string; kind: number; content: stri
   } as unknown as typeof WebSocket));
 }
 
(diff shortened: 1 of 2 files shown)


NotThatKindOfDrLiz (approved):
Reviewed the pagination rewrite and compared it against the prior review. The implementation is bounded, reports partial enumeration honestly, updates the only caller, and the relevant checks pass. Approved.

PR #122, opened 2026-06-23
Age Review: surface enforcement status and conflicts in the moderator UI

mbradley:
The case-update API reports per-leg enforcement status (relay, media/content, account) and returns HTTP 207 when a leg did not apply, plus HTTP 409 with a machine-readable code when a case changed underneath a write. The moderator UI did not yet consume that contract, so a partial enforcement read as full success and a conflict read as an opaque error. This change makes the admin UI reflect what actually happened after an action.

- Surface each enforcement leg. When relay, media/content, or account sign-in fails, show a clear "enforcement incomplete" notice naming what did not apply, instead of treating the update as fully enforced.
- Stop a false positive: the previous account-suspension warning fired even when the account leg was simply not attempted (for example a restricted-to-restricted transition). It now keys off an actual failure.
- Handle the version conflict: a 409 reloads the current state and asks the moderator to review and re-apply, instead of showing "HTTP 409".
- Export `ApiError` with the parsed `code`/`current_version`, parse JSON error bodies in `apiRequest`, and type the enforcement fields on the response.

Independent of the worker hardening PR and back-compatible with a worker that predates the per-leg contract.

Tests: `adminApi` (409 structured fields, 207 returns rather than throws) and `AgeReviewDetail` (207 toast, no toast on ok/not-attempted, 409 reload).

Decision item, not included here: whether to relabel bulk-age-review media (withheld via QUARANTINE) in the media-status badges; the media is withheld either way. Follow-up once the worker change lands: have the client send its known case version for client-side optimistic concurrency.

Part of #110.

Closes #126.

### DIFF
--- a/src/components/AgeReviewDetail.test.tsx
+++ b/src/components/AgeReviewDetail.test.tsx
@@ -148,6 +148,8 @@ describe('AgeReviewDetail', () => {
         variant: 'destructive',
       }));
     });
+    // the toast covers the conflict; the inline error must not also fire for it
+    expect(screen.queryByText(/Failed to update/)).not.toBeInTheDocument();
   });
 
   it('shows Deny & Close without confirmation when auto-delete is off', () => {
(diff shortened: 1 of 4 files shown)


NotThatKindOfDrLiz (approved):
Approved. I reviewed this independently, verified the tests/build, checked the worker contract in #119, and confirmed the main risk is only merge sequencing with #119 rather than a blocker in this PR.

PR #127, opened 2026-06-23
fix(age-review): bump case version in the parent-reply webhook

mbradley:
## Summary

Fix the age-review parent-reply webhook so advancing a case to `submitted_for_review` participates in optimistic concurrency.

The webhook now compare-and-swaps on the version it read, increments `version` on success, and re-reads once after a CAS miss so a legitimate parent reply is not dropped just because an unrelated write happened first.

This PR also keeps the moderator UI aligned with the optimistic-lock contract: case updates send the currently loaded `expected_version`, parse structured 409 responses, and show a reload notice instead of an opaque conflict error.

## Motivation

The webhook previously updated the case state without bumping `version`. A moderator holding the pre-reply version could later PATCH the case with that stale `expected_version`; because the webhook left the version unchanged, the optimistic-lock check could pass and overwrite the parent-reply transition.

## Validation

- `npm run test`
- `cd worker && npm run typecheck`
- `cd worker && npm run test:run`
- `cd worker && npm run test:d1`

## Linked Issue

Closes #129

Related to #110. Surfaced while reviewing #122.

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -45,3 +45,7 @@ jobs:
       - name: Worker tests
         working-directory: worker
         run: npm run test:run
+
+      - name: Worker d1 tests
+        working-directory: worker
+        run: npm run test:d1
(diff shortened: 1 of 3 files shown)


NotThatKindOfDrLiz (approved):
Final pass looks good. The parent-reply webhook now bumps the case version under CAS, the stale moderator write is covered by a real-D1 regression test, and CI now runs the D1 suite so that coverage is enforced. PR metadata now closes #129, and all checks are green.

PR #134, opened 2026-06-25
Fix Ban User modal hanging on "Banning…"

mbradley:
## What

The **Ban User** confirm modal could hang on **"Banning…"** indefinitely while the ban itself succeeded on the relay (reported by Charlie). Two independent defects, both fixed here.

### 1. Audit log no longer blocks the moderation action (`UserActions.tsx`)

The ban mutation awaited `logDecision` (a D1 audit write to `/api/decisions`) in its critical path, so a slow or hung audit write kept the mutation pending and the `ConfirmDialog` could never close. Audit logging is a non-critical side effect, so it's now fire-and-forget via a small `logAudit` helper (with a `.catch`). Applied consistently to all the mutations in this component (ban, suspend, unsuspend, unban, bulk age-restrict, bulk delete). The relay action remains the source of truth and stays awaited.

### 2. Every relay-bound HTTP call now times out (`adminApi.ts`)

`callRelayRpc` was a plain `fetch` with no timeout, so a slow/hung relay (`banpubkey` purges across 16+ tables) left an infinite spinner. I added a shared `fetchWithTimeout` helper with a 30s `AbortSignal.timeout` and applied it to both `callRelayRpc` and `apiRequest`. That second path matters: `bulkModerate` (the "Delete All Content" and "Age Restrict All" actions) routes through `apiRequest`, so without this those destructive modals retained the identical "spins forever" failure mode the ban fix was meant to kill. On timeout it rejects with an actionable message ("The action may still have applied. Re-check before retrying.") that matches the observed reality (the action lands even when the UI gives up). This now protects every RPC and HTTP caller, not just `UserActions`.

## Why this is safe without a reproduction

Both are unambiguous defects in the code path, independent of which call was hanging in the field:
- A non-critical side effect blocking the primary UX violates our graceful-degradation rule.
- No network call should be able to hang forever.
- `UserManagement.tsx` already surfaces relay success immediately and verifies in the background; this brings `UserActions` in line.
- Audit logging is now best-effort: a failed `logDecision` is logged to the console and swallowed, so the decision log can lag the relay with no operator-visible signal. This is the intended graceful-degradation tradeoff (the relay is source of truth); surfacing true state on timeout is tracked in #133.

## Test / checks

- `tsc -p tsconfig.app.json --noEmit` clean
- `eslint` clean
- `vitest run` green (120 passed, 1 skipped; `App.test.tsx` excluded as a pre-existing scaffold smoke test that holds the node process open via the app's relay WebSocket)
- New regression tests: HTTP-path and RPC-path timeouts throw the actionable message and non-timeout errors re-throw unchanged (`adminApi.test.ts`); the ban completes and closes the modal even when `logDecision` never resolves (`UserActions.test.tsx`).

## Out of scope (tracked in #133)

- Same `await logDecision(...)`-in-critical-path pattern in `UserManagement.tsx`, `EventDetail.tsx`, `Labels.tsx`, `EventActions.tsx`.
- Relay-side latency of the purge itself (a queued/async purge if legitimate operations start approaching the 30s client timeout).
- Surfacing true state on timeout via `verifyPubkeyBanned`.

Refs #133

---

## Update — bulk timeout decoupling + accurate read/write copy

Two follow-ups from review, on top of the original ban-modal fix:

- **Decoupled the bulk-moderate timeout from the global 30s.** A large/multi-video account legitimately exceeds 30s, so the flat bound surfaced a false "timed out" on the primary destructive path. `apiRequest` now takes a per-call timeout override; `bulkModerate` uses `BULK_API_TIMEOUT_MS = 180_000`, everything else keeps the 30s default. (This bound is interim — the async job model in #135 removes it.)
- **Read-vs-write timeout copy.** The "may still have applied. Re-check" message now only fires on mutating calls. Reads (`getWorkerInfo`, `list*`, `getbannedevent`, `supportedmethods`) say "Could not reach the relay. Try again." Read-ness is derived from an explicit `READ_RPC_METHODS` set (a `list*` prefix heuristic mislabeled `getbannedevent`/`supportedmethods` as writes).
- Audit-log failure now raises a non-blocking toast; regression tests assert the dialog closes and that a timeout surfaces as the error toast.

Validated locally end-to-end against a real funnelcake relay (ban + delete modals close on success; ban confirmed applied on the relay).

### DIFF
--- a/src/lib/adminApi.ts
+++ b/src/lib/adminApi.ts
@@ -87,8 +87,10 @@ export class ApiError extends Error {
 // Bound every relay-bound HTTP call so a slow or hung relay can't leave the UI
 // spinning forever (the "Banning…" / "Deleting…" confirm modals). Generous
 // because some actions purge across 16+ tables; if legitimate operations exceed
-// this, fix the relay-side latency rather than raise the bound. Covers every
-// caller of apiRequest and callRelayRpc, not just bans.
+// this, fix the relay-side latency rather than raise the bound. Covers callers of
+// apiRequest and callRelayRpc (connection AND body read, via readJsonBounded) plus
+// the standalone check-result / check-classifier / realness fetches — every
+// relay-bound fetch routes through fetchWithTimeout.
 const API_TIMEOUT_MS = 30_000;
 
 // Bulk moderation is a server-side O(N/5) loop that can legitimately run for
@@ -121,6 +123,29 @@ const READ_RPC_METHODS = new Set<string>([
 // even though we stopped waiting (re-check before retrying), but a timed-out
 // read applied nothing (just retry). Emitting "may have applied" on a read would
 // mislead the moderator.
+// Map an AbortSignal.timeout abort to the friendly ApiError copy. Shared by
+// fetchWithTimeout (bounds the connection) and readJsonBounded (the body read): a
+// relay that sends headers then stalls the body aborts during response.json(),
+// AFTER fetch() resolves, so the same mapping has to wrap the read too — otherwise
+// a body stall surfaces as a raw TimeoutError and skips the "may have applied" copy.
+function asTimeoutApiError(err: unknown, label: string, mutates: boolean, timeoutMs: number): unknown {
+  if (err instanceof DOMException && err.name === 'TimeoutError') {
+    const tail = mutates
+      ? 'The action may still have applied. Re-check before retrying.'
+      : 'Could not reach the relay. Try again.';
+    return new ApiError(`${label} timed out after ${timeoutMs / 1000}s. ${tail}`);
+  }
+  return err;
+}
+
+async function readJsonBounded<T>(response: Response, label: string, mutates: boolean, timeoutMs: number): Promise<T> {
+  try {
+    return await response.json() as T;
+  } catch (err) {
+    throw asTimeoutApiError(err, label, mutates, timeoutMs);
+  }
+}
+
 async function fetchWithTimeout(
   url: string,
   init: RequestInit,
@@ -131,13 +156,7 @@ async function fetchWithTimeout(
   try {
     return await fetch(url, { ...init, signal: AbortSignal.timeout(timeoutMs) });
   } catch (err) {
-    if (err instanceof DOMException && err.name === 'TimeoutError') {
-      const tail = opts.mutates
-        ? 'The action may still have applied. Re-check before retrying.'
-        : 'Could not reach the relay. Try again.';
-      throw new ApiError(`${label} timed out after ${timeoutMs / 1000}s. ${tail}`);
-    }
-    throw err;
+    throw asTimeoutApiError(err, label, opts.mutates, timeoutMs);
   }
 }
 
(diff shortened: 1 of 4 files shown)


src/lib/adminApi.ts
```diff
@@ -84,20 +84,78 @@ export class ApiError extends Error {
   }
 }
 
+// Bound every relay-bound HTTP call so a slow or hung relay can't leave the UI
```
dcadenas, line 87: Make this comment match what the code actually bounds.

The comment opens with "Bound every relay-bound HTTP call" but its last line narrows the claim to callers of `apiRequest` and `callRelayRpc`. Four relay-bound entry points still use raw `fetch()` with no timeout: `checkMediaStatus` (432), `getClassifierData` (688), `getAIDetectionResult` (775), and `submitAIDetection` (803). The same hung-backend failure this PR fixes can still hang those panels. Separately, `fetchWithTimeout` wraps only the `fetch()` call, not the `await response.json()` body read in `apiRequest:179` and `callRelayRpc:256`, so a relay that sends headers then stalls the body aborts with a raw `TimeoutError` outside the try/catch and skips the friendly "may have applied" copy.

Pick one scope and make the comment true. Either route those four entry points through `fetchWithTimeout` with the right `mutates` flag (reads for the check/classifier/realness-job calls, write for analyze), which also brings their body read under the same abort handling, or narrow the comment so it only claims to cover `apiRequest`/`callRelayRpc` callers. Reuse `fetchWithTimeout`. Do not add a new abstraction.

dcadenas (dismissed):
This fixes the modal hang well, but detaching the audit write opened a real regression that has to be closed before merge.
For suspend and "Age Restrict All", a report can still show as unresolved after the action succeeds, because those two actions read resolved-state only from the D1 decision row and the refetch now races the detached write.
There is also a smaller follow-up: the new module comment claims every relay-bound call is timeout-bounded, but four raw fetches and the response-body read still are not, so either wrap them or narrow the comment.
Please address both before this lands.

---

- **Suspend and age-restrict reports can stay unresolved after a successful action** ([src/components/ReportDetail.tsx:141](https://github.com/divinevideo/divine-relay-manager/blob/96d88c213e1ba9df257e385c164b1088ec649b13/src/components/ReportDetail.tsx#L141))
  Make report resolution converge after the detached audit write lands.
  
  For suspend and bulk age-restrict, resolved-state comes only from the D1 decision row. `isResolved` here uses live relay state (`isUserBanned`/`isEventDeleted`) for ban and delete, but falls back to `decisionLog.hasDecisions` for the other two:
  
  https://github.com/divinevideo/divine-relay-manager/blob/96d88c213e1ba9df257e385c164b1088ec649b13/src/components/ReportDetail.tsx#L137-L141
  
  `handleActionComplete` invalidates and refetches `['decisions']` synchronously in `onSuccess` (ReportDetail.tsx:329-340), but the audit write is now fire-and-forget in `UserActions.tsx:33-37`. The refetch usually wins the race, so the report reads back before the row exists and shows as still pending. If the POST fails or times out under the new 30s bound, it never resolves. Nothing re-invalidates after the write settles, so the stale view sticks until manual refresh, and a moderator may re-action an already-handled report.
  
  Have `logAudit` return its promise and, on successful persistence, invalidate `['decisions']` (and the target's decision log) so suspend and bulk age-restrict resolve when the row lands. Keep the existing `.catch` toast as the failure path. A report legitimately staying unresolved when no row was written is correct. Do not reintroduce `await` in `mutationFn`. That would bring back the dialog hang this PR fixes. Please add a test where `logDecision` resolves after `onSuccess` and the report still becomes resolved with no manual refresh.

dcadenas (approved):
⏱️ every relay call is time-bounded now, so the ban modal can't hang on a dead RPC.

PR #135, opened 2026-06-26
feat(bulk-moderate): async job model (enqueue + queue consumer + status polling)

mbradley:
## Summary

Makes `/api/bulk-moderate` asynchronous: it now enqueues a job and returns a `jobId` immediately, and the work runs in a Cloudflare Queue consumer. Progress/result land in a `bulk_jobs` D1 table; the UI polls `GET /api/bulk-moderate/status/:jobId` until the job is terminal.

This removes the long-lived request entirely, so a large/multi-video account can no longer hang the moderator's confirm modal (the interim 180s client bound in #134 becomes unnecessary once this lands).

## What's in here

- **Worker:** `runBulkModeration()` (extracted work loop), `ensureBulkJobsTable()`, `handleBulkModerateEnqueue()` (validate → insert pending row → enqueue → return `jobId`), a `queue()` consumer running `processBulkJob()`, and `handleBulkJobStatus()`. Queue producer+consumer bindings added to all three wrangler configs. Age-review enforcement keeps the synchronous path (it needs the result for enforcement leg status).
- **Frontend:** `bulkModerate()` now enqueues and returns the `jobId`; new `getBulkJobStatus()` + `useBulkModerateJob` polling hook; `UserActions` bulk buttons enqueue, show running state, and toast the final counts (or a destructive toast on partial/failed).
- **Hardening (self-review):** terminal state written on any consumer failure; status endpoint self-heals a stale `running` row to `failed` so the poller can't hang; the non-critical decision-log batch is isolated so a D1 audit failure can't mislabel a completed destructive run; enqueue rolls back the orphan row if the queue send fails.

## Dependency / sequencing

This is **stacked on #91** (it reuses the REST media-hash enumeration), so the diff against `main` currently includes #91's commits. Net-new here is the async commits. Once #134 and #91 land, I'll rebase this onto `main` (the #91 commits drop out) and remove the now-moot `BULK_API_TIMEOUT_MS` from #134.

## Validation

- Worker suite green (236), frontend green (122/1 skipped), tsc + eslint clean.
- Full enqueue → queue-drain → status `done` cycle validated locally via Miniflare (status `done`, all hashes QUARANTINE'd).
- Real browser run: the frontend hook enqueued, polled the live status endpoint, and handled the terminal job end-to-end.

## Manual validation plan (staging)

- Age-restrict-all on a real multi-video profile: confirm a `jobId` returns immediately, the UI polls to `done`, and all hashes reach the moderation service.
- Confirm a deliberately-failed job surfaces as `failed` and the UI shows the issues toast.

---

## Merge guidance (read before integrating — cross-branch review)

These three PRs share two frontend files, so the merge needs deliberate 3-way resolution, NOT an "ours/theirs" pick:

- **Order:** land **#134 before #91**. #91 makes the worker enumerate *all* of an author's videos (more work, slower) with no client-side bound; #134 adds the interim client bound. Merging #91 first widens the exact hang window #134 closes. Then #135 last (it already contains #91).
- **`src/components/UserActions.tsx`:** keep #134's `logAudit` fire-and-forget on the four ban/suspend/unban/unsuspend mutations, AND take #135's bulk section (hook + inline `void logDecision().catch()`). Taking #135's file wholesale would silently revert #134's hang fix on ban/suspend.
- **`src/lib/adminApi.ts`:** take #135's `bulkModerate` (enqueue) + `getBulkJobStatus`; keep #134's `fetchWithTimeout` / `API_TIMEOUT_MS` / `READ_RPC_METHODS` / read-write copy / `apiRequest` `timeoutMs` param; remove the now-dead `BULK_API_TIMEOUT_MS` const (nothing references it after async — it won't be flagged by tsc).
- **Test files** (`UserActions.test.tsx`, `adminApi.test.ts`): reconcile both sides' tests; re-run `npx tsc --noEmit` + both vitest suites after.

Confirmed intact in this branch: #91's worker correctness (QUARANTINE-not-AGE_RESTRICTED, paginated `queryRelayEvents`, `deriveFunnelcakeApiUrl` + `AbortSignal` timeout), age-review's synchronous `runBulkModeration` path, and per-env queue isolation (`bulk-moderate-jobs-{local,staging,prod}`).

---

## Update — chunked across queue messages (drains any account size)

Folded in (design + plan committed under `docs/superpowers/`): the consumer now processes a job as **resumable per-page chunks** across queue messages instead of one all-in-one invocation. Each message does one bounded chunk (`MEDIA_CHUNK_SIZE=100`, `EVENT_CHUNK_SIZE=200`), writes incremental counts to the `bulk_jobs` row, then re-enqueues the next chunk (carrying a continuation cursor) or finalizes. `max_batch_size=1` ⇒ one chunk per invocation ⇒ no account size hits the per-invocation subrequest ceiling.

- **Media** enumerates via the funnelcake **v2 cursor** API (`/api/v2/users/{pubkey}/videos`, `next_cursor`) — unbounded, replacing #91's offset+10k-cap.
- **delete-all** chunks both phases: `events` (relay `until`-cursored, ban+kind-5) → then `media`.
- Per-item failures are recorded and the job continues; a thrown infra error → `failed` (partial counts). Stored `failures[]` is capped (50 + "+N more").
- Each chunk refreshes `updated_at`, so an actively-progressing multi-chunk job is never stale-healed.
- `runBulkModeration` (synchronous, age-review enforcement) is kept and shares the per-item helpers; its media enumeration loops the v2 cursor fully. Age-review stays synchronous by design (fails visibly on the case; moderator re-runs from the chunked Users page).

Validated by unit tests (250 videos → exactly 3 messages → all actioned; delete-all events→media transition; infra-error→failed). The full enqueue→multi-chunk→done drive against a local stub is blocked by a post-reboot workerd loopback-fetch regression in this environment (not a code issue); covered by the staging smoke test.

---

## Update — review fixes (`f6af6ba`)

A code-review pass across the three PRs surfaced defects in the new chunked consumer, all now fixed and tested:

- **v2 cursor read from the wrong path (critical, child-safety).** The cursor was read from `body.next_cursor`, but funnelcake's `PaginatedResponse` envelope puts it at `body.pagination.next_cursor` (verified against funnelcake `crates/api/src/handlers.rs`). The always-null cursor silently capped enumeration at the first 100 videos — on the async path AND the synchronous age-review path — and reported success. The test mock encoded the same wrong shape, so it passed green. Fixed the parse path and rewrote the mock to the real `{ data, pagination }` envelope (now load-bearing: the chunking tests fail if the parse path is wrong).
- **Media phase had no termination guard.** A non-advancing or repeating funnelcake cursor would re-enqueue forever, immune to the stale-heal (each chunk refreshes `updated_at`). Now fails closed on a non-advancing cursor and past `VIDEO_MAX_PAGES`, matching the synchronous path's bound.
- **Saturated-second silent under-deletion (events phase).** Stepping strictly past the oldest second on every full page silently dropped events when one second held more than `EVENT_CHUNK_SIZE`, landing `done` with no failure. Now defers the boundary second to the next chunk (no skip, no double-count) and surfaces true single-second saturation so the moderator sees the gap.
- **Pubkey-level zendesk sync** was gated on the final chunk's successes (often an empty short page); now gated on the job's cumulative successes.
- **Async media-only jobs reported `events_processed=0`;** now count one event per video for parity with the synchronous path so the UI's "across N events" stays meaningful.
- **Terminal-failure catch** overwrote per-item failures from earlier chunks; now preserves and appends.
- `.playwright-mcp/` run artifacts gitignored.

Worker suite green (246, +3 new), frontend green (126/1 skipped), tsc + eslint clean.

**Known limitation (documented, deferred):** Cloudflare Queues is at-least-once. A duplicate delivery of the same in-flight chunk message could fork the chunk chain. Every action (banevent/kind-5/moderate-media) is idempotent, so the worst case is count inflation and wasted compute on a rare duplicate — no data corruption. A true fix (an atomic per-chunk claim) is tracked as `TODO(#async-retry)`.


worker/src/bulk-moderate.ts
```diff
@@ -45,90 +222,195 @@ export async function handleBulkModerate(
...
+        : null;
     }
 
-    await runWithConcurrency(successfulEventIds, BULK_ACTION_CONCURRENCY, async (eventId) => {
-      await syncZendeskAfterAction(env, 'delete_event', 'event', eventId, moderatorPubkey);
-    });
-
-    if (successfulEventIds.length > 0) {
-      await syncZendeskAfterAction(env, 'delete_event', 'pubkey', body.pubkey, moderatorPubkey);
+    const status = next ? 'running' : 'done';
```
dcadenas, line 427: Guard every write to `bulk_jobs.status` so a late or duplicate write can't move a finished job backward. Three paths all write status with `WHERE job_id = ?` and no check on the current status:

- the consumer's terminal write (L361-370)
- the consumer's catch write (L380-381)
- the status endpoint's stale-heal (L406-411)

Because none of them assert the status they transition from, and `done`/`failed` are not sticky, the row can flip backward. A stale-heal can flip a still-running job to `failed`, and a chunk that already read it as `running` can then resurrect it. A duplicate queue delivery can flap `done` back to `running` and fire the UI's `onComplete` early while the other branch keeps deleting. The heal also overwrites `failures` with `['job:abandoned ...']`, throwing away the per-item failures the catch block deliberately preserves:

https://github.com/divinevideo/divine-relay-manager/blob/f6af6ba6d2d09e2597591d667d04334bfe76503c/worker/src/bulk-moderate.ts#L373-L381

The fix is small and local:

- add `AND status = 'running'` to the terminal write, read D1 `meta.changes`, and skip `await env.BULK_QUEUE.send(next)` when it changed zero rows
- add `AND status IN ('pending','running')` to the catch write and the stale-heal write so `done`/`failed` are terminal
- make the heal read and append prior failures via the same path the catch block uses, instead of overwriting them

Recovering pages dropped after a heal is separate work (the `TODO(#async-retry)` atomic claim). This change just stops the backward flips.

worker/src/bulk-moderate.ts
```diff
@@ -45,90 +222,195 @@ export async function handleBulkModerate(
...
+  await ensureBulkJobsTable(env.DB);
+  const row = await env.DB.prepare('SELECT * FROM bulk_jobs WHERE job_id = ?').bind(jobId).first<BulkJobRow>();
+  if (!row) {
+    return json({ error: 'job not found' }, 404, corsHeaders);
   }
 
-  result.success = result.failures.length === 0;
-  return json(result, 200, corsHeaders);
+  const job = rowToBulkJob(row);
+  if ((job.status === 'pending' || job.status === 'running') && Date.parse(job.updatedAt) < Date.now() - STALE_JOB_MS) {
```
dcadenas, line 495: Decide what should happen when the stale-heal fires on a job that still has a queued continuation. The heal marks a still-active chunked job `failed` based only on `updated_at` aging past `STALE_JOB_MS`, even when the next chunk is already on the queue. When that continuation is delivered, `processBulkJob` sees the now-terminal row and returns at the short-circuit:

https://github.com/divinevideo/divine-relay-manager/blob/f6af6ba6d2d09e2597591d667d04334bfe76503c/worker/src/bulk-moderate.ts#L293-L296

so the remaining event and media pages are never actioned and the UI shows `failed` on a delete-all that only partly ran. Making `done`/`failed` sticky (the status-guard change above) makes this abandonment correct-by-design but does not recover the dropped pages. Please either confirm and document that a healthy multi-chunk job can never go 30 minutes (`STALE_JOB_MS`) between chunks, so this can't fire on a live job, or have the heal requeue the remaining continuation (ties into the `TODO(#async-retry)` work). Don't leave the silent-drop path undocumented.

worker/src/bulk-moderate.ts
```diff
@@ -45,90 +222,195 @@ export async function handleBulkModerate(
...
+      chunkFailures.push(...media.failures);
+      // Parity with the synchronous path: for media-only actions one video == one
+      // event, so the UI's "across N events" stays meaningful (delete-all counts
+      // events in its own phase, so leave eventsDelta 0 there).
+      if (msg.action !== 'delete-all') eventsDelta = hashes.length;
+      // Guard a non-terminating funnelcake cursor. The sync path is bounded by
+      // VIDEO_MAX_PAGES; the chunked path re-enqueues per page and refreshes
+      // updated_at each chunk, so a cursor that never goes null (or repeats) would
+      // churn the queue forever, immune to the stale-heal. Fail closed instead.
+      if (nextCursor && nextCursor === msg.cursor) {
```
dcadenas: Bound the media phase on pages fetched, not on successful moderations. The guard at L352 trips on `job.mediaProcessed + mediaDelta >= VIDEO_MAX_PAGES * MEDIA_CHUNK_SIZE`, but `mediaProcessed` only counts items that moderated successfully (failures go to `failures[]` with no increment). If the funnelcake cursor keeps advancing through new values while the moderation service is failing, `mediaProcessed` stays near zero and this bound never fires. The frozen-cursor check at L349 only catches a cursor that repeats the exact same value, not one that advances forever or cycles A->B->A. Each chunk refreshes `updated_at`, so the stale-heal can't reclaim it either, and the job re-enqueues itself forever, burning queue and subrequest budget without ever reaching `done`. Carry a page index in `BulkJobMessage`, increment it each media chunk, and throw once it exceeds `VIDEO_MAX_PAGES`, regardless of how many items moderated successfully.

worker/src/bulk-moderate.ts
```diff
@@ -45,90 +222,195 @@ export async function handleBulkModerate(
...
+// second `next`, forking the chunk chain (both branches walk to `done`). The only
+// guard is the terminal short-circuit below, which covers a re-delivered message
+// for an already-finished job but not a mid-run duplicate. There is no data
+// corruption — every action (banevent/kind-5/moderate-media) is idempotent — only
+// count inflation and wasted compute on a rare duplicate. A true fix (an atomic
+// per-chunk claim) is a focused follow-up, not folded in here.
+// TODO(#async-retry): atomic chunk claim to collapse duplicate/forked deliveries.
+const MAX_STORED_FAILURES = 50;
+
+function appendFailures(existing: string[], added: string[]): string[] {
```
dcadenas: Track the dropped-failure count as its own number instead of re-deriving it from the stored list each chunk. `appendFailures` strips the prior `+N more` marker, concatenates the new failures, and recomputes the marker from only the currently-stored set (<= 50). Across chunks this loses earlier overflow counts: chunk 1 with 60 failures stores 50 plus `+10 more`; chunk 2 adding 5 strips the marker, gets 55, and writes `+5 more`, dropping the original overflow. Worse, when a terminal chunk adds zero new failures (your comment notes the last chunk is often an empty short page) and the stored list is exactly 50, `merged.length <= MAX_STORED_FAILURES` holds and the marker is dropped entirely, so a job that hit hundreds of failures finalizes `done` showing exactly 50 and no overflow indicator. On this destructive path `failures[]` is the moderator's only signal of how many items were not actioned, so the count silently understates un-enforced content. Persist a cumulative dropped count in its own column (or a carried-forward marker) so the total survives across chunks and a clean final chunk can't erase it. There is also no test for cumulative `appendFailures` across chunks. Please add one that accumulates more than 50 failures over multiple chunks.

src/hooks/useBulkModerateJob.ts
```diff
@@ -0,0 +1,117 @@
...
+    setJobId(null);
+    setPendingAction(null);
+    setGaveUp(false);
+    notifiedJobId.current = null;
+  }, [pubkey]);
+
+  // Bound the wait: if the job hasn't gone terminal within MAX_POLL_MS, stop.
+  useEffect(() => {
+    if (jobId === null) return;
+    const timer = setTimeout(() => setGaveUp(true), MAX_POLL_MS);
```
dcadenas: Don't re-enable the destructive bulk buttons on a fixed client timer. `MAX_POLL_MS` is 10 minutes; once it fires, `gaveUp` goes true and `isRunning` goes false at L107, which re-enables the buttons. This PR's whole point is chunked jobs that can legitimately run longer than 10 minutes on large accounts, so a still-running delete-all can re-enable its own buttons and let the moderator start a second, duplicate job against the same target. Drive give-up off server-reported state (a terminal status, or the server's stale-heal having fired) instead of a fixed timer, or keep the buttons disabled while a non-terminal job for this pubkey is known to exist.

dcadenas (dismissed):
The async job model is a good shape for this work, but a few paths on this destructive bulk-moderation flow need tightening before merge.
The job status is written by three uncoordinated paths with no "only change it if it's still X" guard, so a finished or failed job can flip backward, and the stale-heal can abandon a job whose next chunk is still queued.
The media-phase runaway guard counts successful moderations instead of pages fetched, so it can loop forever when the upstream is failing.
The failure-overflow count, the 10-minute client give-up timer, and the unguarded enqueue parser each have a concrete gap below. Please address these before merge.

---

- **Enqueue handler parses the body with unguarded request.json(), turning malformed input into a 500** ([worker/src/bulk-moderate.ts:217](https://github.com/divinevideo/divine-relay-manager/blob/f6af6ba6d2d09e2597591d667d04334bfe76503c/worker/src/bulk-moderate.ts#L217))
  Wrap the body parse and return a 400 on malformed input. `await request.json()` here has no try/catch, then you dereference `body.pubkey`. A missing, empty, non-JSON, or non-object body (`null`, `"x"`, `[]`) throws before any field check, and the only handler is the outer route catch, which returns a generic 500. Callers get an opaque 500 where the field checks just below already return clean 400s. The endpoint is admin-auth-gated so exposure is limited, but the wrong status code still misleads logs and alerts.
  
  ```ts
  let body: { pubkey?: string; action?: string; reason?: string };
  try {
    body = await request.json();
  } catch {
    return json({ error: 'Malformed JSON body' }, 400, corsHeaders);
  }
  if (!body || typeof body !== 'object' || Array.isArray(body)) {
    return json({ error: 'Request body must be a JSON object' }, 400, corsHeaders);
  }
  ```

dcadenas (approved):
✅ bulk moderation goes async and chunks per page to stay under the worker budget, ship it.

PR #136, opened 2026-06-26
chore(zendesk): retire dead inbound moderation-execution path (#103)

mbradley:
## Summary

Removes the dead Zendesk → relay-manager inbound moderation-execution path (worker side of #103). Moderation is performed in Relay Manager, not Zendesk; the Zendesk side reports decisions back (auto-solve + internal notes) but must not execute relay actions.

## Motivation

Investigation (recorded in #103) found this path was **live in Zendesk** (active trigger + webhook + `action_requested`/`action_status` fields) and reachable at the worker, but had **never successfully executed** — 0 zendesk-originated rows across the entire prod `moderation_decisions` table. It was inert by accident, not by design, and it bypassed Keycast enforcement + the moderation DM. "Inert by accident" is a latent risk: a secret rotation or a well-meaning fix could silently switch on Zendesk-driven bans that diverge from the main moderation path.

The Zendesk-side surface has already been deactivated (trigger, webhook, both fields — reversible, before/after snapshots captured; see #103). This PR removes the now-dead worker code behind it.

## What changed

**Removed:**
- `handleZendeskWebhook` (`POST /api/zendesk/webhook`) + its route
- `handleZendeskAction` (`POST /api/zendesk/action`) + its route/case
- `updateZendeskTicket` (only caller was `handleZendeskWebhook`)
- Imports orphaned by the removal — `banEvent` / `banPubkey` / `unbanPubkey` (nip86) and `resolveZendeskCreds` (zendesk-sync), confirmed used only by those handlers
- Orphaned `Env` fields `ZENDESK_WEBHOOK_SECRET`, `ZENDESK_FIELD_ACTION_STATUS`, `ZENDESK_FIELD_ACTION_REQUESTED`, and the two field-ID entries in `wrangler.{prod,staging,local}.toml`
- Retirement notes added to `docs/api-webhooks.md`

**Kept (verified independent):**
- The decision-reporting direction: `syncZendeskAfterAction` auto-solve + internal notes
- `parse-report`, `age-review-reply`, `pre-auth`, `mobile-jwt` integrations
- `verifyZendeskWebhook` (still used by `parse-report` and `age-review-reply`)

## Validation

Rebased on `main` @ `234e336` (after #91/#134/#135/#138 landed). Full CI sequence run locally, all green:
- root `npm run test` (tsc-app + eslint + vitest + vite build) — pass
- worker typecheck (`tsc -p tsconfig.json --noEmit`) — clean
- worker `test:run` — 259 passed; `test:d1` — 16 passed

API change only, no UI change. The removed endpoints return 404/401 (retired); no caller exists (mobile/web do not call them; the Zendesk trigger/webhook are deactivated).

Closes #103.

### DIFF
--- a/docs/api-webhooks.md
+++ b/docs/api-webhooks.md
@@ -2,7 +2,7 @@
 
 ## Overview
 
-The Divine Relay Manager exposes several API endpoints that allow external systems to trigger moderation actions. This enables integration with customer support tools like Zendesk, automated moderation pipelines, and other external services.
+The Divine Relay Manager exposes several API endpoints for integration with external systems like Zendesk (ticket report parsing, decision sync, age-review replies). Note: Zendesk does **not** execute moderation actions. Moderation is performed in Relay Manager; Zendesk only receives decision updates back. The former inbound moderation-execution path has been retired (see the note under `POST /api/zendesk/webhook`).
 
 **Base URL:** `https://api-relay.divine.video` (or `https://relay.admin.divine.video/api`)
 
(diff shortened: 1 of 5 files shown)


dcadenas (approved):
🧹 dead inbound path gone, 400 lines lighter and nothing left referencing it.

PR #138, opened 2026-06-29
fix(moderation): ban/delete UI reliability — remove dead kind-5 path + non-blocking audit

mbradley:
## Summary

Makes the ban/delete moderation UI reliably report the truth. Two **coupled** reliability fixes (delete is only correct with both):

1. **Remove the dead NIP-09 kind-5 deletion path.** Admin deletion is NIP-86 `banevent` (authoritative, already in use). funnelcake rejects a non-author's kind-5 (author-only, no admin exception — verified in funnelcake source), so the kind-5 publish always failed and broke every delete flow it sat in: bulk `delete-all` reported total failure despite deleting everything, and both single-event "permanently delete" flows reported failure / skipped media. Removed everywhere — worker `deleteEvents` + `publishKind5Deletion` + the `/api/publish-deletion` route; frontend `publishDeletionRequest` + helper + hook; stale copy.

2. **Make audit-log writes non-blocking** in `EventActions` + `ReportDetail`. They still awaited `logDecision` in the mutation critical path, so a failed `/api/decisions` write reported "Failed to <action>" (and left confirm dialogs stuck) even though the relay/media action already succeeded — the same bug class #134 fixed for `UserActions`. Now fire-and-forget with a post-write `['decisions']` invalidation, mirroring `UserActions`. (`confirmAutoHide` keeps an awaited write — there the decision write *is* the action.)

Closes #137.

## Why it matters
A moderation tool that lies — "delete failed" when the event is already gone — makes moderators re-action, escalate, or lose trust. On child-safety paths, accurate feedback is core to the tool.

## Verification
- Frontend 141 tests (incl. new `EventActions` failure-path regressions), worker 247 tests, tsc + eslint clean.
- **Verified end-to-end on a local funnelcake stack** (real cake relay + funnel API + ClickHouse + Miniflare queue consumer): `/api/moderate delete_event` removes the event and reports success; bulk `delete-all` (events) and `age-restrict-all` (3 × QUARANTINE) + revert (3 × SAFE) complete with accurate reporting; funnelcake confirmed to reject an admin kind-5 (the removal rationale).

## Manual validation plan (staging)
- delete event / permanently delete / delete event + media → success, no spurious failure
- bulk `delete-all` → events removed, reported processed (not all-failures)
- with `/api/decisions` failing → action still reports success + a non-blocking "audit log not recorded" toast


dcadenas (approved):
🧹 Dead kind-5 deletion path gone, deletes now go through NIP-86 only and audit writes no longer block the action.

PR #139, opened 2026-06-29
feat(age-review): Greenlight consent funnel dashboard

mbradley:
## Summary

Adds a v1 Greenlight age-review consent funnel to the relay-manager Age Review tab, plus per-stage tooltips that document the exact criteria behind every count. The funnel shows the whole 13-15 onboarding journey, request through outcome, sourced from the two systems that actually hold the data:

- **Helpdesk side (Zendesk):** intake and the video signal, via tag counts. `age-review` without `age-review-response` (third-party reports), `age-review-response` (parent/teen requests), and `consent_video_received` (the macro shipped 2026-06-29).
- **Moderation side (relay-manager D1, `age_review_cases`):** authoritative outcomes from one band-accurate group-by. Non-terminal states roll up to *In progress*; `cleared` is *Approved* (split into restored vs new minor account by `created_via`); `denied_closed` is *Denied / expired*.

The two sides do not join 1:1 by design. This is a funnel showing drop-off, not a single ledger.

## What's included

- `GET /api/age-review/funnel?age_band=age_13_15` worker endpoint, inside the existing admin-guarded `/api/age-review/*` block (inherits `verifyAdminAccess`).
- Pure `bucketModerationCounts` for state-to-stage bucketing, and `fetchZendeskTagCount` (Zendesk Search `/count.json`).
- Frontend `getAgeReviewFunnel` API client + `useAdminApi` binding.
- `AgeReviewFunnel` panel mounted above the case split-view in the Age Review tab.
- **Per-stage criteria tooltips.** Each card has a hover tooltip with a plain-language meaning and the exact, auditable criteria behind the count (the Zendesk query or the D1 rule). The three Zendesk queries live in one shared `FUNNEL_ZENDESK_QUERIES` constant consumed by both the worker (which counts with them) and the UI (which displays them), so the shown criteria cannot drift from what is measured. The in-progress rule renders from the shared `TERMINAL_STATES` for the same reason.
- Spec and implementation plan imported under `docs/superpowers/` (this repo is now their canonical home since the code lives here).

## Funnel stages and exact criteria

| Stage | Source | Criteria |
|---|---|---|
| Reports in | Zendesk | `type:ticket tags:age-review -tags:age-review-response` |
| Requests in | Zendesk | `type:ticket tags:age-review-response` |
| Video received | Zendesk | `type:ticket tags:consent_video_received` |
| In progress | D1 | band `age_13_15`, state ∉ {`cleared`, `denied_closed`} |
| Approved | D1 | band `age_13_15`, state `cleared` (new = `created_via=minor_onboarding`, restored = the rest) |
| Denied / expired | D1 | band `age_13_15`, state `denied_closed` |

## Criteria verification (against live Zendesk + a seeded local D1)

I ran the full local stack and verified the criteria are correct, not just that they return numbers:

- **Helpdesk partition is clean.** Live Zendesk counts: `337` reports + `44` requests = `381` = total `age-review` tickets. `age-review-response` without `age-review` = `0`. So reports and requests are a clean, non-overlapping partition of all age-review tickets.
- **Tag semantics confirmed by sampling real tickets.** Reports are "Content Report: underageUser" (third-party in-app reports). Requests are parent/teen "account review help" and always also carry `age-review`. The built-in Zendesk discriminator holds.
- **Helpdesk stages genuinely span all ages.** One sampled request was an under-13 ticket, confirming the documented `band_scope: all_bands` caveat. The tooltips and the panel footnote state this explicitly.
- **Moderation counts match the case list.** With representative 13-15 rows seeded into a local D1, the funnel showed In progress `4` / Approved `5` (3 restored, 2 new) / Denied `2`, and the Active case list showed exactly those 4 in-progress cases. An `age_16_plus` row was correctly excluded by the band filter.
- **Explanation == measurement is enforced in code.** A worker test asserts the handler queries Zendesk with the exact `FUNNEL_ZENDESK_QUERIES` strings the tooltips display.

## Graceful degradation

A Zendesk failure (bad creds, timeout, non-200, or Secret Store resolution failure) nulls the helpdesk half and the moderation (D1) half still returns 200. There are explicit no-creds, fetch-failure, and secret-resolution-failure tests asserting this. The UI renders "—" for null helpdesk counts. Verified live: before seeding, the moderation half returned zeros while the helpdesk half still populated from Zendesk.

## Known limitations (v1)

- Helpdesk stages count all age-review tickets, not strictly 13-15 (Zendesk does not tag age band). Flagged in the payload via `band_scope: all_bands`, in each helpdesk tooltip, and in a panel footnote. Moderation stages are band-accurate.
- Video stage is only as complete as macro use / back-fill. It currently reads `0` because the `consent_video_received` macro shipped 2026-06-29 and no tickets are tagged yet; it climbs as the macro is used.

## Out of scope (documented follow-ons)

Zendesk age-band trigger tags, the per-case video join via `zendesk_ticket_id`, admin.divine.video Pulse-card syndication, and the verification webform.

## Test plan

- Frontend `npx vitest run`: green, 12 files, 141 passed, 1 skipped.
- Worker `npx vitest run`: green, 10 files, 259 passed.
- `npx tsc -p tsconfig.app.json --noEmit` and `cd worker && npx tsc -p tsconfig.json --noEmit`: clean.
- `npx eslint .`: clean.
- `npx vite build`: OK.
- Unit coverage: bucketing logic, Zendesk count helper (ok / non-ok / throws), handler with and without Zendesk creds, fetch-failure and Secret Store-resolution graceful degradation, handler queries with the exact shared criteria, and a panel render test.
- Manual: ran the full local stack and confirmed in the Age Review tab that the funnel renders, helpdesk counts populate from live Zendesk, moderation counts match the case list, and the tooltips show the correct per-stage criteria.

No schema change (D1 is maintained at runtime by `ensureSchema`).

Closes divinevideo/support-trust-safety#172.

### DIFF
--- a/docs/superpowers/plans/2026-06-29-greenlight-funnel-dashboard.md
+++ b/docs/superpowers/plans/2026-06-29-greenlight-funnel-dashboard.md
@@ -1,0 +1,621 @@
+# Greenlight Consent Funnel Dashboard v1 Implementation Plan
+
+> **For agentic workers:** REQUIRED SUB-SKILL: Use superpowers:subagent-driven-development (recommended) or superpowers:executing-plans to implement this plan task-by-task. Steps use checkbox (`- [ ]`) syntax for tracking.
+
+**Goal:** Add a `GET /api/age-review/funnel` endpoint and a funnel summary panel to relay-manager, showing the 13-15 Greenlight cohort moving from request through approval, sourced from D1 (moderation outcomes) plus Zendesk tag counts (helpdesk intake).
+
+**Architecture:** A new admin-guarded worker endpoint runs one D1 group-by over `age_review_cases` and a few Zendesk Search counts, buckets them into funnel stages, and returns a small JSON payload. A React panel in the existing Age Review tab renders it. The endpoint is the durable contract that admin.divine.video can syndicate later (out of scope here). All work lives in the `divine-relay-manager` repo.
+
+**Tech Stack:** Cloudflare Workers (TypeScript), D1, Zendesk Search API, React + TanStack Query, shadcn/ui, Vitest.
+
+## Global Constraints
+
+- Repo: `divine-relay-manager`. No new dependencies.
+- D1 schema is maintained at runtime by `ensureSchema`; this feature adds **no** schema change. Never run `wrangler d1 migrations apply`.
+- Reuse existing helpers: `json(body, status, corsHeaders)`, `resolveZendeskCreds(env)`, `apiRequest<T>(apiUrl, path, method)`. Do not reimplement them.
+- Zendesk auth pattern (verbatim from `zendesk-sync.ts`): `btoa(\`${email}/token:${apiToken}\`)`, base `https://${subdomain}.zendesk.com`, header `Authorization: Basic ${auth}`.
+- Graceful degradation: a Zendesk failure must never block the moderation (D1) half of the response.
+- Run `npx eslint .` before any push (tsc + tests are not enough for this repo).
+- Commits: Conventional Commit style, no `Co-Authored-By` lines.
+- Canonical values (verbatim): age bands `under_13` | `age_13_15` | `age_16_plus_claimed`; terminal states `cleared`, `denied_closed`; `created_via` of `report` | `minor_onboarding`. `AGE_BANDS` and `TERMINAL_STATES` are exported from `shared/age-review.ts`.
+
+---
+
+### Task 1: Funnel types and moderation bucketing
+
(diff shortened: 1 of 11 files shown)


NotThatKindOfDrLiz (approved):
Approved after final pass. I pushed f4a74bb to make the Zendesk half fully best-effort, including Secret Store credential-resolution failures, and added regression coverage for that path. Checks are green and the PR body now closes divinevideo/support-trust-safety#172.

PR #150, opened 2026-07-08
fix(age-review): emit snake_case restriction status so the app renders the review screen

mbradley:
## What and why

`/v1/account/moderation-status` emits `restriction.status` as camelCase `restrictedMinorReview`, but the mobile client only accepts snake_case `restricted_minor_review` and falls back to `active` otherwise, so the in-app age-review screen has never rendered for a restricted account. This emits `restricted_minor_review`, matching the client's existing expectation and the endpoint's own (already snake_case) nested fields.

Server-side by design: it reaches already-shipped app builds on deploy; a client change would be App Store gated and wouldn't reach affected users. Not a regression, the two sides shipped with mismatched casing (#67 worker, #3531 mobile) and never matched. Full background in the issue.

## Changes

- `worker/src/age-review.ts`: emit `restricted_minor_review`.
- `worker/src/age-review.test.ts`: assert the snake_case value and lock the rest of the wire contract (`state`, `suspectedAgeBand`, `allowedResolution`) so it can't drift again.

## Testing

- `npm run typecheck` clean
- `npm run test:run` 265 passed
- `npm run test:d1` 16 passed
- Confirmed `divine-web` does not consume this endpoint; `divine-mobile` is the sole consumer and already expects snake_case.

## Residual (follow-up, not in this PR)

Clients predating the age-review screens (#3531) have no screen to render, so some users still need manual outreach. Separate from this contract fix.

Closes #149

### DIFF
--- a/worker/src/age-review.test.ts
+++ b/worker/src/age-review.test.ts
@@ -802,7 +802,7 @@ describe('handleGetModerationStatus', () => {
     expect(body.restriction.status).toBe('active');
   });
 
-  it('returns restrictedMinorReview when active case exists', async () => {
+  it('returns restricted_minor_review when active case exists', async () => {
     const c = makeCase({ state: 'restricted_pending_user_response' });
     const db = createMockDb([c]);
     // Override first() to return the case for the pubkey query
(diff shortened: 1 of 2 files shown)


dcadenas (approved):
✅ snake_case status is pinned

PR #152, opened 2026-07-08
fix(age-review): route underage reports into the flow and guard generic suspend/unsuspend

mbradley:
## What and why

An `NS-underageUser` report shows up in the report queue with the same generic Suspend / Unsuspend / Ban panel every report gets. Using the generic **Suspend** there enforces at the relay but never advances the age-review case, so the case is orphaned (stays `open_reported`), the app reports the user `active`, media is hidden but not age-restricted, and no tracking ticket is created. The generic **Unsuspend** has no age-case check, so it can lift a suspected-minor hold with no verification. This routes moderators into the Age Review flow and guards the generic actions so enforcement can't drift from the case.

Closes #151. Independent of the casing fix (#149 / PR #150).

## How it works

- **Report hand-off:** on an `NS-underageUser` report (that exact category, not the CSAM/high-priority bundle), the Suspend/Unsuspend controls are replaced with "Handle in Age Review", which deep-links to the case, and the bulk content actions (Age Restrict All / Delete All Content) are hidden — content enforcement runs through the case. Ban stays, with an evidence-preservation warning when an open case exists. If `?pubkey=` resolves to no active case (e.g. an already-cleared verified minor), the moderator gets an explicit empty state, not a blank panel.
- **Worker guard:** a bare `suspendpubkey`/`unsuspendpubkey` on a pubkey with an open (non-terminal) age-review case is refused with a structured `age_review_active` response (caseId + state). The same refusal now covers `/api/bulk-moderate` (bulk age-restrict / un-age-restrict / delete-all), so the Users tab and direct API calls can't drift content enforcement from the case either; both guards share one helper so the contract can't diverge. It informs and routes; it does not auto-translate the action into a case transition. Scoped to the interactive path — age-review's own enforcement calls the nip86 helpers and `runBulkModeration` directly, and internal moderation callers use `ban*`/`unban*` only. The lookup fails open on a transient DB error (logs + proceeds) so it can't block core moderation. The bulk check is enqueue-time only: a case opened while an already-legitimate chunked job is draining does not abort it (documented in the route comment).
- **Guard-response routing:** any context (Users tab, a non-underage report on an account that also has an open case) that hits either guard — suspend/unsuspend or a bulk action — routes the moderator to the case instead of showing a raw error.
- **Deep-link:** case selection is URL-driven (`?case=`); `?pubkey=` resolves to the active case (from the loaded list, or a by-pubkey fallback fetch for a just-created case), with a fallback fetch so a case outside the current filter still opens.

## Commits (each independently reviewable)

1. read-only by-pubkey active-case lookup endpoint
2. worker guard on bare suspend/unsuspend
3. `NS-underageUser` report hand-off
4. URL-driven case selection + deep-link resolution
5. route guard-blocked suspend/unsuspend to the case
6. warn before banning an account under age review
7. review follow-up: guard fails open on DB error; validate pubkey before the DB check; terminal-state test coverage
8. review follow-up: explicit empty state for `?pubkey=` with no active case; polish (shared category constant, param encoding, ban-lookup gating, named response type)
9. review follow-up: route bulk content actions through the case — hide the bulk buttons on an underage report, refuse `/api/bulk-moderate` on an open case (shared fail-open guard helper), and route the refusal to the case
10. review follow-up: pin the bulk-guard contract edges (409 `state` field, Delete All confirm-dialog path, documented accepted guard edges)

## Testing

- Worker: typecheck clean; `test:run` 279 passed; `test:d1` 16 passed (endpoint + guard + terminal-state + bulk-guard tests).
- Frontend: tsc clean; eslint clean; vitest 157 passed; `vite build` clean (hand-off, guard-routing, Ban-warning, bulk-hide, bulk-guard-routing tests).
- Three picky code-review passes (backend + frontend + the bulk-guard follow-up), all findings resolved.

## Out of scope

- In-app screen casing fix (#149 / PR #150).
- `AGE_REVIEW_RESTRICTED` user DM backstop (deferred).
- Moderator-initiable parent outreach (dropped; user self-serve is intentional).
- Mid-job re-check for a case opened while a chunked bulk job is draining (accepted edge, documented at the route).
- Lowercasing/normalizing a mixed-case pubkey before the guard lookup (pre-existing behavior; worth verifying relay-side case handling separately).

### DIFF
--- a/src/components/UserActions.tsx
+++ b/src/components/UserActions.tsx
@@ -31,12 +31,17 @@ export function UserActions({
   const api = useAdminApi();
   const queryClient = useQueryClient();
   const navigate = useNavigate();
-  const showBulkActions = context !== 'age-review';
 
   // An under-16 report must be worked through the Age Review flow, not the
   // generic Suspend/Unsuspend (which would enforce without advancing the case).
   // Exact category only: CSAM/child-safety are a separate, non-reversible path.
   const isUnderageReport = context === 'report' && reportCategory === UNDERAGE_REPORT_CATEGORY;
+
+  // Bulk content actions are hidden in the age-review context (the case screen
+  // has its own enforcement controls) and for an underage report (content
+  // enforcement runs through the case; the /api/bulk-moderate guard refuses it
+  // on an open case regardless). Ban stays as the severe-action escape hatch.
+  const showBulkActions = context !== 'age-review' && !isUnderageReport;
 
   // The worker guard refuses a bare suspend/unsuspend on an account under age
   // review (any context). Route the moderator to the case instead of surfacing
@@ -197,6 +202,10 @@ export function UserActions({
       onActionComplete?.();
     },
     onError: (error) => {
+      // A guard-refused enqueue (open age-review case) routes to the case, for
+      // the contexts where the buttons still render (Users tab, non-underage
+      // report on an account that also has an open case).
+      if (routeToAgeReviewIfGuarded(error)) return;
       // Covers both enqueue failure and a persistent status-poll failure
       // (the job may have started; error.message carries the specific reason).
       toast({ title: 'Bulk action failed', description: error.message, variant: 'destructive' });
@@ -229,7 +238,7 @@ export function UserActions({
                   Handle in Age Review
                 </Button>
               </TooltipTrigger>
-              <TooltipContent><p>This account was reported as under 16. Suspend and unsuspend run through the Age Review flow so the case, content age-restriction, and deadline stay in sync.</p></TooltipContent>
+              <TooltipContent><p>This account was reported as under 16. Account and content enforcement run through the Age Review flow so the case, content age-restriction, and deadline stay in sync.</p></TooltipContent>
             </Tooltip>
           ) : isSuspended ? (
             <Tooltip>
--- a/src/components/UserActions.tsx
+++ b/src/components/UserActions.tsx
@@ -5,11 +5,13 @@ import { useToast } from '@/hooks/useToast';
 import { useAdminApi } from '@/hooks/useAdminApi';
 import { useBulkModerateJob } from '@/hooks/useBulkModerateJob';
 import { ConfirmDialog } from './ConfirmDialog';
-import { UserX, UserCheck, ShieldAlert, Trash2, Pause, Play } from 'lucide-react';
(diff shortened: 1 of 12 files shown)


src/components/UserActions.tsx
```diff
@@ -180,7 +220,18 @@ export function UserActions({
...
+          {isUnderageReport ? (
+            <Tooltip>
+              <TooltipTrigger asChild>
+                <Button variant="outline" className="border-blue-500 text-blue-600 hover:bg-blue-50"
+                  onClick={() => navigate(`/age-review?pubkey=${encodeURIComponent(pubkey)}`)} disabled={anyPending}>
+                  <ArrowRight className="h-4 w-4 mr-1" />
+                  Handle in Age Review
+                </Button>
+              </TooltipTrigger>
+              <TooltipContent><p>This account was reported as under 16. Suspend and unsuspend run through the Age Review flow so the case, content age-restriction, and deadline stay in sync.</p></TooltipContent>
```
dcadenas: Hide the generic bulk content actions when `isUnderageReport` is true.
This branch adds `Handle in Age Review`, but `showBulkActions` still lets `Age Restrict All` and `Delete All Content` render for an `NS-underageUser` report.
Those buttons call `/api/bulk-moderate` directly, so a moderator can change content while the age-review case still has no matching enforcement transition.
Keep the hand-off and Ban escape hatch, but route content enforcement through Age Review.

dcadenas (changes requested):
The underage report path still leaves a bypass around the Age Review case workflow. Close that path before merge so underage enforcement state and case state stay aligned.

src/components/UserActions.tsx
mbradley: resolved in the last two commits. the bulk buttons are hidden for an NS-underageUser report (hand-off + ban escape hatch stay), and i closed the path at the endpoint too: /api/bulk-moderate now refuses any bulk action on a pubkey with an open case, with the same structured 409 the suspend guard uses — so the users tab and direct calls can't drift content enforcement either, and the frontend routes the refusal to the case. both guards share one fail-open helper. age review's own enforcement calls runBulkModeration directly, so it's unaffected. one accepted edge, documented at the route: the check is enqueue-time only, so a case opened while an already-running chunked job drains doesn't abort it.

dcadenas (approved):
✅ underage review routing is clean

PR #142, opened 2026-07-01
feat(minor-safety): T&S protected-minor visibility (#141)

mbradley:
## Summary

- Add `GET /api/account-status/:pubkey` behind the existing admin gate to surface keycast's durable `verified_minor` and `verified_minor_at` fields.
- Add the frontend admin API method and an age-review detail query that shows an approved protected-minor badge when keycast confirms the account is an approved protected minor.
- Show an explicit protected-minor status unavailable message when the status check fails, and keep the account-status cache isolated by selected environment.

## Motivation

Moderators need to distinguish a suspected 13-15 case from an account that has already completed the protected-minor approval flow. This PR surfaces only the server-backed `verified_minor` fact and intentionally defers any active-protection display until those enforcement features exist.

## Linked issue

Closes #141.

Follow-on: #143 tracks displaying active protections after the underlying enforcement work exists.

## Manual validation

- `npx vitest run src/components/AgeReviewDetail.test.tsx src/lib/adminApi.test.ts`
- `cd worker && npx vitest run src/account-status.test.ts src/index.test.ts`
- `npx tsc --noEmit`
- `npx vite build`

PR #153, opened 2026-07-08
feat(minor-safety): surface protected-minor protections in the age-review view

mbradley:
## What and why

Closes #143. Part of the protected-minor safeguards epic (divinevideo/support-trust-safety#173).

The age-review case view shows the "Approved protected minor (13-15)" badge (#141, PR #142) but not what that approval actually protects the teen from. #141 deliberately left that out: the protections did not exist yet, and listing them would have given moderators false assurance. The content lock (support-trust-safety#175) has since shipped and the DM restriction (support-trust-safety#176) is in review on both clients, so this adds an honest display of the protections that apply.

## What it shows

When `verified_minor` is true, a small block renders under the badge row, split into two clearly separated sections so shipped protections are never mixed with rollout-only ones:

**Protections that apply to this account** (shipped only)

* Adult content lock: adult content is hidden and the 18+ visibility toggle is disabled, so they cannot re-enable it.
* Qualifier: derived from the approved protected-minor status above; enforced client-side by the Divine apps on supported versions. Not confirmed per device.

**Rolling out (not yet enforced by released apps)** (separate section below a divider, muted styling)

* DM restriction: DMs limited to the pinned official accounts, Divine HQ (`_@divinehq.divine.video`) and Divine Moderation (`moderation@divine.video`); everything else is blocked on send and hidden on receive.

Nothing else changes: the badge, the approval date, the "protected-minor status unavailable" fail-safe, and the non-minor empty state behave exactly as before, and the block never renders unless `verified_minor` is confirmed true.

## Design choice: policy-derived display

The protections are client-enforced and not per-device server-observable (NIP-17 gift-wraps are not relay-attributable; the content lock is in-app). Per the issue's recommended default, the display is policy-derived from `verified_minor`, and the copy is framed as "what policy applies to this account, enforced by the apps," never "confirmed active on their device." A reported-state design (clients reporting applied protections) was considered in the issue and deferred as a larger mechanism.

**Shipped vs rollout-only separation (from review):** the DM restriction is merged in no released client yet (divine-mobile#5754 and divine-web#474 are both open), and an inline caveat under the applied-protections heading still presents it as part of the current protection set. So the DM restriction now lives in its own "Rolling out (not yet enforced by released apps)" section, visually and semantically separate from the applied protections, with a `TODO(divinevideo/support-trust-safety#176)` in the code to move it into the applied section once the client releases ship. Tests assert the DOM separation, so the DM row cannot silently migrate back under the applied heading.

## Testing

* Structural tests pin the shipped/rollout separation: the applied section (queried via its heading) must contain the adult-lock copy and must NOT contain the DM restriction; the rollout section's heading must carry the not-yet-enforced framing and contain the DM row (NIP-05 handles asserted). Verified non-vacuous: moving the DM row back under the applied heading fails the test.
* Framing test pins the policy-derived copy ("enforced client-side by the Divine apps", "not confirmed per device"); absence tests cover `verified_minor` false and status-unavailable (both sections absent), awaiting query settlement (`isFetching() === 0`) so they cannot pass vacuously against a still-loading tree.
* Full gate: component tests 25/25, frontend suite 153 passed, eslint clean, tsc clean, vite build clean.
* Adversarial review loop run to an explicit CLEAN verdict on each round (initial, post-review restructure); all findings resolved (shipped/rollout split, NIP-05 handles instead of impersonable display names, h4 headings for screen-reader navigation, vacuous-pass test windows closed, anchored queries).

## Screenshots

No screenshot: rendering this state requires the full local stack plus a keycast-verified minor fixture. The exact rendered copy is quoted verbatim above; the block is a bordered `text-xs` card in the case header, matching the existing header idiom.

## Out of scope

* Any worker/D1 change (the existing `/api/account-status` endpoint already supplies `verified_minor`).
* Per-device/reported-state protection telemetry (issue lists it as the larger alternative).
* Parent-approved DM allowlist (future work under the epic).

### DIFF
--- a/src/components/AgeReviewDetail.tsx
+++ b/src/components/AgeReviewDetail.tsx
@@ -26,6 +26,9 @@ import {
   AlertTriangle,
   Loader2,
   ExternalLink,
+  ShieldCheck,
+  EyeOff,
+  MessageSquareLock,
 } from "lucide-react";
 import { CopyableId } from "@/components/CopyableId";
 import { UserIdentifier } from "@/components/UserIdentifier";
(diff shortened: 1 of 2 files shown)


src/components/AgeReviewDetail.tsx
```diff
@@ -204,6 +207,41 @@ export function AgeReviewDetail({ caseData: c }: Props) {
...
+              </div>
+              {/* TODO(divinevideo/support-trust-safety#176): drop the
+                  "rolling out" caveat once the DM restriction ships in
+                  released mobile + web builds. Listing it without the caveat
+                  before then would assert a protection no released client
+                  enforces — the false assurance this block must never give. */}
+              <div className="flex items-start gap-1.5 text-muted-foreground">
+                <MessageSquareLock className="h-3.5 w-3.5 mt-0.5 shrink-0" />
+                <span>
+                  DM restriction (rolling out; not yet in released app versions): DMs limited to the pinned official accounts, Divine HQ (_@divinehq.divine.video) and Divine Moderation (moderation@divine.video); everything else is blocked on send and hidden on receive.
```
dcadenas: Move the DM restriction out of this applied-protections block, or omit it until released clients enforce it.
The section is headed `Protections that apply to this account`, so this row still presents a rollout-only safeguard as part of the current protection set even though it says `not yet in released app versions`.
Keep the adult-content lock here and update the test that calls this the applicable protections list so it no longer pins the unreleased DM restriction under that behavior.

dcadenas (changes requested):
The protected-minor safety section needs to separate shipped protections from rollout-only protections before this can merge. The current presentation can make moderators rely on a DM restriction that released clients do not enforce yet.

src/components/AgeReviewDetail.tsx
mbradley: restructured: the applied-protections block now contains only the shipped adult-content lock (plus the policy qualifier), and the DM restriction moved into its own "Rolling out (not yet enforced by released apps)" section below a divider. the tests pin the DOM separation — the applied section must not contain the DM restriction — so it can't drift back under that heading before the client releases ship. the code TODO now says to move it up when #176 ships.

dcadenas (approved):
✅ minor protections read cleanly

PR #155, opened 2026-07-08
fix(zendesk): route resolved tickets to a group instead of the credential owner

mbradley:
## What and why

When relay-manager resolves (solves) a Zendesk ticket, it assigned the ticket to the API credential's email, so every moderation/age-review ticket it auto-resolves landed on a personal admin account rather than a team queue. This instance permits solving without an individual assignee, so this drops the personal assignee and routes resolved tickets to a configurable group (`ZENDESK_GROUP_ID`), set to the **Trust & Safety** queue in prod.

Closes #154.

## Changes

- `worker/src/zendesk-sync.ts` (`addZendeskInternalNote` solve branch): drop `assignee_email = creds.email`; set `group_id` from `ZENDESK_GROUP_ID`.
- `worker/src/age-review.ts` (`syncAgeReviewTicketResolution`): same.
- `worker/wrangler.prod.toml`: `ZENDESK_GROUP_ID = "15225535020687"` (Trust & Safety group).
- Tests in both suites assert group routing and no personal assignee.

## Behavior

- With `ZENDESK_GROUP_ID` set: resolved tickets solve **unassigned, into that group's queue**.
- Unset (e.g. staging today): solve unassigned with no group — safe.

## Testing

- `npm run typecheck` clean
- `npm run test:run` 265 passed
- `npm run test:d1` 16 passed

## Out of scope / follow-up

- Bulk cleanup of the existing backlog of tickets already assigned to the credential owner (a separate Zendesk bulk update).
- Staging can be pointed at the appropriate group id if desired.

### DIFF
--- a/worker/src/age-review.test.ts
+++ b/worker/src/age-review.test.ts
@@ -286,6 +286,7 @@ describe('handleUpdateAgeReviewCase', () => {
       ZENDESK_SUBDOMAIN: 'test',
       ZENDESK_API_TOKEN: 'tok',
       ZENDESK_EMAIL: 'agent@test.com',
+      ZENDESK_GROUP_ID: '15225535020687',
     }), corsHeaders);
     expect(res.status).toBe(200);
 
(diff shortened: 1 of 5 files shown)


dcadenas (approved):
✅ resolved tickets route to the right group

PR #157, opened 2026-07-09
fix(reports): surface reported-user comments and activity drill-down on report detail

mbradley:
Closes #156

## What

a report about an account whose only activity was comment spam showed an empty "Recent Content on Relay" section, with no way to reach the reported user's activity from the report. this connects existing patterns to the reports context — no new views — then hardens the paths it opened. follow-ups spun out: #158 (ErrorBoundary), #159 (BannedUserCard sibling gap), both assigned and in progress.

### the fix (#156)

- **`useUserStats`**: adds `1111` (NIP-22 comments) and `6`/`16` (reposts) to the recent-posts author query. the report page's `UserProfileCard` already had the comment badge rendering (shipped in #47); it just never received comments. zero extra round-trips — same parallel relay subscription, same limit/timeout bounds.
- **`ReportDetail`**: the reported user gets the same in-app drill-down the reporter rows already have — "View Activity" → `/events?pubkey=<reported>` (encoded: the p tag is attacker-authored, unlike the signature-verified reporter pubkey).
- **`EventsList`**: `Comments (1111)` and `Generic Reposts (16)` kind-filter presets.

### hardening + correctness that fell out of self-review (5 adversarial rounds)

- **new `src/lib/nip18.ts`** — shared `isRepostKind` / `parseRepostedEvent` / `getRepostTargetId`, defensively typed with **full element-level tag validation**: hostile inner-repost JSON like `tags:[["imeta",42]]` previously reached MediaPreview's `part.startsWith()` and would white-screen the admin app on exactly the page used to review the hostile account. consumed by `UserProfileCard`, `useUserSummary`, and `useThread` (replacing its local `REPOST_KINDS`).
- **badge chain made exhaustive**: kinds 21/22 were mislabeled "Text note (kind 1) — not visible in Divine apps" (moderation-relevant misstatement); now Video 21/22/34235/34236, Comment 1111, Repost via `isRepostKind`, Note for kind 1 only, `getKindName()` fallback otherwise. kind 20 Picture added to kindNames (verified against NIP-68).
- **repost rendering**: labeled inner content instead of raw NIP-18 JSON, media preview uses the inner event's content+tags, full inner pubkey in the attribution tooltip, empty-content reposts identify their target event id.
- **AI summary integrity**: `/api/summarize-user` payload now carries `kind`; the worker labels comments/reposts in the prompt (extracted `formatRecentPostLine`, backward compatible with kind-less payloads) and receives inner repost text instead of 200 chars of serialized-event hex.
[...]

PR #161, opened 2026-07-09
fix(ui): error boundaries so a render exception can't white-screen the admin UI

mbradley:
Closes #158

## What

follow-up out of #157's review loop: there was no React ErrorBoundary anywhere in src/, so any uncaught render exception unmounted the entire app — white screen, no recovery except reload. the report surfaces render attacker-authored event data, which made that class of crash a moderation-blocking failure on exactly the pages used to review hostile accounts. #157 closed one concrete vector (hostile inner-repost JSON reaching MediaPreview); this closes the class's blast radius.

- **`ErrorBoundary`** (new) — dependency-free class component (React has no hook equivalent). `fallback` takes a node or the render-prop form `(reset) => node`; `resetKeys` clears a caught error when the selection changes. error state is a boolean rather than the thrown value, so a hostile `throw null` can't trick the boundary into re-rendering the crashing subtree.
- **app root** — boundary wraps everything in `App.tsx`, outermost on purpose so the fallback can't depend on a provider that may itself have crashed. fallback: a centered card — warning icon, "Something went wrong", one line of guidance, and a Reload page button.
- **report detail** — the desktop pane and mobile sheet share one `reportDetailPane` element (deduplicated from two verbatim copies) inside a scoped boundary. a crashing report degrades to an inline fallback while the reports list stays usable; selecting a different report, re-clicking the same row, and the fallback's Try again all retry (the same-row click goes through a boundary reset ref, since resetKeys can't see a same-id selection).
- **`ReportDetailErrorFallback`** (new) — degraded mode is an action path, not a dead end: an inline card in the detail pane showing "This report failed to render" with the full untruncated report id, target event id, and target pubkey in a monospace block, plus Try again and Dismiss buttons. target ids go through validated extraction (the tags are reporter-authored — possibly the very garbage that crashed the render); a non-canonical report id renders verbatim, because a case-folded copy would silently fail relay/D1/log lookups; non-strings are dropped — the crash fallback must never crash itself.
- **reports list** — also boundary-wrapped: a hostile list row degrades to an inline "The reports list failed to render" card with a retry button, in the same grid cell, instead of escalating to the app-root card whose reload would deterministically re-crash.
- **`isHex64` / `getReportTargetIds`** in `src/lib/constants.ts` — shared and validated (junk or valueless tags can't mask a later valid id). TODO(#160) tracks consolidating the three legacy extraction copies and inline hex regexes onto shared helpers.
- **`vite.config.ts`** — excludes `**/.claude/worktrees/**` from vitest collection. found during this PR's own development: a session worktree under `.claude/worktrees/` breaks `npm run test` in the main clone (collection descends into the worktree's `worker/test/*.d1.test.ts` and fails resolving miniflare, and `@/` imports resolve against the wrong src/). verified empirically in both directions.

## Design notes

- during review i migrated the queue's `getReportTarget` onto the shared helper and reverted it: presence-based extraction is load-bearing for queue visibility (a malformed report must stay triageable), and `useReportContext`/`ReportDetail` carry identical copies the detail pane relies on. consolidating all four sites with agreed semantics is #160, deliberately not this PR.
- no moderation action buttons inside the crash fallback: identifiers + the surviving list are the action path; pulling action machinery into the degraded path risks the fallback itself throwing.
- `window.location.reload()` isn't click-tested — jsdom's location is unforgeable; covered by the browser verification below.
- the list-pane wrap has no dedicated crash test (row components live inside Reports.tsx and can't be module-mocked); the boundary component itself is heavily tested and the wiring is declarative.

## Verification

- tsc clean, eslint clean, 187 frontend tests pass (25 new). mutation-verified four ways: a reference-only resetKeys comparator, a deleted componentDidCatch, a removed boundary wrap, and a removed reset-ref line each fail the suite.
- `Reports.test.tsx` integration test locks the acceptance wiring: mocked always-throwing ReportDetail + stubbed worker API → inline fallback, surviving list, both retry gestures, full identifiers.
- exercised end-to-end in a real browser with temporary render throws: a crashing report degrades to the inline fallback (identifiers populated, Try again and Dismiss both working, deep-link URL cleared on dismiss) with the reports list fully interactive beside it; an app-level crash renders the root error card instead of a blank page, and its reload button reloads. the fallback markup is small and fully visible in the diff.
- worktree exclusion verified empirically: a fake nested worktree is not collected; all tests still list cleanly.

### DIFF
--- a/src/components/ReportDetailErrorFallback.tsx
+++ b/src/components/ReportDetailErrorFallback.tsx
@@ -1,0 +1,64 @@
+// ABOUTME: Degraded-mode UI for a report whose detail pane crashed during render
+// ABOUTME: (hostile event data): full target identifiers + retry/dismiss (#158)
+
+import { AlertTriangle, RotateCcw, X } from 'lucide-react';
+import { Button } from '@/components/ui/button';
+import { getReportTargetIds, isHex64 } from '@/lib/constants';
+import type { NostrEvent } from '@nostrify/nostrify';
+
+interface ReportDetailErrorFallbackProps {
+  /** The selected report event — it rendered in the list, so showing its
+   * fields is useful; IdentifierRow still guards every value individually. */
+  report: NostrEvent | null;
+  onRetry: () => void;
+  onDismiss: () => void;
+}
+
+// Target ids arrive pre-validated by getReportTargetIds; report.id is
+// whatever the worker returned (normally relay-verified, but this is the
+// degraded path — trust nothing). Valid hex lowercases to NIP-01 canonical
+// form; anything else renders VERBATIM, because a case-folded copy of a
+// non-canonical id would silently fail relay/D1/log lookups. Non-strings
+// are dropped — the crash fallback must never crash itself.
+function IdentifierRow({ label, value }: { label: string; value: string | undefined }) {
+  if (typeof value !== 'string') return null;
+  return (
+    <p className="text-xs font-mono break-all text-left">
+      <span className="text-muted-foreground select-none">{label}: </span>
+      {isHex64(value) ? value.toLowerCase() : value}
+    </p>
+  );
+}
+
+export function ReportDetailErrorFallback({ report, onRetry, onDismiss }: ReportDetailErrorFallbackProps) {
+  const targetIds = report ? getReportTargetIds(report) : null;
+  return (
+    <div className="flex h-full flex-col items-center justify-center space-y-3 p-8 text-center">
+      <AlertTriangle className="h-6 w-6 text-destructive" />
+      <p className="font-medium">This report failed to render</p>
+      <p className="text-sm text-muted-foreground">
+        It may contain malformed event data. The identifiers below are from the
+        report event itself; the crash details are in the browser console.
+      </p>
+      {report && (
+        <div className="w-full max-w-xl space-y-1 rounded border bg-muted/30 p-3">
+          <IdentifierRow label="report id" value={report.id} />
+          <IdentifierRow label="target event" value={targetIds?.eventId} />
+          <IdentifierRow label="target pubkey" value={targetIds?.pubkey} />
+        </div>
+      )}
+      <div className="flex items-center gap-2">
+        <Button variant="outline" size="sm" onClick={onRetry}>
+          <RotateCcw className="mr-1 h-3 w-3" />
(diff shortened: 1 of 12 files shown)


src/components/ReportDetailErrorFallback.tsx
```diff
@@ -0,0 +1,64 @@
...
+  return (
+    <p className="text-xs font-mono break-all text-left">
+      <span className="text-muted-foreground select-none">{label}: </span>
+      {isHex64(value) ? value.toLowerCase() : value}
+    </p>
+  );
+}
+
+export function ReportDetailErrorFallback({ report, onRetry, onDismiss }: ReportDetailErrorFallbackProps) {
+  const targetIds = report ? getReportTargetIds(report) : null;
```
dcadenas, line 34: Normalize report `tags` before this fallback and the reports list read them.
This fallback calls `getReportTargetIds(report)`, and the parent reports page also builds consolidated and filtered reports before the list error handling can help.
Both paths still assume `event.tags` is an array, while `fetchReports()` treats raw `/api/reports` payloads as `NostrEvent[]`.
Add a small guard in `fetchReports()` or `/api/reports`, then cover missing, null, and non-array `tags` through both this fallback and the parent reports list path.

dcadenas (changes requested):
The scoped crash handling still leaves malformed report event data able to escape the new UI containment. Add a shared shape guard before merge so hostile or malformed reports cannot push moderators back to the app-level reload screen.

src/components/ReportDetailErrorFallback.tsx
mbradley: resolved in a45a18b. `fetchReports`/`fetchResolutionLabels` now normalize the payload at ingestion — non-object events dropped, tags validated to `string[][]` with the same element-level filter `parseRepostedEvent` uses — and `getReportTargetIds` tolerates non-array tags directly, so the fallback can't crash itself even if something bypasses ingestion. covered at all three layers: adminApi unit tests (missing/null/non-array/junk-member tags), the fallback component test, and the reports-list integration test, where a `tags: null` report now leaves the queue intact and stays reachable in the individual view.

dcadenas (approved):
🛡️ one bad render can no longer white-screen the whole admin UI, ship it.

PR #163, opened 2026-07-09
fix(users): banned-user card surfaces all moderation-relevant kinds, not just text notes

mbradley:
Closes #159

## What

sibling of #156, surfaced during #157's review: `BannedUserCard`'s recent-activity query asked the relay for `kinds: [1]` only, so a banned account whose activity was comments (1111), reposts (6/16), or videos showed "No posts found on this relay" — the same blind spot #157 fixed on the report page, here on the surface where a moderator reviews an account that's *already been actioned*.

- **shared kinds list** — new `RECENT_CONTENT_KINDS` in `src/lib/constants.ts` (video kinds derived from a new `VIDEO_KINDS` in `lib/kindNames`), consumed by `useUserStats` and `BannedUserCard` so the surfaces can't drift. TODO(#162) tracks the third, still-drifted list in `UserProfilePreview` — aligning it needs the shared rendering pieces too, so it's deliberately its own change.
- **shared `KindBadge`** — #157's badge chain extracted verbatim from `UserProfileCard` into its own component, consumed by both surfaces; #157's existing badge-mapping tests guard the extraction.
- **NIP-18 display semantics centralized** — new `parseRepostForDisplay()` in `lib/nip18.ts`, consumed by `BannedUserCard`, `UserProfileCard`, and `useUserSummary` (the AI-summary payload). one place now decides what text a recent-content row shows:
  - reposts render the inner event's content labeled "reposted:", never raw NIP-18 JSON
  - the inner kind is resolved from the parsed JSON or the NIP-18 `k` tag, so kind-1064 base64 (draft NIP-95 file data) can't be smuggled through a repost — and direct 1064 events are suppressed the same way on both cards (the report page had this latent too)
  - addressable-video reposts (a-tag coordinate, empty content — the NIP-18 latest-version pattern, and Divine's primary content kinds) identify their target coordinate; previously they rendered nothing
  - out-of-spec reposts (plain text in content) render that text instead of silently discarding moderation evidence
  - reposter-authored e-tag values only render when they are well-formed 64-hex event ids, and the target line is clamped — hostile tags can't balloon the card
- **query correctness** — rows sort newest-first on a copy before taking 3 (relay order is unguaranteed; mutation-verified test); `staleTime` on the content query since UserManagement remounts one card per banned/suspended user on every tab switch; profile now comes from the shared `useAuthor` hook (Funnelcake REST fast path, validated metadata) instead of a hand-rolled query.
- count reads "N events on relay", header "Recent Content on This Relay", per-user aria-labels on the icon buttons.

## Known tradeoffs

- `useAuthor` has no request timeout on its REST fast path (pre-existing hook behavior shared by every consumer; the old hand-rolled query had a 3s abort). adopting the shared hook keeps this card consistent with the rest of the app; a timeout belongs in the hook itself if we want one.
- the repost-row JSX is similar-but-not-identical between the two cards (different clamp/sizing); extracting a shared row component is part of #162's shared-rendering work, where a third consumer forces the shape.

## Verification

- tsc/eslint clean, 198 frontend tests pass (24 new: 13 nip18 helper/derivation tests incl. the smuggling, a-tag, out-of-spec, and hostile-e-tag cases; 10 BannedUserCard tests; repost-of-1064 parity test on UserProfileCard). the newest-3 ordering test is mutation-verified — an inverted sort comparator fails the suite.
- verified against the live production relay in a browser (banned list stubbed browser-locally; no real ban involved): an account whose only activity is kind-1111 comments, which showed "No posts found" before this change, renders its comments newest-first with Comment badges and a correct event count. no screenshot attached — the capture shows a real user's identity under a "Banned Users" heading and this repo is public.

### DIFF
--- a/src/lib/nip18.ts
+++ b/src/lib/nip18.ts
@@ -1,8 +1,6 @@
 // ABOUTME: NIP-18 repost helpers — kind 6/16 events carry the reposted event
 // ABOUTME: as stringified JSON in their content field (or empty content)
 
-import { hasDisplayableContent } from '@/lib/kindNames';
-
 // Local, not exported: PR #161 adds a shared isHex64 to lib/constants and
 // #160 tracks consolidating hex-id validation — this stays private until
 // that lands to avoid two exported definitions colliding at merge time.
@@ -16,19 +14,11 @@ export interface RepostedEvent {
   content: string;
   tags: string[][];
   pubkey: string;
-  /** Present when the inner JSON carried a numeric kind. */
-  kind?: number;
 }
 
 /** The reposted event's id from a repost's `e` tag (NIP-18 MUST for kind 6). */
 export function getRepostTargetId(tags: string[][]): string | undefined {
   return tags.find(t => t[0] === 'e')?.[1];
-}
-
-/** The reposted event's kind from a kind-16 repost's `k` tag (NIP-18 SHOULD). */
-export function getRepostKind(tags: string[][]): number | undefined {
-  const value = tags.find(t => t[0] === 'k')?.[1];
-  return value !== undefined && /^\d+$/.test(value) ? Number(value) : undefined;
 }
 
 /**
@@ -78,7 +68,6 @@ export function parseRepostedEvent(content: string): RepostedEvent | null {
         ? o.tags.filter((t): t is string[] => Array.isArray(t) && t.every(x => typeof x === 'string'))
         : [],
       pubkey: typeof o.pubkey === 'string' ? o.pubkey : '',
-      ...(typeof o.kind === 'number' ? { kind: o.kind } : {}),
     };
   } catch {
     return null;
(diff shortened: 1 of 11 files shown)


src/lib/nip18.ts
```diff
@@ -36,8 +78,58 @@ export function parseRepostedEvent(content: string): RepostedEvent | null {
...
+    const suppressed = !hasDisplayableContent(event.kind);
+    return {
+      isRepost: false,
+      inner: null,
+      displayContent: suppressed ? '' : event.content,
+      contentSuppressed: suppressed,
+    };
+  }
+  const inner = parseRepostedEvent(event.content);
+  const innerKind = inner?.kind ?? getRepostKind(event.tags);
```
dcadenas: Decide whether to blank a repost using the wrapped event's own kind, not the `k` tag the reposter wrote.

The account rendered on these cards is the one under review, and it authors its own `k` tag. If it reposts readable text through a kind-16 wrapper but tags it `['k','1064']`, and the content does not parse to an inner event (plain text, or JSON that omits the inner `kind`), `innerKind` falls back to `1064`, `suppressed` becomes true, and the text is thrown away. The card shows an empty row and `useUserSummary` sends `[reposted unknown target]` to the summarizer. Before this change that inner text reached the AI, so the account now looks cleaner during review than it really is.

Keying off `inner?.kind` alone loses no protection. A genuine kind-1064 repost carries `"kind":1064` inside its JSON, so it is still suppressed:

https://github.com/divinevideo/divine-relay-manager/blob/71c8a0f6126b36a5c4de79d7f1538627e0875837/src/lib/nip18.test.ts#L124-L130

The `k`-tag branch only fires when the content did not parse to an inner event, and there is no decoded file data there to protect.

```suggestion
  const innerKind = inner?.kind;
```

Keep the `k`/`a`/`e` tags for the target label only.

There is a second, smaller leak on the next line. When content is valid JSON whose inner `content` field is not a string, `parseRepostedEvent` returns null, and with no `k` tag `displayContent` falls back to the raw `event.content`. That renders the JSON envelope as `reposted: {json}` and forwards it to the AI, against the never-raw-JSON intent. Render bounded out-of-spec text instead of the raw blob.

Two test updates go with this: change the `viaKTag` expectation at nip18.test.ts:132-137 to assert the text now renders, and add a case for valid JSON with a non-string `content` field and no `k` tag.

dcadenas (changes requested):
This narrows the moderation evidence a moderator and the AI risk summary get to see, so it needs a fix before merge. A repost's display text is blanked using a tag the reposter wrote, which means the account under review can hide its own repost text from the card and the AI summary with one forged tag. Please decide suppression from the wrapped event's own kind, and stop the raw JSON fallback, before this merges.

src/lib/nip18.ts
mbradley: You're right that blanking on a reposter-authored signal hides evidence. Moving it to the inner JSON kind wouldn't fix it, since that's equally reposter-authored and the forge-to-hide hole just relocates there. So I traced why the suppression existed at all: it only ever fired to hide base64 from kind 1064, which is a defunct NIP-95 draft, not in the current spec, with zero events on our relay (Divine media lives on Blossom as URLs, not inline base64). So I removed 1064 from the query and deleted the suppression outright (4292cd3).

`parseRepostForDisplay` no longer blanks on any claimed kind, so a reposter can't hide its own readable text from the card or the AI by any tag. Your point 2 stays: a serialized-event object envelope renders as the target label, not raw JSON. I also loosened it so JSON arrays and plain text still show (those were being over-blanked), so readable evidence is never dropped.

One divergence I want to flag: you'd suggested keeping genuine-1064 suppression via the inner kind, and I dropped it. There are no genuine 1064 events to suppress and any "1064" claim is unverifiable, so keeping the mechanism only reopens the forge-to-hide surface. If you'd rather we defensively suppress content that structurally looks like a base64 blob regardless of kind, I can add that, but it's a heuristic with its own tradeoffs so I left it out.

Separately: this surfaced that `/api/summarize-user` interpolates post content into the prompt with no delimiting or injection guard, and the worker trusts the model's raw JSON output. Pre-existing and orthogonal, filed as #169 (scoped to output validation, with the injection residual acknowledged) rather than widening this PR.

dcadenas (approved):
✅ banned-user card now surfaces every moderation-relevant kind, and dropping the forgeable repost suppression is the right call.

PR #165, opened 2026-07-09
feat(reports): surface NIP-22 comment context on report surfaces (#164)

mbradley:
Closes #164

## What

surface NIP-22 comment context on the moderation surfaces, so a moderator can tell at a glance whether a comment-spam account sprayed across many videos vs talked in one thread, and can see the comments left on a reported video. motivated by a real T&S report where the report page showed a comment-spammer's activity as blank because the surfaces only queried kind-1 notes.

two independent improvements, both in this PR (they answer different questions and live in different sections).

## A — recent-content context ("what has this account been doing, and where")

on the reported user's recent content (`UserProfileCard`, `BannedUserCard`, and the kind-1111 rows in the Events list):

- **per-card spray roll-up**: an at-a-glance line, "8 comments across 7 videos" (spray) vs "8 comments on 1 video" (conversation), with a note when the same text hit multiple targets. computed offline from the comments' own tags — the noun ("video"/"post") is driven by the NIP-22 `K` root-kind tag.
- **per-row "on \<parent\>" link**: each comment row identifies what it's attached to, resolving the parent's title via one batched relay lookup and degrading to the id/coordinate when unavailable. the link lands the moderator on the parent **inside the tool** (see the decision below), never on the public page.

## B — reported-content context ("what's happening on this specific video")

- the report's thread section now lists the kind-1111 comments on the reported event, newest-first, with the reported account's own rows flagged (this landed in the two foundation commits).
- the "Full Thread" modal (`ThreadModal`) now loads NIP-22 comments too (previously kind-1 only, so Divine comments never appeared) and nests them correctly: it roots addressable-video reports properly and attaches each comment under its immediate NIP-22 parent (`e`/`a`), with orphans hanging under root.

## Key decision — internal event view for the parent link

the parent link routes to the internal Events tab (`/events?event=<nevent|naddr>`) rather than the public divine.video page. reason: the Events tab's direct-event lookup already falls back to the `getbannedevent` RPC, so it can show a **removed or banned** parent — exactly the content a moderator most needs to see, and the case the public page 404s on. this reuses the existing in-tool event viewer and the established `?pubkey=` URL-param pattern.

`EventsList` also parses `naddr` and resolves an address coordinate (`kind`+`author`+`#d`), so internal nav works for addressable-video parents, not just event-id parents.

**contract, narrowed in review:** banned-parent retrieval works for event-id targets only. `getbannedevent` takes an id and the relay has no by-coordinate banned lookup, so an naddr target gone from the live relay lands on a labeled miss. in practice this is the rare path: `getCommentTarget` prefers the `E` root tag, both Divine clients always write one, and a production sample found 257/257 comments carry `E` (zero A-only). divinevideo/divine-funnelcake#649 tracks a by-coordinate banned lookup for the residual strict-NIP-22 case.

## Verification

- ~50 new tests; full frontend suite passes (290 pass, 1 skipped, 25 files), `tsc -p tsconfig.app.json --noEmit` clean (bare `tsc --noEmit` is a false green here, fixed in AGENTS.md in this PR), `vite build` clean, eslint clean.
- the pure logic is TDD'd directly: comment-target parse/canonicalize/encode/title extraction, batched resolution + graceful degradation, the roll-up line, `parseSearchInput` naddr, and `buildThreadTree` NIP-22 nesting incl. disagreeing e/a tags, uppercase hex, orphans, and the NIP-10 no-regression case.
- `EventsList` render tests cover the internal-nav runtime paths: the naddr lookup filter shape, `?event=` consumption (the param leaves the URL so a later clear/re-search sticks), the banned id-fallback, the labeled address-mode miss, and that only surviving rows feed the batched title query.
- reviewer feedback plus six fresh-context review rounds drove the review-response commits; the final pass surfaced nothing.

## Manual validation plan

1. open a report on a comment-spam account: the card shows the spray roll-up, and each comment row shows an "on \<parent\>" link.
2. click a parent link: land on the Events tab showing that video/post in the detail pane. a removed or banned parent resolves via the admin fallback when the link carries an event id (the common case, both Divine clients tag the root event id); an naddr link whose target is gone shows the labeled miss instead.
3. after following a link, clear the search / run a different search: it sticks (the `?event=` param does not revert it).
4. open a video-level report and the "Full Thread" modal: the kind-1111 comments on that video appear, nested, with the reported user's rows flagged.

## Out of scope

- the public divine.video event page showing a comment's parent (divine-web, separate repo).
- archival of already-scrubbed content (separate discussion).


src/components/EventsList.tsx
```diff
@@ -502,34 +551,43 @@ export function EventsList({ relayUrl }: EventsListProps) {
...
+        try {
+          const bannedEvent = await callRelayRpc<NostrEvent>('getbannedevent', [searchMode.hex]);
+          if (bannedEvent) return { event: bannedEvent, banned: true };
+        } catch {
+          // Not banned or RPC failed
+        }
+        return null;
+      }
+
+      if (searchMode.type === 'address') {
```
dcadenas, line 515: The internal "on <parent>" link was chosen so a moderator can still see a parent after it is banned or removed. This address branch drops that ability.

The event id branch falls back to `getbannedevent` when the relay no longer serves the event:

https://github.com/divinevideo/divine-relay-manager/blob/c01dcff5e2d649820ac937aa4edbfa472c607533/src/components/EventsList.tsx#L558-L570

The address branch at 573-580 only queries the live relay and returns null when the event is gone. Divine's videos are addressable content, so most parents route through here. A comment on a taken-down video then links to "No event found on relay", which is the dead-end the internal link was built to avoid.

There is no clean in-repo fix. `getbannedevent` needs an event id this branch does not have, and `listbannedevents` returns no coordinate to match on. Decide the scope: either narrow the contract and tests so naddr links are not claimed to show banned parents, or file a relay-side follow-up for a by-coordinate banned-event lookup.

src/components/BannedUserCard.tsx
```diff
@@ -77,6 +82,12 @@ export function BannedUserCard({ pubkey: rawPubkey, reason, onUnban, actionButto
     staleTime: 2 * 60_000,
   });
 
+  // #164 A: spray roll-up over the full fetched set + parent-link resolution
+  const allEvents = postStats?.allEvents ?? [];
+  const commentTargets = allEvents.map(getCommentTarget).filter((t): t is string => !!t);
+  const { titles } = useEventTitles(commentTargets, undefined);
```
dcadenas: Scope this to what actually renders, and defer it until the card opens.

`useEventTitles` is fed from the full fetched set (up to 50 events), but its only consumer renders parent links for at most 3 recent comment rows, shown only when the card is expanded. There is no open guard either, so a collapsed card fires a parent-title relay query on mount and resolves up to ~47 targets that can never render.

Build `commentTargets` from the rendered `recentPosts` rows, and gate resolution on the card being open. `EventsList` already scopes resolution this way at 766-770, so follow that pattern. `UserProfileCard` has the same over-scope, so fix it there too.

The cost is bounded: id targets collapse into a single filter, and the query is cached and deduped. So this is a cleanup rather than a visible bug. It is worth doing to keep the three components consistent and stop the off-screen relay work.

dcadenas (changes requested):
Two things need to be settled before this can merge.
The internal parent link is meant to still show a parent after it is banned or removed, but that only works for parents identified by an event id.
Addressable video is the common parent shape, and for those a removed parent lands on "No event found on relay", so decide whether to narrow that promise or track a relay-side lookup.
Separately, the parent-title resolution runs too wide and too early, so scope it to the rows that actually render and only once the card is open.

src/components/EventsList.tsx
mbradley: validated this against both clients and the relay before deciding. the gap is real: the address branch has no banned fallback, and funnelcake's get_banned_event is an id-only primary-key read with no by-coordinate lookup. but the promise holds on the operative path. getCommentTarget prefers the E root tag, both our clients always write one (mobile unconditionally, web for addressable roots too), and a prod sample of 257 kind-1111 events found all 257 carrying E, zero A-only. so real parents route through the id branch, which does show banned and removed parents.

resolved by doing both halves of what you laid out: the contract is narrowed in-repo (the address-mode miss message says banned events resolve by event id only, the branch comment documents why no fallback exists, and the spec's contract section matches), and divinevideo/divine-funnelcake#649 tracks a by-coordinate banned lookup to close the residual gap for strict-NIP-22 clients that write A-only roots.

src/components/BannedUserCard.tsx
mbradley: fixed as outlined: targets come from the rendered rows and resolution waits for the card to open. BannedUserCard gates on isOpen and scopes to its 3 rendered rows, UserProfileCard gates on the section being expanded and widens on show-all, and tests pin the gating by recording what the hook is asked to resolve.

the same sweep caught EventsList feeding pre-filter events to the batched query (and keeping the list query hot in direct-lookup mode), so it now scopes to the filtered rows or the lookup card. it also surfaced that the caches this PR touches weren't environment-keyed, so event-titles, event-search, user-posts-stats, and full-thread now carry the env in their query keys.

dcadenas (approved):
🧵 clearer comment context across report surfaces

PR #171, opened 2026-07-10
fix(summarize): validate and clamp /api/summarize-user model output (#169)

mbradley:
Closes #169

## What

the `/api/summarize-user` handler trusted the model's raw response: it regex-grabbed the first `{...}`, `JSON.parse`d it, then cached (KV, 1h) and returned it with no schema check and no `riskLevel` enum-clamp. untrusted post content is interpolated into the claude-3-haiku prompt, so a malformed or injection-nudged reply flowed straight to the moderator-facing `<AISummary>` card.

low-severity by design: the summary is advisory display only, shown beside the actual reported content, nothing auto-actions off `riskLevel`, and the card already degrades an unrecognized level to "unknown". worst realistic case was a misleading advisory line a moderator can discount. surfaced while reviewing #163.

## Approach

- new `normalizeUserSummary()` is the single validation gate on the model output. it keeps only `{ summary, riskLevel }`, clamps an out-of-enum/missing/non-string `riskLevel` to `'unknown'`, trims the summary, and rejects non-object or blank-summary output as malformed (`null`).
- both call sites use it: the fresh-generate path throws to the graceful fallback (no caching) when validation fails, and the cache-read path re-validates every entry so the guard also covers pre-#169 leftovers, not just fresh writes.
- KV read and write are each wrapped in try/catch. the cache is a non-critical optimization, so a KV failure degrades to regeneration (read) or still returns the fresh summary (write) rather than collapsing a healthy summary into the error card.
- the untrusted user content in the prompt is wrapped in a `<user_data>` block with a treat-as-data instruction.

## Residual risk (accepted)

prompt injection can't be fully prevented while untrusted post content shares the model's context. the delimiting lowers the odds but content can still forge the closing delimiter. accepted because the summary is advisory and reviewed alongside the source content, per the issue. a more current model would lower the success rate if this is revisited.

## Verification

- worker `tsc --noEmit` clean; full worker suite passes (313 tests). 17 tests cover this change: 12 unit tests on `normalizeUserSummary` (every enum level, missing/non-string/out-of-enum clamp, casing+whitespace normalize, trim, extra-key stripping, malformed rejection, idempotency) and 5 handler-integration tests on cache behavior (serves valid cache, re-validates+clamps a drifting cached entry, regenerates on unparseable/failed-normalization cache, degrades when the read throws, still returns when the write throws).
- no visual change: the response contract is unchanged for well-formed output; only malformed output now falls back instead of returning garbage.

### DIFF
--- a/worker/src/index.test.ts
+++ b/worker/src/index.test.ts
@@ -790,3 +790,117 @@ describe('bulk-moderate age-review guard', () => {
     expect(send).not.toHaveBeenCalled();
   });
 });
+
+describe('summarize-user cache validation', () => {
+  const FALLBACK_SUMMARY = 'Unable to analyze user behavior at this time.';
+
+  function makeSummarizeEnv(
+    cached: string | null,
+    opts: { anthropicKey?: string; getThrows?: boolean; putThrows?: boolean } = {},
+  ) {
+    const kv = {
+      get: vi.fn(async () => {
+        if (opts.getThrows) throw new Error('KV read down');
+        return cached;
+      }),
+      put: vi.fn(async () => {
+        if (opts.putThrows) throw new Error('KV write down');
+      }),
+    };
+    const env = {
+      ALLOWED_ORIGINS: 'https://app.divine.video',
+      RELAY_URL: 'wss://relay.divine.video',
+      ADMIN_API_KEY: 'test-admin-key',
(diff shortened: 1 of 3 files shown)


dcadenas (approved):
🛡️ model output gets validated, clamped, and trimmed before it ever hits the cache.

PR #176, opened 2026-07-13
feat(minor-safety): keycast actor contract for status-change attribution

mbradley:
## what

Adds the relay-manager half of the keycast#295 contract: keycast-client's `suspendUser` / `unsuspendUser` / `banUser` can attach an optional `actor` (moderator hex pubkey) to the status PUT, so keycast can write a durable `admin_audit_events` row for the change. A malformed actor is dropped with a warning (keycast 400s on it, which would fail the whole status change) and keycast falls back to log-only. Mirrors the merged `clearVerifiedMinor` actor guard.

## why this is contract-only

The first cut also wired the interactive age-review path: the UI sent a logged-in `moderator_pubkey`, the PATCH persisted it, and it fed the keycast status legs. Review correctly held that a client-declared pubkey must not populate a durable audit actor without server-side identity binding — and that binding is not reachable from here:

- `verifyAdminAccess` proves the caller MAY act (CF Access JWT present, or a shared admin key); it derives no identity.
- CF Access carries an email, not a Nostr pubkey, and there is no email→pubkey map.
- No login surface is mounted, so the UI has no verified moderator pubkey to send.

So attribution needs the identity plane tracked in #178. This PR is narrowed to the contract half; #178 wires it end-to-end with a verified actor. The age-review PATCH + UI are reverted to main (no behavior change there): the status legs call keycast actor-less, as before.

## relationship to #175

#175 is only fully resolved once a verified actor flows, so it stays open and is closed by #178. This PR lands the contract #178 builds on — hence "Relates to #175".

## verification

- worker: `tsc --noEmit` clean; `vitest run` 320 pass (contract tests: valid actor forwarded on suspend/ban/unsuspend; malformed/uppercase/wrong-length dropped with a warn; omitted-actor log-only fallback).
- frontend: full `npm run test` green (tsc + eslint + vitest 290 + build); UI unchanged from main.
- forward-compatible before keycast#295 deploys: keycast ignores the unknown `actor` field until the contract lands (`SetUserStatusRequest` has no `deny_unknown_fields` — verified in the earlier review round; worth a re-confirm if keycast has moved).

## no visual change

UI reverted to main; no rendered UI differs.

### DIFF
--- a/worker/src/age-review.ts
+++ b/worker/src/age-review.ts
@@ -436,13 +436,16 @@ export async function handleUpdateAgeReviewCase(
       console.error(`[age-review] Bulk action failed for case ${caseId}:`, error);
     }
 
-    // actor = the case's assigned moderator, attributing BOTH keycast legs below
-    // to the human who acted: the status change (durable admin_audit_events row,
-    // keycast#279 / #175) and the verified_minor clear (#147). Best-effort:
-    // relay-manager auths with a shared admin pubkey, so there's no per-actor
-    // signal, and moderator_pubkey is unvalidated on write. A malformed/absent
-    // actor is dropped client-side (keycast 400s on it) → keycast logs-only.
-    const moderatorActor = (updated?.moderator_pubkey ?? existing.moderator_pubkey) ?? undefined;
+    // actor = the case's moderator as of THIS write, attributing BOTH keycast
+    // legs below to the human who acted: the status change (durable
+    // admin_audit_events row, keycast#279 / #175) and the verified_minor clear
+    // (#147). The UI sends the logged-in moderator with state changes and the
+    // PATCH boundary validates canonical 64-hex, so post-#175 rows are clean;
+    // the client-side drop guard remains for rows written before validation.
+    // Reading the updated row (not a stale ?? chain) means an unassign in the
+    // same PATCH is honored rather than attributing the previous moderator.
+    // An absent actor degrades to keycast's log-only audit fallback.
+    const moderatorActor = (updated ?? existing).moderator_pubkey ?? undefined;
 
     // Keycast account status.
     const keycastLeg = await runStatusLeg('Keycast', () =>
--- a/worker/src/age-review.ts
+++ b/worker/src/age-review.ts
@@ -20,7 +20,7 @@ import {
 import { runBulkModeration, type BulkModerateEnv } from './bulk-moderate';
 import { resolveZendeskCreds } from './zendesk-sync';
 import type { BulkAction } from '../../shared/bulk-moderation';
-import { suspendUser, unsuspendUser, banUser, clearVerifiedMinor, createMinorAccount, type KeycastEnv } from './keycast-client';
+import { suspendUser, unsuspendUser, banUser, clearVerifiedMinor, createMinorAccount, type KeycastEnv, HEX_64 } from './keycast-client';
 import { suspendPubkey, unsuspendPubkey, banPubkey, type SecretStoreSecret } from './nip86';
 
 export interface AgeReviewEnv extends BulkModerateEnv, KeycastEnv {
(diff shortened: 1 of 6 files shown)


worker/src/age-review.ts
```diff
@@ -429,11 +429,19 @@ export async function handleUpdateAgeReviewCase(
       console.error(`[age-review] Bulk action failed for case ${caseId}:`, error);
     }
 
+    // actor = the case's assigned moderator, attributing BOTH keycast legs below
+    // to the human who acted: the status change (durable admin_audit_events row,
+    // keycast#279 / #175) and the verified_minor clear (#147). Best-effort:
+    // relay-manager auths with a shared admin pubkey, so there's no per-actor
+    // signal, and moderator_pubkey is unvalidated on write. A malformed/absent
+    // actor is dropped client-side (keycast 400s on it) → keycast logs-only.
+    const moderatorActor = (updated?.moderator_pubkey ?? existing.moderator_pubkey) ?? undefined;
```
dcadenas: At `worker/src/age-review.ts:438`, populate `moderatorActor` from a trusted authenticated moderator identity before calling the Keycast status methods. Report-created cases leave `moderator_pubkey` null, and `AgeReviewDetail` state-changing PATCHes do not send an actor, so this expression normally yields `undefined` and Keycast drops the actor while enforcement still succeeds. Validate the canonical 64-hex pubkey and persist it in the same compare-and-swap update as the status transition. Add an end-to-end test that starts with a report-created case and exercises Restrict, Clear, or Deny through the real UI path, asserting that the same moderator is stored and sent to Keycast.

dcadenas (dismissed):
The new Keycast audit attribution is not wired into the normal interactive age-review path. Before merging, establish a trustworthy authenticated moderator identity so Restrict, Clear, and Deny actions persist and send the acting moderator instead of silently producing unattributed audit records.

worker/src/age-review.ts
mbradley: implemented per your mechanics: the ui sends the logged-in moderator with state changes, the PATCH boundary validates a canonical 64-hex pubkey (400 on malformed, null still unassigns) and persists it in the same compare-and-swap update as the transition, and an end-to-end test drives exactly your scenario: a report-created case with no moderator, one PATCH, asserting the same moderator is stored in the CAS binds and sent to keycast. a same-PATCH unassign is honored now too instead of attributing the previous moderator.

this feedback also forced back up implementing a real login for proper attribution of moderation actions: this has still always leaned on CF Access options, so attribution here (and in the rest of the tool) can't populate until that ships. there's a stale issue from before to do this via divine-login; i'm closing it in favor of a fresh one (#178) in follow-up to this PR and can keep the plates spinning to see it through. the open question in #178, and one you'll have a view on: divine-login (nostr pubkey attribution, fits admin_audit_events.actor_pubkey natively) or google workspace OIDC per the identity-plane split sketched in iac#814 (email attribution, which would need an email-to-pubkey mapping for the keycast audit shape). this PR ships the contract half either way: attribution flows the moment the login lands, with no further changes here.

worker/src/age-review.ts
```diff
@@ -240,10 +240,17 @@ export async function handleUpdateAgeReviewCase(
     binds.push(newDeadline);
   }
 
-  // Moderator assignment
+  // Moderator assignment/attribution. The value is UI-declared within the
+  // CF Access perimeter (house pattern, same trust model as
+  // moderation_decisions.moderator_pubkey) and feeds the keycast audit actor
+  // (#175) -- reject malformed input loudly at the boundary rather than
+  // letting the audit silently degrade to log-only downstream.
   if (body.moderator_pubkey !== undefined) {
```
dcadenas, line 244: Derive the moderator actor from verified request identity before persisting or forwarding it. The current admin gate proves that the caller may perform the action, but it does not bind this editable JSON field to that caller, so an authorized request can record any canonical pubkey and a normal UI action can remain unattributed. Add a routed test that authenticates a real UI action, verifies the persisted row and Keycast actor, and rejects a mismatched body pubkey.

dcadenas (changes requested):
The age-review handler still accepts moderator_pubkey from request JSON without proving it belongs to the authenticated moderator. Bind the actor to verified per-request identity, reject mismatches, and cover the real authenticated UI path before merging.

worker/src/age-review.ts
mbradley: you're right that the actor has to be bound to a verified identity, and that isn't reachable from here — verifyAdminAccess proves access, not identity (CF Access carries an email, not a pubkey, and there's no email->pubkey map), and no login surface is mounted, so the UI has no verified pubkey to send. that identity plane is #178.

so i've narrowed this PR to the half that's safe now: keycast-client can attach an optional actor to a status change (the keycast#295 contract), with unit tests. reverted the age-review PATCH + UI that declared and forwarded a client pubkey, so nothing writes an unverified actor to the durable audit anymore. #178 wires it end to end once there's an identity to attribute, and that's what closes #175 now.

left the pre-existing clearVerifiedMinor actor path alone — the same question applies to it, but it's already on main and out of scope here.

dcadenas (approved):
✅ clearer actor attribution, ship it

PR #177, opened 2026-07-13
ci(minor-safety): synthetic check for the minor-review endpoints

mbradley:
Closes #174

## what

a scheduled workflow (every 15 minutes) probing the two minor-review endpoints on both relay-manager workers (production and staging), unauthenticated, expecting 401:

- GET /v1/account/moderation-status
- POST /v1/minor-review-cases/<id>/parent-contact

the healthy signal is 401 **plus the worker's NIP-98 rejection marker** in the body ("expected: Nostr", single-sourced in verifyNip98Auth). review caught that status alone is not enough: every unmatched path on these hosts also answers 401 via the generic admin fallback (verified live), so a status-only probe would stay green through a route regression, the exact #108 failure class. a marker-less 401 now fails with a sanitized body snippet naming the fallback. other alarms: 302 means CF Access got applied to /v1 (devices locked out), 200 means the auth layer is gone (worst case), 5xx/transport is an outage. one in-run retry damps transient blips. the marker coupling is fail-safe: if the worker's error copy changes, the probe goes red and the two get updated together.

## why

the minor-review client fails open by design, so when these endpoints broke (#108) nothing alerted and the gate was silently dead in production for months. this makes that failure mode page instead of hide.

## alerting posture (limits documented in the file)

- a failed run goes red and GitHub emails the user who last touched the cron line (me), per personal notification settings. that is the whole story until the SLACK_WEBHOOK_ALERTS repo secret is configured, at which point failures also post to the alerts channel (the step skips with a warning while unconfigured).
- public repo: GitHub auto-disables cron workflows after 60 days of no repo activity; the header documents the re-enable path.
- schedules run from the default branch only, so this goes live at merge, not on the PR.

## scope notes

- the PR now also touches worker/src/index.ts with a one-line comment coupling the NIP-98 error string to the monitor (rewording it reds the check; the comment says to update both together).
- staging is probed too: its /v1 paths bypass CF Access (verified live: /v1 answers the worker's own 401 while /api/* 302s to Access — devices must reach /v1, so this is by design). a staging 302 means someone extended the Access app over /v1 and locked devices out. my first cut excluded staging on a wrong premise; the adversarial review round caught it.
- TODO(#173): api.divine.video joins as a second probe target once the ingress route lands.

## validation

- all four probes (both endpoints, both workers) verified 401 + marker live; an unmatched /v1 route on both workers verified to flag as unhealthy (the false-green case from review).
- the probe script body was executed locally under bash -e across three legs: healthy (401/401), wrong-host 404 (simulating exactly the #108 breakage), and DNS transport failure. correct diagnostics, GITHUB_OUTPUT, and single-line error annotation in all three. shellcheck clean; YAML validated.
- after merge: run once manually via workflow_dispatch to see the green path in CI proper.
- no visual change; CI-only.

### DIFF
--- a/.github/workflows/minor-review-endpoint-health.yml
+++ b/.github/workflows/minor-review-endpoint-health.yml
@@ -17,6 +17,9 @@
 # worker AND the auth layer is alive. Anything else means trouble:
 #   404 -- routing broke (the #108 failure mode)
 #   302 -- CF Access got applied to /v1 paths (devices would be locked out)
+#   200 -- auth layer removed/bypassed (worst case: unauthenticated reads)
+#   403 -- the NIP-98 handler's status contract changed (mobile clients
+#          share this contract; investigate the worker auth path)
 #   5xx/transport -- worker or edge outage
 #
 # Runbook on failure:
@@ -35,10 +38,26 @@
 # https://api.divine.video as a second BASE_URL probe target so coverage
 # for pre-fix app installs is monitored too.
 #
-# Alerting: a failed run notifies via GitHub (scheduled-workflow failure
-# email + the red run). If the SLACK_WEBHOOK_ALERTS secret is configured
-# (shared with the ReportWatcher health check design), it also posts to
-# the alerts channel; absent secret just skips that step.
+# Alerting -- know the limits:
+#   - A failed run turns red, and GitHub emails ONLY the user who last
+#     modified the cron line below (subject to their personal notification
+#     settings). Until Slack is wired, that one person is the whole
+#     alerting story.
+#   - If the SLACK_WEBHOOK_ALERTS secret is configured, failures also post
+#     to the alerts channel; absent secret, that step skips with a warning.
+#   - This repo is public: GitHub auto-disables scheduled workflows after
+#     60 days without repository activity -- the monitor can die silently
+#     during a quiet period. Re-enable from the Actions tab; consider a
+#     keepalive if the repo goes dormant.
+#   - Schedules only run from the default branch: this check goes live at
+#     merge, not on the PR.
+#
+# Setup:
+#   1. (Recommended) Repo secret SLACK_WEBHOOK_ALERTS -- incoming webhook
+#      for the alerts channel (same value as vmalertmanager-slack-webhook-prod
+#      in GCP Secret Manager; shared with the ReportWatcher health design).
+#   2. To test manually: Actions tab > "Minor-Review Endpoint Health" >
+#      Run workflow.
 
 name: Minor-Review Endpoint Health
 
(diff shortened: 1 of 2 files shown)


.github/workflows/minor-review-endpoint-health.yml
```diff
@@ -0,0 +1,164 @@
...
+              if [ "$method" = "GET" ]; then
+                code=$(curl -s -o /dev/null -w '%{http_code}' --max-time 15 \
+                  "$BASE_URL$path" 2>/dev/null) || code="transport_error: $(diagnose $?)"
+              else
+                code=$(curl -s -o /dev/null -w '%{http_code}' --max-time 15 \
+                  -X POST -H 'Content-Type: application/json' \
+                  -d '{"email":"synthetic@check.invalid"}' \
+                  "$BASE_URL$path" 2>/dev/null) || code="transport_error: $(diagnose $?)"
+              fi
+              if [ "$code" = "401" ]; then
```
dcadenas: Require each probe to retain the response body and verify both HTTP 401 and the endpoint-specific NIP-98 missing-auth marker. The current status-only check also accepts the generic admin 401 for an unknown or method-mismatched route, so a broken minor-review endpoint can keep the workflow green. Keep the existing retry, failure aggregation, and alert behavior while adding this route-specific assertion.

dcadenas (changes requested):
The synthetic checks can report healthy when the monitored routes are unreachable because they accept any 401 from the generic admin fallback. Require each probe to verify the endpoint-specific NIP-98 missing-auth response before merging so route regressions trigger an alert.

.github/workflows/minor-review-endpoint-health.yml
mbradley: verified and implemented. the premise checked out harder than stated: on these hosts an unmatched /v1 path can never 404, the worker's generic admin fallback answers 401 for it (confirmed live with /v1/nonexistent-route and a method-mismatch probe on both workers), so the status-only check was blind to exactly the route-regression class this monitor exists for.

each probe now retains the body and requires 401 plus the endpoint-specific NIP-98 rejection marker ("expected: Nostr", single-sourced in verifyNip98Auth). a marker-less 401 fails the run with a sanitized body snippet naming the generic fallback. retry, aggregation, and alerting are unchanged. the coupling is fail-safe and documented on both sides: the workflow header explains it, and the error string in worker/src/index.ts now carries a comment pointing at the monitor so a reword doesn't red the check with no hint at the edit site.

verified live: all four probes healthy with the marker; an unmatched /v1 route on both workers now flags as unhealthy.

dcadenas (approved):
🔍 synthetic endpoint checks are ready to run

PR #179, opened 2026-07-14
fix(age-review): show a resolving state during the report hand-off lookup

mbradley:
Relates to #152 (the report -> age-review hand-off).

## what

the "Handle in Age Review" hand-off resolves `?pubkey=` to the account's active case and opens it. this PR makes every intermediate state of that hand-off honest:

- while the lookup is in flight, the detail pane says "Opening this account's age-review case…" instead of the generic "Select a case to view details" (which moderators read as "nothing happened"; surfaced by the T&S team 2026-07-14; the lookup measures ~0.4s on prod from a fast connection).
- the hand-off's second window is gone entirely: the lookup already returns the full case, so the ?case= normalization seeds the per-case cache instead of paying a second uncached round-trip (during which the pane used to regress to the generic prompt and the mobile sheet bounced closed and open again). caught by adversarial review of the first cut.
- a failed lookup gets its own message ("Couldn't look up this account's case — retry from the report") instead of landing in the no-active-case text, which was a wrong diagnosis for a server error.

## review round: cache coherence across mutations

review caught that the seed created a staleness hole: a terminal action drops the case from the active list, and the detail's fallback served the still-fresh seeded row — actionable controls with the old expected_version. fixed, then an adversarial pass proved the same class survived one key over (the hand-off's lookup cache), where re-entry within its cache lifetime re-seeded the stale active row and clobbered the first fix:

- mutation success writes the PATCH's returned row through to ['age-review-case', id] AND writes the truth to ['age-review-active-case', pubkey] (terminal = no active case); the version-conflict reload invalidates the per-case entry and REMOVES the lookup entry (the seed fires synchronously from cached data, so invalidation alone provably cannot close it).
- two new integration tests drive the real detail component: one-click Clear within the seed staleTime (terminal state renders, no second action), and hand-off re-entry after a terminal action (no resurrected active controls). both verified to fail against the pre-fix code.

## review round 2: the list caches

review round 2 caught the third reader in the same class: selectedCase prefers the list row, and every cached list retained the pre-action row until its refetch landed (or forever on failure), shadowing both previously repaired keys. earlier tests missed it because their mock lists were empty. fixes:

- mutation success reconciles the response row into every cached ['age-review-cases', ...] entry via a pure helper that mirrors the worker's list filters (drop where it no longer belongs, replace in place, leave absent rows for the refetch — absence can't render stale controls). the refetch still supersedes with full server truth.
- the adversarial pass then caught the reconciler's own footgun: writing after invalidating un-invalidates unobserved list caches (setQueryData clears isInvalidated and fresh-stamps), suppressing refetch-on-mount for up to gcTime. the invalidation now runs AFTER the writes, and no-op reconciliations skip the write entirely so untouched entries keep their invalidation state.
- coverage: 8 unit tests on the filter mirror; integration tests with a NONEMPTY active list against a hanging refetch, a failing refetch, and a remount of the same ?pubkey= hand-off; plus the un-invalidation repro (closed-tab within staleTime). every one verified red against the pre-fix code.

## verification

- new AgeReview hand-off render tests (first test harness for this page) pin the sequencing with deferred promises: spinner shows during resolution with no generic-prompt interlude and **zero** second fetches (`getAgeReviewCase` asserted never called); settled-empty shows the explicit no-case message; lookup failure shows the error state; no hand-off shows the generic prompt with no lookup fired.
- `tsc -p tsconfig.app.json` clean, eslint clean, full frontend suite 309 pass, vite build clean.
- visual change: three transient/terminal pane states (spinner, error, unchanged no-case). all use the standard muted empty-state styling matching the siblings; screenshots impractical for sub-second states, and the tests pin the sequencing directly.

### DIFF
--- a/src/components/AgeReview.tsx
+++ b/src/components/AgeReview.tsx
@@ -122,11 +122,12 @@ export function AgeReview() {
     // the ?case= normalization renders immediately instead of paying a second
     // round-trip (during which the pane regressed to the generic prompt and
     // the mobile sheet bounced closed) — the hand-off's second half.
-    if (pubkeyLookup?.case) {
-      queryClient.setQueryData(['age-review-case', pubkeyResolvedId], { success: true, case: pubkeyLookup.case });
+    const resolvedCase = pubkeyMatch ?? pubkeyLookup?.case;
+    if (resolvedCase) {
+      queryClient.setQueryData(['age-review-case', pubkeyResolvedId], { success: true, case: resolvedCase });
     }
     setSelectedCaseId(pubkeyResolvedId);
-  }, [pubkeyParam, pubkeyResolvedId, pubkeyLookup, queryClient, setSelectedCaseId]);
+  }, [pubkeyParam, pubkeyResolvedId, pubkeyMatch, pubkeyLookup, queryClient, setSelectedCaseId]);
 
   // A ?pubkey= that resolves to no active case (already cleared/verified, or
   // still being created) gets an explicit empty state, not a blank panel — but
--- a/src/components/AgeReview.tsx
+++ b/src/components/AgeReview.tsx
@@ -1,6 +1,6 @@
 import { useState, useMemo, useEffect, useCallback } from "react";
 import { useSearchParams } from "react-router-dom";
-import { useQuery } from "@tanstack/react-query";
+import { useQuery, useQueryClient } from "@tanstack/react-query";
 import { useAdminApi } from "@/hooks/useAdminApi";
 import { Card } from "@/components/ui/card";
 import { Badge } from "@/components/ui/badge";
@@ -54,6 +54,7 @@ export function AgeReview() {
   const api = useAdminApi();
   const isMobile = useIsMobile();
   const [searchParams, setSearchParams] = useSearchParams();
+  const queryClient = useQueryClient();
   const selectedCaseId = searchParams.get('case');
   const pubkeyParam = searchParams.get('pubkey');
 
@@ -108,7 +109,7 @@ export function AgeReview() {
   const pubkeyMatch = pubkeyParam
     ? activeData?.cases?.find(c => c.pubkey === pubkeyParam) ?? null
     : null;
-  const { data: pubkeyLookup, isFetching: pubkeyFetching } = useQuery({
+  const { data: pubkeyLookup, isFetching: pubkeyFetching, isError: pubkeyLookupFailed } = useQuery({
     queryKey: ['age-review-active-case', pubkeyParam],
     queryFn: () => api.getActiveAgeReviewCase(pubkeyParam!),
     enabled: !!pubkeyParam && !pubkeyMatch,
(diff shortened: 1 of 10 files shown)


src/components/AgeReview.tsx
```diff
@@ -108,22 +109,37 @@ export function AgeReview() {
...
-    if (pubkeyParam && pubkeyResolvedId) setSelectedCaseId(pubkeyResolvedId);
-  }, [pubkeyParam, pubkeyResolvedId, setSelectedCaseId]);
+    if (!pubkeyParam || !pubkeyResolvedId) return;
+    // The lookup already returned the full case: seed the per-case cache so
+    // the ?case= normalization renders immediately instead of paying a second
+    // round-trip (during which the pane regressed to the generic prompt and
+    // the mobile sheet bounced closed) — the hand-off's second half.
+    const resolvedCase = pubkeyMatch ?? pubkeyLookup?.case;
+    if (resolvedCase) {
+      queryClient.setQueryData(['age-review-case', pubkeyResolvedId], { success: true, case: resolvedCase });
```
dcadenas: Update or invalidate `['age-review-case', c.id]` in the mutation success path after a terminal action. The hand-off seeds this entry here and keeps it fresh for 30 seconds, while the existing success handler only invalidates the list keys at `src/components/AgeReviewDetail.tsx:119-120`; once the active list drops the closed case, the detail can fall back to this stale record and show actionable controls with the old `expected_version`. Add a regression test that opens a `?pubkey=` hand-off, completes a terminal transition within 30 seconds, and verifies the detail shows the terminal state without allowing a second action.

dcadenas (dismissed):
Synchronize the seeded per-case cache after terminal moderation actions. Without that update, the hand-off can leave stale active controls visible and send an old `expected_version`, so this must be fixed before merge.

src/components/AgeReview.tsx
mbradley: oof - we missed this. was actually certain of the opposite. thanks as always.

fixed, and it went two layers further than the diff. mutation success now writes the PATCH's returned row through to the per-case entry AND writes the truth to the hand-off's active-case lookup key (terminal means no active case), because re-entering the hand-off from the report within that key's cache lifetime re-seeded the pre-action row and clobbered the first fix. the writes key off the response row and mutate-time identity rather than the render closure, so a mid-flight case switch can't misroute them. and the seed itself is now stamped with its source's timestamp instead of "now", so a case closed by the cron or another moderator self-corrects in one round-trip instead of riding the cache. the version-conflict reload invalidates the per-case entry and removes the lookup entry outright, since the seed reads cached data synchronously and an invalidation can't win that race.

your regression test is in as specified: a ?pubkey= hand-off, a one-click terminal transition inside the seed's staleTime, asserting the terminal state renders with no second action possible. three more integration tests cover the re-entry, mid-flight-switch, and cross-actor scenarios the follow-up rounds surfaced. all verified to fail against the pre-fix code.

dcadenas (changes requested):
The active-list hand-off can still render stale moderation controls after a successful terminal action. Reconcile the mutation response across active-list caches, add regression coverage for delayed or failed refetches and remounts, and address this before merge.

---

- **The active-list cache still shadows the repaired per-case cache** ([src/components/AgeReview.tsx:104](https://github.com/divinevideo/divine-relay-manager/blob/1a70b3ace99ae0495ffc17ef477f7e8a96c486f6/src/components/AgeReview.tsx#L104))
  `selectedCase` still prefers `inFilteredList` over the response-updated per-case entry at `src/components/AgeReview.tsx:104`. Reconcile the successful terminal mutation result into each matching `['age-review-cases', ...]` cache, removing terminal rows or making the response-backed row authoritative before rendering. The current update only writes `['age-review-case', id]` and `['age-review-active-case', pubkey]` in `src/components/AgeReviewDetail.tsx:125-142`, so a retained list row can keep Clear and Deny visible while refetch is pending or fails. Add a regression test with a nonempty active list, delayed or failed refetch, and a remount of the same `?pubkey=` hand-off to verify stale controls never return.

src/components/AgeReview.tsx
mbradley: round two landed. the list caches were indeed the remaining reader: selectedCase prefers the list row, and every cached filter combo kept the pre-action row while its refetch was pending, or forever if it failed. mutation success now reconciles the response row into each cached list through a pure helper that mirrors the worker's list filters: dropped where it no longer belongs, replaced in place where it does, absent rows left for the refetch since absence can't render stale controls. the adversarial pass then caught the reconciler's own footgun, writing after invalidating un-invalidates unobserved list caches and suppresses refetch on mount, so the invalidation now runs after the writes and no-op reconciliations skip the write entirely.

your three scenarios are covered as specified: a nonempty active list against a hanging refetch, against a failing refetch, and a remount of the same ?pubkey= hand-off, plus the un-invalidation repro (closed tab within staleTime). every test verified red against the pre-fix code. the branch now carries nine regression tests on this state machine across the review rounds.

honest note: four holes in one cache topology is a signal. if you see anything else in this class, i'd rather do the structural fix (single source of truth for a case, likely the select-from-list pattern) as a follow-up than patch a fifth reader. happy to file that either way.

dcadenas (approved):
✅ clearer report hand-off state, ship it

PR #181, opened 2026-07-14
feat(auth): moderator identity via divine-login (phase 1)

mbradley:
## what this does

moderators can now sign in with their divine account inside relay-admin, and their nostr pubkey gets attributed to the moderation actions they take. before this, nothing rendered a login at all, so every `moderation_decisions` row was either unattributed, `auto` (ReportWatcher), or the shared worker admin key. zero rows were ever attributed to an individual human.

this is phase 1 of #178: the login surface + attribution. it's frontend only, no worker changes.

## why

verified against prod d1 (2026-07-14): across all `moderation_decisions` rows, 2217 are `null`, 833 are `auto`, 45 are the shared worker key, and zero are a real human. the biggest null buckets come from the primary surfaces: report dismissals (1199) and event deletes (447). UserManagement's `moderator_pubkey` already reads `useCurrentUser`, so it stayed null until a login shipped, and the other surfaces never even sent one.

## how it works

- **access stays CF Access, identity is divine-login.** CF Access still decides who reaches the tool (workspace email). the login's job here is purely attribution. a keycast outage degrades attribution to log-only, it never locks anyone out.
- **built on the `@divinevideo/login` SDK** (OAuth2 + PKCE against login.divine.video). the SDK owns the flow, token storage, refresh, and the REST signing RPC. the only glue i own is a thin `NostrSigner` adapter over the SDK's `DivineRpc` (with a `TODO(#178)` to swap for the published `divine-signer` when it lands), a session provider that resolves the moderator's pubkey once, an `/auth/callback` page, and a sign-in button in the shell header.
- **attribution wired across every moderation surface.** `useCurrentUser` now yields the moderator's pubkey, and i wired `moderator_pubkey`/`moderatorPubkey` into UserManagement (already there), ReportDetail, EventDetail, EventActions, UserActions, BulkDeleteByKind, and DebugPanel. every code path that writes a `moderation_decisions` row now attributes the acting moderator.
- **removed the dead nostrify login stack.** `useCurrentUser` no longer reads the nostrify login store, so LoginArea/LoginDialog/SignupDialog/AccountSwitcher + their hooks + NostrLoginProvider were fully bypassed dead code. removed them.

## what's not in this PR (follow-ups)

- **age-review attribution.** #176 merged, but only its worker keycast actor contract, not the `AgeReviewDetail` frontend change that sends `moderator_pubkey`, nor the worker actor-forwarding on the suspend/ban legs. so age-review attribution needs both re-landed (frontend sender + worker forwarding). that's a separate follow-up, not wired here.
- **keycast client registration (#282 hardening).** we ship against a new unregistered `divine-relay-admin` client_id, which login.divine.video accepts today (open OAuth, verified against the live server). registering the client to lock the redirect_uri to our domain is a follow-up, not a blocker.
- phase 2 (worker verifies the token server-side) and phase 3 (per-moderator NIP-98).

## known limitation

the signer has no `onUnauthorized` refresh, so if the access token expires during a long focused session, a profile-edit signature hard-fails rather than auto-refreshing. session-level refresh on focus/mount covers the common case, and attribution needs only the pubkey (not signing). documented in `divineSigner.ts`; worth revisiting in phase 2.

## adversarial review

ran fresh-eyes reviewers (correctness, end-to-end inertness, test quality). they caught real holes that this PR fixes: OAuth calls had no timeout, the signer's fetch wrapper broke the SDK's body timeout, a logout could be undone by an in-flight refresh, a token-absent login became a silent dead state, and (the big one) login only lit up UserManagement attribution while the primary surfaces sent nothing. a composed integration test now drives the real provider → useCurrentUser → a real moderation action and asserts the pubkey lands on the audit write.

## manual validation

local + staging:
- [ ] fresh profile → header shows "Sign in"
- [ ] click → redirect to login.divine.video with the right client_id / callback / PKCE (verified: server accepts our client)
- [ ] after auth → back at the return path, header shows the moderator
- [ ] ban a user / dismiss a report → confirm `moderator_pubkey` persists on the `moderation_decisions` row
- [ ] confirm the `/auth/callback` route works behind CF Access on relay.admin.divine.video (staging)
- [ ] sign out → header returns to "Sign in", subsequent actions write null attribution

checks run: `npx tsc -p tsconfig.app.json --noEmit` clean, `npx eslint` clean, `npx vitest run` all green (326 passed), `npx vite build` succeeds. CI green on the PR.

## screenshots

signed-out shell header (Sign in next to the environment selector):
`docs/screenshots/2026-07-14-signed-out-header.png`

login.divine.video accepting our client:
`docs/screenshots/2026-07-14-login-accepts-client.png`

phase 1 of #178 (leave the issue open for phases 2 and 3)


src/components/EventActions.tsx
```diff
@@ -45,7 +47,7 @@ export function EventActions({
   // log so the report converges without a manual refresh; on failure, surface a
   // non-blocking toast. Mirrors UserActions.logAudit.
   const logAudit = (params: Parameters<typeof api.logDecision>[0]) =>
-    void api.logDecision(params)
+    void api.logDecision({ moderatorPubkey: user?.pubkey, ...params })
```
dcadenas: Capture the moderator pubkey when the action starts and pass that captured value into this audit write. `user?.pubkey` is read only after `await api.banEvent(...)`, so a signed-in load whose `getPublicKey()` is still pending records `null` even though the action succeeds. The bulk completion callback in `src/components/UserActions.tsx:203` has the same problem when logout or account switching changes the current user before the job finishes. Keep moderation non-blocking: let only the detached audit path wait briefly for the in-flight identity, then fall back to `null`.

dcadenas (dismissed):
Capture the moderator identity when each moderation action starts and carry it through the audit write. The current render-time `user?.pubkey` can record null during normal async startup and can attribute a long-running bulk job to a later account, so these races must be fixed before merge.

Out-of-PR follow-ups
- **Publish the refresh-concurrency fix** in `divinevideo/divine-login`: publish the existing refresh-concurrency fix before relying on `@divinevideo/login@1.1.0`.

src/components/EventActions.tsx
```diff
@@ -44,13 +46,16 @@ export function EventActions({
...
+  // Detached audit write: waits briefly for the moderator identity (captured at
+  // call time), then attributes or falls back to null. Never blocks the action.
   const logAudit = (params: Parameters<typeof api.logDecision>[0]) =>
-    void api.logDecision(params)
-      .then(() => { queryClient.invalidateQueries({ queryKey: ['decisions'] }); })
-      .catch((e) => {
-        console.warn('[EventActions] audit log failed', e);
-        toast({ title: 'Action applied; audit log not recorded' });
-      });
+    void getModeratorPubkey().then((moderatorPubkey) =>
```
dcadenas: Capture the moderator promise before `banEventMutation` awaits the authoritative request, then pass that same promise to `logAudit`. At this line, `getModeratorPubkey()` reads the active signer only after `banEvent` completes at lines 62-63, so a logout or account switch during the request can record no moderator or the later account. Apply the same action-start capture to `UserActions`, `ReportDetail`, and `DebugPanel`, reusing one capture for combined or partial writes, without making moderation wait for identity.

dcadenas (changes requested):
Capture the moderator identity before the first authoritative moderation request and reuse it for every audit write.

dcadenas (approved):
🔐 Divine Login now tracks moderator identity cleanly

PR #183, opened 2026-07-15
fix(reports): resolve deep-link targets to found/gone/unavailable instead of a blank pane

mbradley:
## Problem

Clicking "View in Relay Admin" from a Zendesk ticket could land on the Reports list with a blank right pane and no explanation, whenever the report's target wasn't in the fetched set -- most starkly when a reported account self-deleted (NIP-62 vanish) and its report event came off the relay before anyone acted.

Closes #193.

## Change

The deep-link now resolves into four explicit states instead of "found, or blank forever":

- **found** -- target present (bulk list or targeted lookup) -> opens the report.
- **resolving** -- a brief "looking up this report" state.
- **gone** -- a targeted relay query confirms the report is no longer on the relay -> clear message + the target id + any prior moderation actions on it.
- **unavailable** -- the relay lookup failed or couldn't confirm -> "couldn't reach the relay, retry", never mislabeled as deleted.

To make "gone" honest (and to resolve reports aged out past the 200-record bulk window), `/api/reports` accepts an optional `?event=`/`?pubkey=` filter and the client does a targeted lookup on a bulk miss. **Existence is gated on the relay's own `#e`/`#p` filter result** -- authoritative, since every returned report tags the queried target -- not on a re-filter by resolved target. An earlier cut re-filtered by resolved target, which false-flagged a report that only p-tags a pubkey (but resolves to an event) as "gone"; resolved-target matching is now only a display preference for *which* report to show.

Defensive about failure modes:
- A slow-but-healthy relay (no EOSE within timeout) surfaces as **unavailable** with retry, never a false "deleted" -- the worker distinguishes an EOSE-confirmed empty result from a timed-out one.
- A lookup resolving after the moderator navigated away is dropped rather than firing a late navigate.
- A "found" target lands on `/reports/:id`, so the deep-link query params are cleared and the in-session selection is stable. A full reload of an aged-out `/reports/:id` is still limited by the current report list window.
- Hex target params are lowercased so an uppercase-hex deep link still matches (client and worker in parity).
- Targeted `?event=`/`?pubkey=` report queries keep the same 200-result limit as the bulk list.
- If ban/label/decision data marks the selected deep-link target resolved after navigation, the UI re-opens the resolved filter so the selected row does not disappear.

## Out of scope

- The 200-record cap on the Reports **list** view (verified currently binding; a list-view fix is separate).
- Full reload/shared-link recovery for aged-out `/reports/:id` route IDs. Zendesk links use `?event=`/`?pubkey=`, which this PR resolves.
- `getReportTarget`'s e-tag-first resolution -- a separate latent item (`TODO(#160)`).

## Testing

- Unit tests: the worker filter + unconfirmed-miss guard, the client call, and the resolution helpers -- including regressions for the two false-"gone" cases (p-tag-only `?pubkey=`, multi-e-tag `?event=`), mutation-verified to fail on the old behavior.
- Integration tests (RTL): gone / found (asserting the `/reports/:id` URL) / unavailable / resolve-after-unmount / bulk-hit URL cleanup / late ban-data unhide.
- `npm run test`
- `cd worker && npx vitest run`

### DIFF
--- a/src/components/Reports.tsx
+++ b/src/components/Reports.tsx
@@ -859,6 +859,10 @@ export function Reports({ relayUrl, selectedReportId }: ReportsProps) {
     detailBoundaryRef.current?.reset();
     setSelectedReport(report);
     setDeepLinkStatus('idle'); // clear any deep-link fallback once the user interacts
+    // Invalidate any in-flight targeted lookup: a user selection/dismissal
+    // supersedes the deep link, so a late response must not pass the resolution
+    // guard and re-navigate (e.g. closing the mobile resolving Sheet mid-lookup).
+    attemptedTargetRef.current = null;
     if (report) {
       navigate(`/reports/${report.id}`, { replace: true });
     } else {
--- a/src/components/Reports.tsx
+++ b/src/components/Reports.tsx
@@ -45,7 +45,7 @@ import { ReportDetail } from "@/components/ReportDetail";
 import { ErrorBoundary } from "@/components/ErrorBoundary";
 import { ReportDetailErrorFallback } from "@/components/ReportDetailErrorFallback";
 import { DeepLinkFallback } from "@/components/DeepLinkFallback";
-import { classifyTargetedFetch, decisionsForTarget, type DeepLinkStatus } from "@/lib/deepLinkResolution";
+import { classifyTargetedFetch, decisionsForTarget, reportsMatchingTarget, type DeepLinkStatus } from "@/lib/deepLinkResolution";
 import { useAdminApi } from "@/hooks/useAdminApi";
 import { AUTO_HIDE_ACTIONS, CATEGORY_LABELS, HIGH_PRIORITY_CATEGORIES, getReportCategory } from "@/lib/constants";
 import { useIsMobile } from "@/hooks/useIsMobile";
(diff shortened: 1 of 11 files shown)


src/components/Reports.tsx
```diff
@@ -732,39 +747,91 @@ export function Reports({ relayUrl, selectedReportId }: ReportsProps) {
...
+      attemptedTargetRef.current = null;
+      return;
+    }
 
-      // Select the report
-      setSelectedReport(targetReport.latestReport);
-      navigate(`/reports/${targetReport.latestReport.id}`, { replace: true });
+    // Bulk list is loaded and the target isn't in it. Do one targeted relay fetch
+    // per target to tell "aged out / vanished" apart from "still loading".
+    if (attemptedTargetRef.current === target.value) return;
```
dcadenas: Scope the de-duplication and stale-result guards to the complete target identity here. The flow distinguishes target type and value at `src/components/Reports.tsx:759-768`, but these guards compare only the raw value at `src/components/Reports.tsx:788-801` (and the stale-result check at `src/components/Reports.tsx:830`), so same-value targets across types or environments can reuse an old attempt. Clear or override `selectedReport` when a new target resolves to gone or unavailable, and open the mobile Sheet for resolving and fallback states as well. Otherwise moderators can see stale detail or no result at all; add tests for each of these transitions.

dcadenas (dismissed):
Keep deep-link resolution scoped to the complete target identity and make its status authoritative over stale selection. The current value-only guards can reuse old attempts or results, while gone and unavailable deep links are hidden on mobile because the Sheet opens only for selected reports. Address these paths before merge and add coverage for target type or environment changes, stale selection, and mobile fallback visibility.

NotThatKindOfDrLiz (approved):
Approved after takeover re-review. I pushed 1127099 with the remaining fixes: selected resolved deep-links unhide once resolution data catches up, targeted report filters are capped, the helper contract/comment no longer contradicts the relay-result existence gate, and the bulk-hit plus late-ban-data paths have regression coverage. Local required checks and GitHub checks are green.

src/components/Reports.tsx
```diff
@@ -1077,7 +1190,10 @@ export function Reports({ relayUrl, selectedReportId }: ReportsProps) {
 
       {/* Mobile Sheet - Report Detail */}
       {isMobile && (
-        <Sheet open={!!selectedReport} onOpenChange={(open) => !open && handleSelectReport(null)}>
+        <Sheet
+          open={!!selectedReport || showDeepLinkResolving || showDeepLinkFallback}
```
dcadenas, line 1198: Invalidate the pending targeted lookup when the mobile resolving Sheet closes. At `src/components/Reports.tsx:1194`, the Sheet uses `handleSelectReport(null)` but leaves `attemptedTargetRef` active, so a later response can pass the guard at `src/components/Reports.tsx:819` and navigate back at `src/components/Reports.tsx:840-843`. Clear the attempt in the close path or abort the request, and add a mobile test that closes the Sheet while the fetch is pending, then confirms `/reports` has no selected detail after the request succeeds.

dcadenas (changes requested):
The mobile resolving Sheet can be dismissed while its targeted lookup remains active. Cancel or invalidate that lookup before merge so a late response cannot reopen the report after the moderator returns to the queue.

src/components/Reports.tsx
mbradley: Fixed. `handleSelectReport` — the shared select/dismiss path, including the mobile Sheet's `onOpenChange` — now clears `attemptedTargetRef.current = null` alongside setting the status idle. So closing the resolving Sheet mid-lookup makes a late targeted response fail the resolution guard (`null !== targetKey`) at both the post-fetch and catch sites, dropping it — it can't re-navigate and reopen the report. Added the mobile close-while-pending test you asked for; mutation-verified it reopens without the fix. `1ad28a0`.

dcadenas (approved):
✅ deep-link targets now resolve clearly

PR #186, opened 2026-07-16
fix(reports): treat relay query timeout/early-close as errors, not empty results

### LINKED ISSUE #185: Relay query timeouts render as an empty moderation queue
## Summary

The reports queue (and the resolution-labels fetch) can silently render empty when the relay is slow, replacing a populated list on the next 15-second poll. Surfaced while testing on staging, and diagnosed in the wake of Daniel's analysis, seen this morning, of staging ClickHouse memory pressure (divinevideo/divine-iac-coreconfig#1230): staging reads intermittently stall or get killed, which made the failure mode visible. The defect is ours and applies to production too: any relay response slower than 5 seconds makes the prod queue render as "no reports pending." For a moderation tool that is failing quiet in the wrong direction.

## Root cause

`queryRelay()` (worker/src/index.ts) resolves `{ success: true, events: <partial> }` on both its 5-second timeout and on a WebSocket close before EOSE. The API layer and the frontend cannot distinguish "relay answered: queue is empty" from "relay didn't answer in time." `Reports.tsx` compounds it: the list refetches every 15 seconds and each response replaces the data wholesale, so one slow poll blanks a healthy queue mid-session (observed live on staging: list populated, then self-blanked on the next poll).

## Measured (staging, 2026-07-16)

- `/api/reports` returned an empty success on 3 to 4 of 5 attempts.
- Direct relay queries for the same filter returned 200 events on one attempt, then 0 events after 7.2 seconds (past the worker's 5-second timeout) two minutes later.

## Fix

- Worker: `queryRelay` resolves `success: false` (with a distinguishing error) on timeout and on close-before-EOSE; only EOSE-complete results count as success. `/api/reports` and `/api/resolution-labels` already return 502 on failure; the reopen label cleanup already guards on `success` and now skips cleanly (with a log) instead of acting on partial reads.
- Frontend: on a refresh error with previous data present, keep the stale list rendered with a warning notice and last-updated time, instead of the destructive full-pane error. The full-pane error remains for first-load failures.

## Out of scope / follow-up

When the relay's backing store kills a query, funnelcake sends a normal empty EOSE, indistinguishable from a genuinely empty queue at the protocol level. Fixing that requires a funnelcake-side change (error the subscription instead of returning an empty success). Candidate follow-up issue there.

mbradley:
## Summary

One slow relay response could blank the moderation queue. `queryRelay()` resolved `success: true` with whatever events had arrived when its 5 second timeout fired or the socket closed before EOSE, so the API returned an empty 200 that the UI rendered as "no reports pending." With the list polling every 15 seconds and each poll replacing data wholesale, a single slow response wiped a populated queue mid-session.

Now only an EOSE-complete result reports success, and the relay's own failure signal is honored: NIP-01 `CLOSED` fails immediately with the reason the relay gave, instead of waiting out the timeout and reporting it as something else.

## Motivation

For a moderation tool an empty queue is a load-bearing claim: it tells a moderator there is no work left. Returning a partial read as a successful one lets a slow relay make that claim falsely, and the failure is silent and self-correcting on the next poll, which is what let it survive this long.

Measured against the production relay over 45 trials, the resolution-labels query runs at p50 2831ms and p90 3177ms, with a maximum of 4595ms against the hard 5 second cap. One trial timed out and returned zero events as a success. This is not a staging-only condition; production sits close enough to the cap that ordinary variance crosses it.

Closes #185.

## An incomplete read is not an empty one

`success` now answers exactly one question: did the relay confirm the end of the set? That is the only question an absence-sensitive caller may act on, and `/api/reports` and `/api/resolution-labels` still 502 on anything less.

It does not mean the partial data is worthless, and an earlier revision of this branch conflated the two. Every failure path now carries the events that did arrive, because the one caller that removes things rather than counting them is strictly better off acting on a label it received than discarding it.

That distinction is load-bearing for reopen. `handleDeleteDecisions` deletes the D1 decision rows unconditionally while relay-side label cleanup is best-effort, so discarding the labels made a slow relay *worse* than before this branch: every reopen destroyed the decision rows, cleared no labels, and left the report hidden. On `main` the same read returned `complete: false` and the labels it saw were removed.

## Reopen stops asserting a queue state it cannot verify

`resolvedTargets` hides a report using four sources: resolution labels, banned pubkeys, banned events, and auto-hide decisions. Reopen clears two of them. Delete-event and ban-pubkey are the most common resolution paths, so a report resolved either way stays hidden after a completely successful reopen, while the toast said it was back in the queue. The tooltip made the same promise.

Both now report what the action did rather than where the report ends up, and the tooltip names the case it cannot fix.

Three narrower reopen failures close with it:

- A label read that fills its page is indistinguishable from a truncated one, so it now reports an incomplete cleanup rather than a clean reopen. The cap was 10, against an event kind that accumulates per review cycle; it is now 50, and the check is honest that it detects our own limit rather than one the relay might impose. 50 is not set by the expected label count (a real target carries one or two) but by the client's 30s abort: every label costs one sequential signed NIP-86 round-trip, so a full page is ~15-25s of work per filter.
- `publishLabel` tags a label `e` **or** `p`, never both, so of the two filters reopen ran against every target, one could never match. Harmless when a stall was an empty success; with a stall now a failure it produced a destructive "try reopening again" toast for a query that never had anything to find. The client names the type it already knows; omitting it stays valid and checks both, because worker and Pages deploy separately.
- The two deletes are sequential and the first commits server-side before the second runs, so a failure on the second left the panel rendering decisions the server no longer had.

## Resolution-state completeness belongs to #221

An earlier revision added a banner for an unavailable resolution filter. It only fired when the labels query had *never* succeeded — `placeholderData` keeps the last good set, so one success made it permanently silent, which is the shape the outage actually takes: labels load when the tool opens, then start failing mid-session.

Rather than patch the condition, the banner is gone. `fix/221-resolution-state-completeness` reports per-source truncation and coverage from the worker and gates the queue on every source, covering this case and the three sibling sources (`banned-pubkeys`, `banned-events`, `decisions`) that fail just as silently. A second competing signal in the same component would have to be reconciled or ripped out. `retry: 1` on the labels query stays; its justification never depended on the banner.

## Relay behavior, verified rather than assumed

An earlier draft recorded a residual: that a backend killing a query while still sending an empty EOSE is indistinguishable from an empty queue. Checking the relay implementation, that is not what happens — on a store failure funnelcake sends `CLOSED` with `error: could not complete query` and never sends EOSE, which is exactly the signal NIP-01 defines. The gap was on our side: `queryRelay` parsed only `EVENT` and `EOSE`, so a failure the relay had already reported sat until the timeout and was then reported as a timeout.

The residual that genuinely remains is narrower: a relay returning a truly empty EOSE after an internal failure is still indistinguishable at the protocol level. That is not the current relay behavior. (Thanks @realmeylisdev for confirming the second `stored replay timed out` path also skips EOSE and is caught by the same branch.)

## Behavior changes worth calling out

**Deep links.** A targeted `?event=`/`?pubkey=` lookup that received events and *then* timed out previously rendered as "found". It now renders "Relay unavailable". That is deliberate — a truncated targeted read is not proof of what it found — but it is a real change, and an earlier version of this description wrongly called deep-link behavior unchanged.

**Zendesk enrichment.** `handleParseReport`'s kind-0 and event lookups now get `success: false` where a close-before-EOSE previously yielded partial data. Best-effort path; the note still posts, without the display name or restored-OG-Vine flag.

## Removed

`isUnconfirmedTargetedMiss` and the `complete` field it read. Success now implies an EOSE-confirmed result, so its `!complete` branch was unreachable while its unit tests kept passing, which read as coverage for a live guard.

## API changes

`DELETE /api/decisions/:id` gains an optional `?targetType=event|pubkey` and a `labelCleanupFailed` field. Any value other than `event` or `pubkey` — including empty, wrong case, or junk — falls back to querying both tags, which is the same behavior an older frontend gets.

```jsonc
// DELETE /api/decisions/<id>?targetType=event  -> 200
{ "success": true, "deleted": 3, "labelsDeleted": 1, "labelCleanupFailed": false }

// same, when a label was found but could not be removed -> 200
{ "success": true, "deleted": 3, "labelsDeleted": 0, "labelCleanupFailed": true }

// GET /api/reports, relay timed out before EOSE -> 502 (was 200 with partial events)
{ "success": false, "error": "Relay query timed out before EOSE (2 events received)" }

// GET /api/reports, relay refused the subscription -> 502
{ "success": false, "error": "Relay closed the subscription: error: could not complete query" }
```

## UI changes

One new alert strip, three copy changes, and one new interactive affordance (the degraded toast's `Try again`). No layout or component changes.

The degraded toast deliberately does not name the relay: one of the three causes is our own page cap, where the relay did exactly what it was asked, so blaming it would send a moderator chasing a problem that is not there. `ReportDetail.reopen.test.tsx` pins that wording.

| Surface | Before | After |
|---|---|---|
| Reports header (new) | queue silently blanked | `Live refresh is failing. Showing reports loaded 2m ago; retrying automatically.` |
| Reopen toast, clean | `Report reopened` / `This report is now back in the pending queue` | `Report reopened` / `Decisions and resolution labels cleared. A report hidden by a relay ban or deletion stays hidden until that is lifted.` (10s, since the caveat does not fit Radix's 5s default) |
| Reopen toast, degraded (new) | — | `Reopened, but resolution labels could not all be cleared` / `Some may remain, so this report may stay hidden.` (destructive, with a `Try again` action and `duration: Infinity`) |
| Reopen tooltip | `Reopen this report for review. Removes the dismiss decision and puts it back in the pending queue.` | `Reopen this report for review. Clears its moderation decisions and resolution labels. A report hidden by a relay ban or deletion stays hidden until that is lifted.` |

## Validation

Every guard was mutation-tested individually — one guard per mutation, so a coarse mutation cannot certify an inert test.

The first pass was not sufficient, and it is worth naming why. Adversarial re-review found six guards that could be reverted with a fully green suite, all of them behavior the commit messages had already claimed. The worst: only one of the four `events` passthrough paths was pinned — including, unpinned, the timeout path this branch is named after — and the truncation guard, the entire argument for raising the label cap, had no test at all. A commit message is not coverage. Each now dies to exactly one mutation.

A second round found one more, and it was the most dangerous thing in the branch. The `targetType` whitelist had no test, and replacing it with the cast it invites survived all 386 tests while producing a filter keyed `"undefined"` — an untargeted query. The relay ignores the unknown key and returns resolution labels for arbitrary targets, which the cleanup loop then bans: one reopen would wipe resolution labels queue-wide and still report success. Pinned now.

Two review findings worth carrying beyond this PR:

- `useAdminApi` is a silent-drop trap for any new argument. Its wrapper signature keeps declaring the parameter it forgets to forward, so TypeScript sees nothing, and every component test mocks the hook wholesale. It now has its own test; the same exposure applies to the other bound functions in that file.
- A guard sitting upstream of a computed object key fails open into a *broader* query rather than a narrower one, which is the direction that does damage.

Two further rounds found two more, both in fixes made by earlier rounds. Resolving before closing the socket was applied to one exit and not the four beside it — and on two of those a throwing `close()` is swallowed by the parse-error catch, so `queryRelay` hangs and `/api/reports` never answers, stalling the poll to the client's 30s abort. That is this PR's own failure arriving by a different door, and it is demonstrated rather than argued: those tests hang to the runner's timeout. Closing now goes through a helper that swallows and logs, which is what actually guarantees settlement. The same round caught the longer toast duration sitting on the wrong toast — the destructive one carrying the retry kept the 5s default, so the affordance expired while the Reopen button was already unmounted.

Gates: frontend 522 passed + 1 skipped, worker 395 passed, worker D1 21 passed, frontend build, both typechecks and eslint clean.

One follow-up found and deliberately left out: `publishToRelay` carries the identical latent hang on both its exits, and it is awaited on the label-publish path that backs `markAsReviewed`. Filed as #225 rather than widened into this PR.

**Manual validation plan (staging).** The queue holds its last data across a failing poll with the stale-data warning rather than blanking. A reopen against a slow relay removes the labels it did receive and reports cleanup as incomplete. A cold load against a healthy relay renders normally. Screenshots of the three states to be captured during this pass.

## Merge order

This should land before #222. That branch relocates `queryRelay` into `worker/src/relay-profile.ts` carrying the pre-fix semantics, and `relay-profile.ts:134` reads `res.complete === true` — a field this PR removes. Rebasing onto this simplifies it: `success` already means EOSE-confirmed, so the check becomes `res.success`. Left as-is it silently evaluates false forever, with no type error, and identity capture never stamps.

#221 has recorded its own rebase plan against this branch; this PR moved toward it rather than against it.

### DIFF
--- a/src/components/ReportDetail.tsx
+++ b/src/components/ReportDetail.tsx
@@ -263,13 +263,15 @@ export function ReportDetail({ report, allReportsForTarget, allReports = [], onD
       // ban-resolved report stays hidden even on a fully clean run.
       toast(labelCleanupFailed
         ? {
-            title: "Reopened, but resolution labels could not be cleared",
-            description: "The relay did not clear them, so this report may stay hidden. Try reopening again.",
+            // Deliberately does not name the relay: one of the causes is our
+            // own page cap, where the relay did exactly what it was asked.
+            title: "Reopened, but resolution labels could not all be cleared",
+            description: "Some may remain, so this report may stay hidden. Try reopening again.",
             variant: "destructive" as const,
           }
         : {
             title: "Report reopened",
-            description: "Moderation decisions and resolution labels were cleared",
+            description: "Moderation decisions and resolution labels were cleared. A report hidden by a relay ban or deletion stays hidden until that is lifted.",
           });
     },
     onError: (error: Error) => {
--- a/src/components/ReportDetail.tsx
+++ b/src/components/ReportDetail.tsx
@@ -262,7 +262,7 @@ export function ReportDetail({ report, allReportsForTarget, allReports = [], onD
       toast(labelCleanupFailed
         ? {
             title: "Reopened, but resolution labels could not be cleared",
-            description: "The relay did not answer, so this report may stay hidden. Try reopening again.",
+            description: "The relay did not clear them, so this report may stay hidden. Try reopening again.",
             variant: "destructive" as const,
           }
         : {
--- a/src/components/ReportDetail.tsx
+++ b/src/components/ReportDetail.tsx
@@ -266,12 +266,22 @@ export function ReportDetail({ report, allReportsForTarget, allReports = [], onD
             // Deliberately does not name the relay: one of the causes is our
             // own page cap, where the relay did exactly what it was asked.
             title: "Reopened, but resolution labels could not all be cleared",
-            description: "Some may remain, so this report may stay hidden. Try reopening again.",
+            description: "Some may remain, so this report may stay hidden.",
             variant: "destructive" as const,
+            // The decisions ARE gone on this path, so hasDecisions goes false
+            // and the Reopen button unmounts. Retrying is the right advice and
+            // the cleanup is idempotent, so the retry has to come with it.
+            action: (
+              <ToastAction altText="Retry the reopen" onClick={() => reopenMutation.mutate()}>
+                Try again
+              </ToastAction>
+            ),
           }
         : {
             title: "Report reopened",
-            description: "Moderation decisions and resolution labels were cleared. A report hidden by a relay ban or deletion stays hidden until that is lifted.",
+            description: "Decisions and resolution labels cleared. A report hidden by a relay ban or deletion stays hidden until that is lifted.",
+            // Longer than the 5s Radix default can be read in.
+            duration: 10000,
           });
     },
     onError: (error: Error) => {
--- a/src/components/ReportDetail.tsx
+++ b/src/components/ReportDetail.tsx
@@ -242,20 +242,33 @@ export function ReportDetail({ report, allReportsForTarget, allReports = [], onD
     mutationFn: async () => {
       if (!context.target) throw new Error('No target');
       // Delete all decisions for this target
-      await deleteDecisions(context.target.value);
+      const primary = await deleteDecisions(context.target.value);
       // Also delete decisions for the pubkey if this is an event report
+      let secondary = { labelCleanupFailed: false };
       if (context.target.type === 'event' && context.reportedUser.pubkey) {
-        await deleteDecisions(context.reportedUser.pubkey);
+        secondary = await deleteDecisions(context.reportedUser.pubkey);
       }
+      return { labelCleanupFailed: primary.labelCleanupFailed || secondary.labelCleanupFailed };
     },
-    onSuccess: () => {
+    onSuccess: ({ labelCleanupFailed }) => {
       queryClient.invalidateQueries({ queryKey: ['decisions'] });
+      // Resolution labels also gate whether the report reappears, so a reopen
+      // that does not refresh them can leave the target hidden for a poll cycle.
+      queryClient.invalidateQueries({ queryKey: ['resolution-labels'] });
       decisionLog.refetch();
       pubkeyDecisionLog.refetch();
-      toast({
-        title: "Report reopened",
-        description: "This report is now back in the pending queue",
-      });
+      // The decisions are gone either way, but a surviving resolution label
+      // keeps the report out of the queue. Do not claim it is back.
+      toast(labelCleanupFailed
+        ? {
+            title: "Reopened, but resolution labels could not be cleared",
+            description: "The relay did not answer, so this report may stay hidden. Try reopening again.",
+            variant: "destructive" as const,
+          }
+        : {
+            title: "Report reopened",
+            description: "This report is now back in the pending queue",
+          });
     },
     onError: (error: Error) => {
       toast({
--- a/src/components/ReportDetail.tsx
+++ b/src/components/ReportDetail.tsx
@@ -257,8 +257,10 @@ export function ReportDetail({ report, allReportsForTarget, allReports = [], onD
       queryClient.invalidateQueries({ queryKey: ['resolution-labels'] });
       decisionLog.refetch();
       pubkeyDecisionLog.refetch();
-      // The decisions are gone either way, but a surviving resolution label
-      // keeps the report out of the queue. Do not claim it is back.
+      // Report what the reopen did, not where the report ends up. Whether it
+      // returns to the queue depends on resolvedTargets, which is also built
+      // from relay bans and deletions that reopen never undoes -- so a
+      // ban-resolved report stays hidden even on a fully clean run.
       toast(labelCleanupFailed
         ? {
             title: "Reopened, but resolution labels could not be cleared",
@@ -267,10 +269,18 @@ export function ReportDetail({ report, allReportsForTarget, allReports = [], onD
           }
         : {
             title: "Report reopened",
-            description: "This report is now back in the pending queue",
+            description: "Moderation decisions and resolution labels were cleared",
           });
     },
     onError: (error: Error) => {
+      // The first delete may already have committed server-side before the
+      // second threw, so the cached decision state can be stale even though
+      // the action reports failure. Refresh it rather than leaving the panel
+      // showing decisions the server no longer has.
+      queryClient.invalidateQueries({ queryKey: ['decisions'] });
+      queryClient.invalidateQueries({ queryKey: ['resolution-labels'] });
+      decisionLog.refetch();
+      pubkeyDecisionLog.refetch();
       toast({
         title: "Failed to reopen",
         description: error.message,
@@ -1088,7 +1098,7 @@ export function ReportDetail({ report, allReportsForTarget, allReports = [], onD
                     </Button>
                   </TooltipTrigger>
                   <TooltipContent side="bottom" className="max-w-xs">
-                    <p>Reopen this report for review. Removes the dismiss decision and puts it back in the pending queue.</p>
+                    <p>Reopen this report for review. Clears its moderation decisions and resolution labels. A report hidden by a relay ban or deletion stays hidden until that is lifted.</p>
                   </TooltipContent>
                 </Tooltip>
               )}
--- a/src/components/ReportDetail.tsx
+++ b/src/components/ReportDetail.tsx
@@ -270,9 +270,10 @@ export function ReportDetail({ report, allReportsForTarget, allReports = [], onD
             variant: "destructive" as const,
             // The decisions ARE gone on this path, so hasDecisions goes false
             // and the Reopen button unmounts. Retrying is the right advice and
-            // the cleanup is idempotent, so the retry has to come with it --
-            // and it must not expire, because when it goes there is no other
-            // way back to this action short of re-resolving the report.
+            // the cleanup is idempotent, so the retry has to come with it, and
+            // it does not expire on a timer. That is not a guarantee it will
+            // still be there: TOAST_LIMIT is 1, so any later toast evicts it.
+            // The fallback is to resolve the report again and reopen.
             action: (
               <ToastAction altText="Retry the reopen" onClick={() => reopenMutation.mutate()}>
                 Try again
--- a/src/components/ReportDetail.tsx
+++ b/src/components/ReportDetail.tsx
@@ -242,11 +242,11 @@ export function ReportDetail({ report, allReportsForTarget, allReports = [], onD
     mutationFn: async () => {
       if (!context.target) throw new Error('No target');
       // Delete all decisions for this target
-      const primary = await deleteDecisions(context.target.value);
+      const primary = await deleteDecisions(context.target.value, context.target.type);
       // Also delete decisions for the pubkey if this is an event report
       let secondary = { labelCleanupFailed: false };
       if (context.target.type === 'event' && context.reportedUser.pubkey) {
-        secondary = await deleteDecisions(context.reportedUser.pubkey);
+        secondary = await deleteDecisions(context.reportedUser.pubkey, 'pubkey');
       }
       return { labelCleanupFailed: primary.labelCleanupFailed || secondary.labelCleanupFailed };
     },
--- a/src/components/ReportDetail.tsx
+++ b/src/components/ReportDetail.tsx
@@ -280,9 +280,15 @@ export function ReportDetail({ report, allReportsForTarget, allReports = [], onD
             // The decisions ARE gone on this path, so hasDecisions goes false
             // and the Reopen button unmounts. Retrying is the right advice and
             // the cleanup is idempotent, so the retry has to come with it, and
-            // it does not expire on a timer. That is not a guarantee it will
-            // still be there: TOAST_LIMIT is 1, so any later toast evicts it.
-            // The fallback is to resolve the report again and reopen.
+            // it does not expire on a timer.
+            //
+            // It is still single-use. Radix renders ToastAction as a
+            // ToastClose, so the click that fires the retry also dismisses this
+            // toast. A retry that comes back still-degraded raises a fresh one
+            // carrying a fresh action; a retry that THROWS lands in onError,
+            // which has no action, and ends the chain. TOAST_LIMIT is 1, so any
+            // later toast evicts it too. The fallback in both cases is to
+            // resolve the report again and reopen.
             //
             // Re-sends the target this toast was raised for, not whatever is
             // selected when it is clicked -- see ReopenTarget above.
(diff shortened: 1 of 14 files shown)


src/components/Reports.tsx
```diff
@@ -1001,6 +1010,31 @@ export function Reports({ relayUrl, selectedReportId }: ReportsProps) {
             </div>
           </div>
 
+          {/* Refresh failed but we still hold a previous list: warn instead of
+              blanking the queue. The poll keeps retrying every 15s. */}
+          {error && (
```
realmeylisdev, line 1017: **No negative-side test.** Flipping this to `{true && (` leaves the whole suite green (508 passed, 1 skipped), so an always-on "Live refresh is failing" strip would ship without a test noticing.

Commit 5 added exactly this pin for the resolution-state banner — *"a warning that is always on is a warning nobody reads."* Same one-liner applies here; the healthy-path test that already renders a loaded queue is the natural home:

```js
expect(screen.queryByText(/refresh is failing/i)).not.toBeInTheDocument();
```

src/components/ReportDetail.tsx
```diff
@@ -242,20 +242,33 @@ export function ReportDetail({ report, allReportsForTarget, allReports = [], onD
...
+      queryClient.invalidateQueries({ queryKey: ['resolution-labels'] });
       decisionLog.refetch();
       pubkeyDecisionLog.refetch();
-      toast({
-        title: "Report reopened",
-        description: "This report is now back in the pending queue",
-      });
+      // The decisions are gone either way, but a surviving resolution label
+      // keeps the report out of the queue. Do not claim it is back.
+      toast(labelCleanupFailed
```
realmeylisdev, line 273: **This branch is untested.** Swap `labelCleanupFailed` for `false` here — always reporting a clean reopen — and the full suite stays green (508 passed, 1 skipped).

The worker setting the flag and `deleteDecisions` surfacing it are both well covered, so the chain is pinned everywhere except the hop where the moderator actually learns the report may still be hidden. That hop is the whole point of the flag: a reopen that says "back in the pending queue" when the label survived is the bug this PR set out to close.

`ReportDetail` has no test file today, so this needs a new one — the destructive toast on `labelCleanupFailed: true`, the clean one on `false`.

src/components/Reports.tsx
```diff
@@ -397,12 +397,16 @@ export function Reports({ relayUrl, selectedReportId }: ReportsProps) {
...
+  // failed fetch therefore makes the queue bigger and wrong, not smaller and
+  // safe, so the error has to reach the UI (#221). One retry, because a single
+  // slow relay read should not cost a moderator their whole resolution filter.
+  const { data: resolutionLabels, error: labelsError } = useQuery({
     queryKey: ['resolution-labels', relayUrl],
     queryFn: fetchResolutionLabels,
     refetchInterval: 15 * 1000,
     placeholderData: (previousData) => previousData,
-    retry: false,
+    retry: 1,
```
realmeylisdev, line 411: Minor, non-blocking.

This makes the comment at line 451 inaccurate — it documents `retry: false` as deliberate *"across all polling queries in this component"*, which no longer holds. Worth a word there so the next reader doesn't "restore" the pattern.

And a question rather than a request: the retry went to the labels query, but the latency you measured is on `/api/reports` (p50 2831ms, p90 3177ms, max 4595ms against the 5s cap) — and that one keeps `retry: false`. Deliberate? A retry there would also push the stale-data banner out by the backoff, so I can see it either way; just want to know it was weighed.

realmeylisdev (changes requested):
Reviewed the worker change against the funnelcake relay implementation and NIP-01, and ran every gate locally.

The diagnosis holds. `queryRelay` really did resolve `success: true` on both timeout and close-before-EOSE, and with the 15s poll replacing data wholesale that is enough to blank a populated queue. The additive/subtractive split is the right call too — erroring the labels read would hide nothing and grow the queue, so surfacing it beats failing it.

**Two claims I checked rather than took on trust:**

- **NIP-01 `CLOSED`** — the spec says `CLOSED` MUST be sent when a relay refuses to fulfil a `REQ`. Treating any `CLOSED` as a failed read is correct, and being reason-agnostic is the right shape.
- **The relay claim** — funnelcake sends `CLOSED` with `error: could not complete query` on a store failure and never sends EOSE (`crates/relay/src/handler.rs:29,544,637`). It has a *second* reason you don't mention, `error: stored replay timed out`, emitted when a partial replay or the EOSE send itself fails — that path also skips EOSE, and your branch catches it. No path reports a query failure *after* EOSE, so "success implies EOSE" holds. The one EOSE-then-`CLOSED` case (subscription registration rejected after a complete replay) resolves as a success, which is right — the stored read did finish.

Gates green locally: worker tsc, frontend tsc, worker 376, worker D1 21, frontend 508 + 1 skipped. Removing `isUnconfirmedTargetedMiss` checks out — no callers left, and a targeted empty-timeout still reaches `unavailable` via the 502.

## Requested changes

I ran my own mutation pass over the new guards — thirteen mutations, eleven die as you'd expect. **Two survive the entire suite**, both on the reporting side:

1. The stale-data banner has no negative-side test — an always-on "Live refresh is failing" strip ships green.
2. The degraded reopen toast has no test — always claiming a clean reopen ships green.

Details inline. Neither is a code defect: the logic is right, the tests just don't hold it in place. But #2 is the exact hop this PR exists to fix, and #1 is the check commit 5 already added for the sibling banner, so it reads as an oversight rather than a decision. Worth noting the PR description's "twelve mutations, twelve confirmed failures" doesn't cover these two.

Happy to push both tests myself if you'd rather not write them — say the word.

One minor, non-blocking note inline on the labels `retry: 1`. Nothing else blocking.

realmeylisdev (approved):
Supersedes my earlier request for changes. Both blockers landed, and then some — the reopen suite now covers a hazard neither of us had named.

**The two I asked for:**

- The stale-data banner has its negative side (`Reports.test.tsx:139` asserts it is absent on a healthy load), plus the cold-failure control at `:149` — an always-on strip no longer ships green.
- The degraded reopen toast is pinned at `ReportDetail.reopen.test.tsx:179`, with both halves of the cleanup OR held separately, so returning `secondary.labelCleanupFailed` alone now dies.

## What I pushed

Ten commits during takeover, plus one more today. One commit per finding, each revertable alone.

The one worth your attention is `4c32dd3`. The degraded toast carries `duration: Infinity`, so it outlives the selection that raised it, and selecting another report re-renders `ReportDetail` rather than remounting it. Because `mutate()` rebuilds from the observer's latest options, a `mutationFn` reading the target off `context` sent the retry at whichever report was selected when it was clicked — deleting an unrelated report's decisions and banning its resolution labels, then reporting a clean reopen. The target is now mutation variables captured at click time. `ReportDetail.reopen.test.tsx:230` fails against the old `mutationFn`.

`85c654d` and `1714dd0` are the other behavioral ones: a `close()` that throws used to strand the promise with the timeout already cleared, and on the two exits inside the message listener the parse-error catch swallowed it, so `queryRelay` hung and `/api/reports` never answered. Those tests hang to the runner's timeout against the old code rather than merely failing.

Today's `3392486` is comment-only. It corrects a comment I wrote in `df23096`, which credited the `resolved` guard for the late-frame behavior on the failure path and called the `events.slice()` copies redundant. Mutation-checked in both directions, baseline 75 passed:

```
guard deleted, slice kept  -> 1 failed | 74 passed   (only the after-EOSE test dies)
slice dropped, guard kept  -> 75 passed
both removed               -> 2 failed | 73 passed
```

The success exit hands back the live `events` array, so the guard is its only cover — which is why the after-EOSE case is the one that dies to the guard alone. The four socket failure exits hand back a snapshot, so on that path the mechanisms are independently sufficient and the test dies only to their conjunction. The old comment invited deleting the copies as dead weight with the suite staying green.

## What I checked beyond the gates

I ran an adversarial pass over my own takeover commits, since nobody else had reviewed them — four lenses (logic, test efficacy, prose claims, regression against your commits), with every candidate finding sent to an independent refutation step and mutations run in isolated worktrees.

No correctness defects, no regressions against your commits, and no inert tests. Two claims that looked alarming and turned out to be wrong, both worth recording:

- That the two-filter fallback blows the 30s client abort across its whole range. It cannot: `publishLabel` tags a label `e` **or** `p`, never both, so for a given target only one filter can match — the other reads zero events and performs zero bans. The doubled cost is unreachable.
- That the takeover had left the reopen toast copy stale. The copy reached its current form in your `37911d4`; what was stale was the PR description quoting it.

## Description corrected — please check my edit

I edited the description directly, which touches your prose, so flagging it plainly. Four spans, nothing else:

- **Label cap** said "it is now 200". The shipped constant is `LABEL_CLEANUP_LIMIT = 50` (your `37911d4` lowered it), and the comment above it says a cap of 200 blows the client abort outright — so the description was advertising the number the code calls unsafe.
- **Both reopen toast rows** quoted copy that does not ship. The degraded row quoted "The relay did not clear them…", which `ReportDetail.reopen.test.tsx:192` asserts must *not* ship; the clean row omitted the ban/deletion caveat that is the entire point of the change.
- **The `Try again` affordance** (`duration: Infinity`) was missing from the UI table, though it is the branch's one new interactive element.
- **Gate counts** trailed by five tests, which was my doing.

Revert any of it if you'd rather say it your way.

## Still open

One thing I could not fix. `cb3abf2`'s message says "Four is the count against the pre-branch code, not against the parent commit." At the merge-base `b0af9bf`, `close()` runs before `resolve()` on exactly **two** exits — timeout and EOSE — and there is no `CLOSED` branch at all; four was only ever true at `37911d4`. The shipped code comment says three and is correct, so this is confined to a commit message, and fixing it means a history rewrite I did not want to do unilaterally. Amend it if you rebase for #222 anyway.

Two items are yours: the staging screenshots of the three states, which the validation plan still lists as pending, and the merge order — this wants to land before #222, whose `relay-profile.ts:134` reads the `complete` field this PR removes.

## Gates

Frontend 522 passed + 1 skipped; worker 395 passed; worker D1 21 passed; frontend build, both typechecks, and eslint clean. Run against `3392486`.

Approving the resulting state. The merge call stays yours, including on everything I pushed.

hm21 (approved):
LGTM

PR #188, opened 2026-07-16
Content report note: carry its own context (nip05, event type, restored-OG flag)

mbradley:
Closes #187.

## Why
Working the support queue, we found the content-report internal note in Zendesk had lost clarity as the context around it shifted. On ticket #3991 it showed only a bare hex "Reported Author" line — an agent couldn't tell which account was reported, that the pubkey was the subject (not the reporter), or that the "impersonator" was actually the reporter's own restored OG Vine account. Establishing that took a manual investigation.

## What
The note now carries its own context regardless of who reads it:
- Clear heading + "open the full report in Relay Manager" CTA.
- Explicit "reported subject, not the reporter" labeling.
- Scope (account vs content) and the reported event's kind (video / comment / note / ...).
- Human-legible identity: profile name + nip05, alongside npub and hex.
- A restored-OG-Vine-account flag when the subject is a vine-archive import — the exact signal that would have resolved #3991 at a glance — pointing to the "Recovered OG account" macro with a confirm-first caveat.
- Direct lookups: Relay Manager report view, login.divine.video support-admin (?q=), and the public divine.video profile.

## How
- Note formatting is a pure, unit-tested module (report-note.ts).
- handleParseReport fetches the subject's kind-0 and the reported event's kind best-effort via queryRelay, in parallel, each capped at 3s; on a slow/unreachable relay the note still posts with pubkey/npub/links, minus enrichment.
- Profile fields (name / nip05 / vine_username) come from the reported account's own kind-0, so they're sanitized before interpolation (control chars/newlines to space; markdown link/image/span/emphasis/html stripped) to stop the reported party forging note structure or injecting links.
- Event-kind labels (comment = 1111, video = 34235/34236) verified against divine-web and moderation-service source.

## Rendered example (account report, restored OG account)

    **🛡️ RELAY REPORT — INTERNAL LOOKUP**

    Reported on the Divine relay. Full report details (reporter, reason, timeline, prior actions) live in Relay Manager:
    **→ [Open this report in Relay Manager](https://relay.admin.divine.video/reports?pubkey=9f59c820...&env=production)**

    Reason: **other** · Scope: **account-level (whole profile)**

    ---

    **Reported account** — the subject of this report, *not* the person who filed it:
    • Profile: **kofi** · nip05 `_@kofi.divine.video`
    • ⚠️ **Restored OG Vine account** (vine_username `kofi`) — self-reported import markers; likely a name mix-up, not impersonation, but confirm in Relay Manager before applying the "Recovered OG account" macro
    • Look up in login.divine.video → [support-admin](https://login.divine.video/support-admin?q=9f59c820...)
    • Public profile → https://divine.video/profile/9f59c820...
    • npub: `npub1navusg9...`
    • hex: `9f59c820...`

    **Filed by:** the ticket requester above.

## Known limitation (accepted)
The `support-admin?q=<hex>` lookup link becomes one-click only once keycast divinevideo/keycast#299 is deployed to prod. Until then it lands on a blank support-admin search box and the agent pastes the hex/npub — both printed in the note. That is a graceful degradation, not a break, and the whole note is a large net improvement over the version that has been augmenting these tickets since December. We accept this transitional state; #299 is tracked and lands the one-click behavior.

## Testing
- report-note unit tests cover account / content / no-enrichment / staging + kind-0 parsing + sanitization; full worker suite 330/330; typecheck clean.
- The injection sanitization was verified end-to-end against adversarial kind-0 input.
- Not yet exercised against a live ticket (relay fetch + Zendesk render); planned at deploy.

### DIFF
--- a/worker/src/report-note.ts
+++ b/worker/src/report-note.ts
@@ -23,8 +23,10 @@ export interface ReportNoteInput {
   authorPubkey: string | null;
   /** Report category, verbatim from the ticket (e.g. "other", "sexualContent"). */
   violationType: string | null;
-  /** Worker environment, drives link env params + keycast host. */
+  /** Worker environment, drives the relay-admin `env` link param. */
   environment?: string;
+  /** Configured Keycast base URL (env.KEYCAST_URL); the note's lookup host + link derive from it. */
+  keycastUrl?: string;
   /** Reported account's profile (enrichment); null/undefined if the fetch failed. */
   profile?: ReportedProfile | null;
   /** Kind of the reported event (enrichment); null/undefined if unknown. */
@@ -115,10 +117,17 @@ function toNpub(hex: string): string {
   }
 }
 
-function keycastBase(environment?: string): string {
-  return environment === 'staging'
-    ? 'https://login.staging.dvines.org'
-    : 'https://login.divine.video';
+/** Normalize the configured Keycast URL and derive its visible host, so the note's
+ *  label and link always agree with the deployed environment. */
+function keycastTarget(keycastUrl?: string): { url: string; host: string } {
+  const url = (keycastUrl || 'https://login.divine.video').replace(/\/+$/, '');
+  let host = url;
+  try {
+    host = new URL(url).host;
+  } catch {
+    // Not a parseable URL: fall back to showing the raw value.
+  }
+  return { url, host };
 }
 
 /**
@@ -131,6 +140,7 @@ function keycastBase(environment?: string): string {
 export function buildReportNote(input: ReportNoteInput): string {
   const { eventId, authorPubkey, violationType, environment, profile, reportedEventKind } = input;
   const envParam = environment ? `&env=${environment}` : '';
+  const keycast = keycastTarget(input.keycastUrl);
   const scope = eventId ? 'content (specific event)' : 'account-level (whole profile)';
   const primaryLink = eventId
     ? `${RELAY_ADMIN}/reports?event=${eventId}${envParam}`
@@ -155,18 +165,19 @@ export function buildReportNote(input: ReportNoteInput): string {
     lines.push('**Reported account** — the subject of this report, *not* the person who filed it:');
     if (profile?.name || profile?.nip05) {
       const bits: string[] = [];
-      if (profile.name) bits.push(`**${profile.name}**`);
-      if (profile.nip05) bits.push(`nip05 \`${profile.nip05}\``);
+      // Code span (not bold) so a name containing a bare URL can't auto-link in Zendesk.
+      if (profile.name) bits.push(`\`${profile.name}\``);
+      if (profile.nip05) bits.push(`nip05 \`${profile.nip05}\` (claimed, unverified)`);
       lines.push(`• Profile: ${bits.join(' · ')}`);
     }
     if (profile?.isVineImport) {
       const uname = profile.vineUsername ? ` (vine_username \`${profile.vineUsername}\`)` : '';
       lines.push(
-        `• ⚠️ **Restored OG Vine account**${uname} — self-reported import markers; likely a name mix-up, not impersonation, but confirm in Relay Manager before applying the "Recovered OG account" macro`,
+        `• 🏷️ **Vine-archive import markers**${uname} — self-reported on the account's own kind-0, so a signal, not proof. Verify the reporter's ownership before resolving; if confirmed, the "Recovered OG account" macro applies.`,
       );
     }
     lines.push(
-      `• Look up in login.divine.video → [support-admin](${keycastBase(environment)}/support-admin?q=${authorPubkey})`,
+      `• Look up in ${keycast.host} → [support-admin](${keycast.url}/support-admin?q=${authorPubkey})`,
     );
     lines.push(`• Public profile → https://divine.video/profile/${authorPubkey}`);
     lines.push(`• npub: \`${toNpub(authorPubkey)}\``);
--- a/worker/src/report-note.ts
+++ b/worker/src/report-note.ts
@@ -6,7 +6,7 @@ import { nip19 } from 'nostr-tools';
 
 /** Profile facts pulled from the reported account's kind-0 event (best-effort enrichment). */
 export interface ReportedProfile {
-  /** display_name || name from kind-0 content, if any. */
+  /** display_name || name from kind-0 content, if any (sanitized for note interpolation). */
   name?: string;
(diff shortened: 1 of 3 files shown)


worker/src/report-note.ts
```diff
@@ -0,0 +1,178 @@
...
+    if (profile?.name || profile?.nip05) {
+      const bits: string[] = [];
+      if (profile.name) bits.push(`**${profile.name}**`);
+      if (profile.nip05) bits.push(`nip05 \`${profile.nip05}\``);
+      lines.push(`• Profile: ${bits.join(' · ')}`);
+    }
+    if (profile?.isVineImport) {
+      const uname = profile.vineUsername ? ` (vine_username \`${profile.vineUsername}\`)` : '';
+      lines.push(
+        `• ⚠️ **Restored OG Vine account**${uname} — self-reported import markers; likely a name mix-up, not impersonation, but confirm in Relay Manager before applying the "Recovered OG account" macro`,
```
dcadenas: Remove the conclusion at worker/src/report-note.ts:165 that this is likely a name mix-up rather than impersonation. `isVineImport` comes only from tags on the reported account’s own kind-0, so the account can self-assert the signal and the suggested Relay Manager check reads the same data. Keep the markers, vine username, and confirmation instruction, but present them as evidence without pre-judging the report.

worker/src/report-note.ts
```diff
@@ -0,0 +1,178 @@
...
+    if (reportedEventKind != null) lines.push(`• Kind: \`${reportedEventKind}\``);
+    lines.push(`• Event ID: \`${eventId}\``, '');
+  }
+
+  if (authorPubkey) {
+    lines.push('**Reported account** — the subject of this report, *not* the person who filed it:');
+    if (profile?.name || profile?.nip05) {
+      const bits: string[] = [];
+      if (profile.name) bits.push(`**${profile.name}**`);
+      if (profile.nip05) bits.push(`nip05 \`${profile.nip05}\``);
```
dcadenas: Qualify the NIP-05 at worker/src/report-note.ts:159 as claimed and unverified, for example `nip05 `${profile.nip05}` (claimed, unverified)`. This value comes from the account’s own kind-0 and is not resolved against `/.well-known/nostr.json`, unlike the npub and hex shown below. Keep the bound identifiers unchanged and update the corresponding test.

worker/src/report-note.ts
```diff
@@ -0,0 +1,178 @@
...
+      lines.push(`• Profile: ${bits.join(' · ')}`);
+    }
+    if (profile?.isVineImport) {
+      const uname = profile.vineUsername ? ` (vine_username \`${profile.vineUsername}\`)` : '';
+      lines.push(
+        `• ⚠️ **Restored OG Vine account**${uname} — self-reported import markers; likely a name mix-up, not impersonation, but confirm in Relay Manager before applying the "Recovered OG account" macro`,
+      );
+    }
+    lines.push(
+      `• Look up in login.divine.video → [support-admin](${keycastBase(environment)}/support-admin?q=${authorPubkey})`,
```
dcadenas: Derive the visible host and the href at worker/src/report-note.ts:169 from the configured Keycast URL. On staging the current note says `login.divine.video` while linking to `login.staging.dvines.org`, which makes the support instruction self-contradictory. Pass `env.KEYCAST_URL` into the note input, preserve the local fallback, and assert label/href agreement in the staging test.

worker/src/report-note.ts
```diff
@@ -0,0 +1,178 @@
...
+ * kept so legitimate nip05 values (e.g. `_@name.divine.video`) survive intact.
+ */
+function sanitizeInline(value: string | undefined, maxLen = 80): string | undefined {
+  if (!value) return undefined;
+  const cleaned = Array.from(value, (ch) => {
+    const code = ch.charCodeAt(0);
+    return code < 0x20 || code === 0x7f ? ' ' : ch; // control chars / newlines -> space
+  })
+    .join('')
+    .replace(/[`[\]*<>|]/g, '') // markdown link/image/span/emphasis + html/table
```
dcadenas, line 69: Make the sanitizer guarantee true for bare URLs. The current `sanitizeInline` at worker/src/report-note.ts:67 leaves a URL in `display_name` intact, and worker/src/report-note.ts:158 then renders it bolded as the reported account’s name. Wrap the name in a code span like the adjacent NIP-05, or defang URL schemes in the sanitizer, and add a test that asserts the bare-URL case is inert.

dcadenas (changes requested):
The report note currently treats self-asserted Vine markers as enough to call an impersonation report a name mix-up, presents an unverified NIP-05 as identity evidence, and labels a staging link as the production Keycast host. Its sanitizer also allows bare URLs in reported names despite claiming to prevent link injection. Remove the unsupported conclusion, qualify the NIP-05, derive the label from the configured Keycast URL, and make the name rendering match the sanitizer’s documented safety guarantee before merging.

worker/src/report-note.ts
mbradley: Done (d4699c6). Dropped the "name mix-up, not impersonation" conclusion. The line now presents the import markers + vine_username as self-reported evidence ("a signal, not proof") and instructs to verify the reporter's ownership before resolving; the "Recovered OG account" macro is referenced only conditionally ("if confirmed"). No independent-check claim on Relay Manager.

worker/src/report-note.ts
mbradley: Done (d4699c6). nip05 now renders `nip05 \`<value>\` (claimed, unverified)`; the bound npub/hex identifiers are unchanged, and the staging/account tests are updated.

worker/src/report-note.ts
mbradley: Done (d4699c6). Added `keycastUrl` to the note input; `handleParseReport` passes `env.KEYCAST_URL` (typed via `Env extends KeycastEnv`, set per-env in wrangler.{staging,prod}.toml). `keycastTarget()` derives both the visible host and the href from it (login.divine.video fallback), so on staging both read `login.staging.dvines.org`. The staging test asserts label/href agreement and `not.toContain('login.divine.video')`.

worker/src/report-note.ts
mbradley: Done (d4699c6). The reported name now renders in a code span (`` `${name}` ``, like the adjacent nip05) instead of bold, so a bare URL in the name can't auto-link in Zendesk. `sanitizeInline` still strips backticks, so a name can't break out of the span. Added a "renders a bare URL in the name inertly" test.

dcadenas (approved):
✅ self-contained report notes, ship it

PR #194, opened 2026-07-21
fix(nip98): accept allowlisted public host on the mobile minor-review endpoints (#173)

mbradley:
## What this does

Part of #173 (the worker half). The two mobile minor-review endpoints authenticate with NIP-98, which binds each signed request to an exact URL. Requests that older app installs sign for the public host `api.divine.video` reach the worker on its own host (`api-relay-prod.divine.video`), so the exact-URL check rejects them and the minor-review gate does not take effect on those devices. This change lets the worker accept a NIP-98 signature made for a configured public host in addition to its own host, so the gate reaches the existing install base once the routing half is in place.

## Approach

`verifyNip98Auth` keeps its exact-string URL match as an unchanged fast path. Only when the strings differ does it fall back to an additive check that relaxes the host and nothing else: scheme, path, and query must still match the worker's own request exactly, and the signed host must be a member of a per-environment allowlist. An empty or unset allowlist preserves the original strict same-host behaviour, so this does not loosen anything by default.

The allowlist is threaded into only the two mobile endpoints (`GET /v1/account/moderation-status` and `POST /v1/minor-review-cases/*/parent-contact`). The Zendesk pre-auth endpoint keeps strict same-host checking and is left unchanged.

## Config

New non-secret var `NIP98_PUBLIC_HOST_ALLOWLIST` (comma-separated hostnames):
- prod: `api.divine.video`
- staging: empty, pending confirmation of the staging public host
- local: empty

## Scope and dependencies

- The routing half of #173 (directing `api.divine.video/v1/*` to this worker) is infra work in another repo and is not in this PR. This worker change stays inert until that route exists.
- Staging stays inert until its public host is set.

## Testing

New unit tests cover the host allowlist, the scheme/path/method binding, freshness, signature, a userinfo-host rejection, mixed-case config, and the empty-allowlist default. Router tests exercise the wiring end to end, and a route-level test confirms the Zendesk endpoint still rejects an allowlisted public host. Full worker suite green (unit and D1), typecheck and lint clean.

## Follow-ups (tracked separately)

This PR closes the specific fail-open gap for existing installs. Three related hardening items are filed and complete the picture:

- #195 — a single-use nonce for NIP-98 replay protection (the 60s freshness window is the only replay guard today, and relaxing the host slightly widens it).
- #196 — bind the POST parent-contact body with a NIP-98 payload hash (identity, URL, and method are bound today; the body is not).
- #197 — make the moderation-status DB-unavailable fail-open observable and alerted (done, separate PR).

#195 and #196 modify the same `verifyNip98Auth` function as this PR and are sequenced after it to avoid stacking.

### DIFF
--- a/worker/src/index.test.ts
+++ b/worker/src/index.test.ts
@@ -937,6 +937,18 @@ describe('mobile NIP-98 endpoint host allowlist (#173)', () => {
     expect(res.status).toBe(200);
   });
 
+  // Config is user-edited free text (unlike URL.hostname, which the platform always
+  // lowercases), so a mixed-case entry must still normalize to match at compare time.
+  it('accepts a public host configured with mixed case (config is case-insensitive)', async () => {
+    vi.spyOn(console, 'warn').mockImplementation(() => {});
+    const res = await worker.fetch(
+      new Request(OWN_HOST_URL, { method: 'GET', headers: { Authorization: nip98Header(PUBLIC_HOST_URL) } }),
+      { NIP98_PUBLIC_HOST_ALLOWLIST: 'API.Divine.Video' } as never,
+      ctx,
+    );
+    expect(res.status).toBe(200);
+  });
+
   it('accepts an own-host-signed request with no allowlist configured (regression)', async () => {
     vi.spyOn(console, 'warn').mockImplementation(() => {});
     const res = await worker.fetch(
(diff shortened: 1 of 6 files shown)


dcadenas (approved):
🔒 allowlisted host auth is wired cleanly

Issue #196, opened 2026-07-21
fix(nip98): bind parent-contact POST body to its signature


mbradley:
## Problem

The mobile `POST /v1/minor-review-cases/{caseId}/parent-contact` request is NIP-98 authenticated for its URL and method, but the current worker does not verify a NIP-98 `payload` tag. The JSON body contains the parent email, so an authenticated request can have its body changed without invalidating the signature.

## Required work

- Update the mobile client and relay-manager worker to agree on the NIP-98 payload contract for this endpoint.
- The mobile client must serialize the JSON request body once, compute SHA-256 over the exact UTF-8 bytes it sends, and include that digest in the NIP-98 event `payload` tag. Do not hash a separately reconstructed or differently encoded JSON value.
- In the worker, require and verify the payload tag against the exact raw request-body bytes for `POST /v1/minor-review-cases/{caseId}/parent-contact`, in addition to the existing NIP-98 signature, URL, and method checks. Reject a missing, malformed, or mismatched digest before calling `handleParentContact` or making any state change.
- Preserve the verified body bytes for the existing JSON parser without consuming the request stream before the handler needs it.
- Keep GET requests, including moderation-status GET, unaffected; they have no request body and must not require a payload tag. Keep unrelated endpoints unchanged.

## Acceptance criteria

- A correctly signed parent-contact POST with a payload hash matching the exact transmitted JSON body passes authentication and retains the existing endpoint behavior.
- A POST with a missing, malformed, or incorrect payload tag is rejected before parent email or case state is written.
- Tests cover matching and mismatching body hashes, missing payload tags, malformed tags, and body parsing after verification.
- Tests confirm existing GET requests remain valid without a payload tag and unrelated NIP-98 endpoints are unchanged.
- The mobile test proves the hash is calculated from the exact bytes sent on the request.

## Scope boundary

This issue covers the JSON `parent-contact` endpoint only. It does not implement or define multipart signing for the parent-consent video upload; that work is tracked separately in [relay-manager#281](https://github.com/divinevideo/divine-relay-manager/issues/281).

Related to #173.

NotThatKindOfDrLiz:
Rollout revision:

1. Ship the mobile client fix first: [divine-mobile PR #9516](https://github.com/divinevideo/divine-mobile/pull/9516) serializes the parent-contact JSON once and signs the same bytes it sends.
[...]

PR #198, opened 2026-07-21
feat(age-review): alert on the moderation-status DB-unavailable fail-open (#197)

mbradley:
## What this does

Closes #197. When the worker's D1 binding is unavailable, `handleGetModerationStatus` deliberately fails open (returns an unrestricted `active` status) so a database problem does not lock out normal users. Until now that happened silently, so the minor-review gate could be failing open with no signal. This keeps the fail-open behaviour but makes it observable.

## Approach

- The request-path fail-open branch now logs a distinct, greppable error marker (`MODERATION_STATUS_DB_UNAVAILABLE`) instead of a quiet warning, so log-based monitoring can alert on it. The returned status and body are unchanged.
- The scheduled cron now sends a Slack alert (reusing the existing age-review alert path) when the DB binding is absent. It fires from the cron rather than per request, so it cannot storm, and a Slack failure cannot disrupt the rest of the scheduled work.

Keeping fail-open rather than switching to fail-closed was a deliberate decision: the client already fails open when the server is unreachable, so failing closed here would add little protection at the cost of locking users out during any outage. Observability was the real gap.

## Scope

Covers the binding-absent (`!env.DB`) case, which is the condition that produces the fail-open. A transient D1 query failure already surfaces as a 500 and is out of scope here.

## Relation to #173

Part of the minor-safety hardening surfaced alongside #173 (the public-host allowlist fix). Sibling follow-ups #195 (replay nonce) and #196 (payload-hash binding) remain open.

## Testing

New tests cover the request-path marker, the Slack alert helper, and the cron branch (fires only when the DB binding is absent and a webhook is configured, never when the DB is present). Full worker suite green, typecheck and lint clean.

### DIFF
--- a/worker/src/age-review.test.ts
+++ b/worker/src/age-review.test.ts
@@ -9,6 +9,7 @@ import {
   handleAgeReviewReplyWebhook,
   handleCreateMinorAccount,
   checkAgeReviewDeadlines,
+  sendDbUnavailableAlert,
   syncAgeReviewTicketResolution,
   getAgeReviewConfig,
   updateAgeReviewConfig,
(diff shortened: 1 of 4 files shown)


dcadenas (approved):
🚨 DB-unavailable alerts now carry environment context

PR #199, opened 2026-07-22
feat(age-review): account-type indicator for moderation cases

mbradley:
## Summary

Adds an **account-type indicator** to the age-review case view: it tells the moderator whether a reported target is a Divine keycast account, a self-custody pubkey we hold no login lever over, or unknown (keycast unavailable) — and states plainly what enforcement is actually effective, so the tool never implies an action that can't land.

## Motivation

From the 2026-07-21 legacy-reset investigation and moderator feedback:
- Much of the reported-under-16 backlog is self-custody pubkeys with no Divine login — "suspend sign-in" does nothing there, but the tool implied it was available.
- Our own suspension hides content, so an already-enforced account can read as "nothing here."
- Partial/incidental prior enforcement (e.g. a single event-ban) looks the same as a fully-compliant age-review outcome.

Design spec + plan: `support-trust-safety/docs/superpowers/specs/2026-07-22-age-review-account-type-indicator-design.md` and the matching plan.

Closes #191.

## What it does

- Backend: `getUserStatus` / `/api/account-status` gain a structured `not_found` (self-custody) state, distinct from `error` (unavailable). The 404 path now requires a Keycast user-not-found body so route/config 404s remain unavailable.
- Pure `deriveAccountVerdict` helper: maps account-status + relay content-presence + ticket into an account type, a compliance-leg checklist, and a one-line "what's effective" verdict — never advertising an action that can't land, never under-stating a real one.
- `AccountTypeIndicator` component + `useAccountStatus` hook, wired into `AgeReviewDetail` (content-presence via `useUserStats`).

## Scope

- Account-type badge + effectiveness verdict + content-presence note + a partial compliance checklist (sign-in and ticket legs truthfully sourced; content legs show applicability + "not tracked yet").
- Addresses self-custody as a not-applicable/informational leg, not a Keycast outage.
- Deferred to #123: content-leg *done*-tracking needs durable per-leg enforcement persistence.

## Manual validation

Rendered all seven states live on the local stack: divine active/suspended, self-custody with/without content, unknown, cold-load "checking", and an unresolved-read case. Confirmed self-custody reads as informational (not an error), the checklist reflects done/missing/n-a/not-tracked, and an indeterminate content read says "verify content directly" rather than falsely concluding "no effective enforcement."

## Testing

- `npx vitest run src/lib/accountVerdict.test.ts src/components/AccountTypeIndicator.test.tsx src/components/AgeReviewDetail.test.tsx`
- `cd worker && npx vitest run src/keycast-client.test.ts src/account-status.test.ts`
- `npm run test`

Note: `cd worker && npx vitest run` currently fails in existing unrelated `age-review.test.ts` funnel mocks and `bulk-moderate.test.ts` delete-all expectations; the touched worker account-status/keycast tests pass.

### DIFF
--- a/src/lib/accountVerdict.ts
+++ b/src/lib/accountVerdict.ts
@@ -14,7 +14,7 @@ export interface ComplianceLeg {
 export interface AccountVerdict {
   accountType: AccountType;
   suspended: boolean;
-  contentPresence: 'visible' | 'hidden_suspended' | 'none';
+  contentPresence: 'visible' | 'hidden_suspended' | 'none' | 'unknown';
   legs: ComplianceLeg[];
   verdict: string;
 }
@@ -23,11 +23,15 @@ export interface DeriveInput {
   accountStatus: AccountStatusResponse | undefined;
   accountStatusError: boolean;
   postCount: number | undefined;
+  // Whether the relay content read has resolved. A failed/in-flight read leaves
+  // postCount undefined, which must NOT be read as "no content" — that would
+  // under-state available enforcement (the content lever silently dropped).
+  contentPresenceKnown: boolean;
   ticketLinked: boolean;
 }
 
 export function deriveAccountVerdict(input: DeriveInput): AccountVerdict {
-  const { accountStatus, accountStatusError, postCount, ticketLinked } = input;
+  const { accountStatus, accountStatusError, postCount, contentPresenceKnown, ticketLinked } = input;
 
   // Account type: a query error, OR success:false without not_found, is "unknown"
   // (keycast unavailable). not_found is self-custody; success is a Divine account.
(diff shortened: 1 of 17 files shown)


src/lib/accountVerdict.ts
```diff
@@ -0,0 +1,128 @@
...
+    accountType = 'unknown';
+  } else if (accountStatus?.not_found) {
+    accountType = 'self_custody';
+  } else if (accountStatus?.success) {
+    accountType = 'divine';
+  } else {
+    accountType = 'unknown';
+  }
+
+  const suspended = accountStatus?.status === 'suspended';
```
dcadenas, line 51: Normalize account-status semantics in one shared path before deriving the checklist and header.
At `src/lib/accountVerdict.ts:49`, a `banned` response is treated as sign-in active because only `suspended` counts as blocked, while the unavailable branch at `src/lib/accountVerdict.ts:94` discards known content and ticket inputs; `src/components/AgeReviewDetail.tsx:258` also maps `not_found` into the protected-minor unavailable label.
Treat both `suspended` and `banned` as blocked, limit an unavailable Keycast lookup to the sign-in leg, and keep `not_found` out of the unavailable header.
Add focused coverage for each state so the case view and enforcement verdict cannot disagree.

dcadenas (changes requested):
Before merging, normalize the account-status semantics used by the case header and enforcement verdict. The current paths disagree for banned, unavailable, and self-custody states, so update the shared handling and focused tests to keep moderator guidance consistent.

NotThatKindOfDrLiz (approved):
Final takeover pass is clean. The pushed commit addresses the remaining account-status/verdict edge cases, links #191 for closure, and current GitHub checks are green.

dcadenas (approved):
✅ clearer account context for moderators

PR #202, opened 2026-07-23
feat(nip98): single-use replay nonce for the mobile minor-review endpoints (#195)

mbradley:
## What this does

Closes #195. The two mobile minor-review endpoints (`GET /v1/account/moderation-status`, `POST /v1/minor-review-cases/*/parent-contact`) authenticate a NIP-98 event but guarded replay only with the 60-second freshness window, so a captured signed request could be replayed within that window (and, after #173, at more than one host). This adds single-use replay protection, mirroring the Zendesk pre-auth nonce already in this worker.

## Approach

`verifyNip98Auth` stays pure and now also returns the signature-bound `event.id`. Each mobile call site consumes that id via `consumeNip98Nonce`, an atomic `INSERT ... ON CONFLICT(event_id) DO NOTHING RETURNING` against a new `nip98_used_nonces(event_id, expires_at)` table in `ensureSchema`. A returned row means first use (allow); no row means replay (reject, 401). The check sits behind the existing `if (env.DB)`, so it fails open on a DB outage, consistent with the moderation-status endpoint's existing posture. TTL is 120 seconds (twice the freshness window), and the scheduled cleanup purges expired rows.

`event.id` is the sha256 of the canonical event, verified equal to the recomputed hash and covered by the signature, so it is unforgeable and unique per distinct signed request. No client change is needed. The Zendesk pre-auth caller keeps its own nonce and is untouched.

## Scope

Applies to only the two mobile endpoints. The Zendesk pre-auth caller is unchanged (it has its own nonce).

## Testing

A unit test for the returned `event.id`; a real-D1 test proving replay is rejected (atomic insert against Miniflare SQLite, mutation-checked); a route-level end-to-end test that a replayed request through `worker.fetch` with a live DB returns 200 then 401; plus fail-open coverage. Full worker suite green (unit and D1), typecheck and lint clean.

One documented behaviour: two byte-identical GETs signed in the same wall-clock second share an `event.id`, so the second is rejected as a replay. Benign for status polling (a retry the next second succeeds) and desirable for the POST.

## Relation

Part of the #173 minor-safety follow-ups. Sibling item #196 (bind the POST body with a payload hash) remains open.

### DIFF
--- a/worker/src/index.ts
+++ b/worker/src/index.ts
@@ -428,10 +428,21 @@ export default {
         if (env.DB) {
           await ensureSchemaOnce(env.DB);
           // `authResult.eventId` is always set here (it's `event.id`, populated on a
-          // `valid: true` result after verifyEvent) -- the `&&` is TS narrowing
+          // `valid: true` result after verifyEvent) -- the `if` is TS narrowing
           // `eventId?: string` to `string` for consumeNip98Nonce, not a fail-open branch.
-          if (authResult.eventId && !(await consumeNip98Nonce(env.DB, authResult.eventId))) {
-            return jsonResponse({ success: false, error: 'Replayed NIP-98 token' }, 401, corsHeaders);
+          if (authResult.eventId) {
+            // Read-only endpoint: a nonce-storage outage must not turn a status
+            // check into a 500, so only a THROWN error fails open here. A
+            // detected replay (consume resolves `false`) still 401s below --
+            // the asymmetry with the POST route's fail-closed catch is intentional
+            // (#202 review).
+            try {
+              if (!(await consumeNip98Nonce(env.DB, authResult.eventId))) {
+                return jsonResponse({ success: false, error: 'Replayed NIP-98 token' }, 401, corsHeaders);
+              }
+            } catch (e) {
+              console.error('[nip98] nonce storage failed on moderation-status; proceeding fail-open', e);
+            }
           }
         }
         return handleGetModerationStatus(authResult.pubkey, env, corsHeaders);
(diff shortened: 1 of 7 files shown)


worker/src/index.ts
```diff
@@ -424,14 +424,29 @@ export default {
       if (path === '/v1/account/moderation-status' && request.method === 'GET') {
         const authResult = await verifyNip98Auth(request, request.url, getNip98AllowedHosts(env));
         if (!authResult.valid || !authResult.pubkey) return jsonResponse({ success: false, error: authResult.error ?? 'Unauthorized' }, 401, corsHeaders);
-        if (env.DB) await ensureSchemaOnce(env.DB);
+        if (env.DB) {
+          await ensureSchemaOnce(env.DB);
+          // `authResult.eventId` is always set here (it's `event.id`, populated on a
+          // `valid: true` result after verifyEvent) -- the `&&` is TS narrowing
+          // `eventId?: string` to `string` for consumeNip98Nonce, not a fail-open branch.
+          if (authResult.eventId && !(await consumeNip98Nonce(env.DB, authResult.eventId))) {
```
dcadenas: Guard the `consumeNip98Nonce` failure on this moderation-status GET so a D1 write error does not escape to the worker-wide 500 handler. The current `await` makes the read-only status response unavailable when the nonce INSERT fails, even though the PR's fail-open policy is intended to preserve this endpoint during storage outages. Catch and log the storage error here, continue to `handleGetModerationStatus`, and keep the parent-contact POST fail-closed when nonce persistence fails.

dcadenas (changes requested):
The moderation-status GET now depends on a successful nonce write, so a D1 write outage can turn a read-only status check into a 500. Make the GET's nonce-storage failure behavior explicit and preserve fail-closed behavior for the parent-contact POST before merging.

worker/src/index.ts
mbradley: Addressed. The moderation-status GET now wraps the nonce write (and `ensureSchemaOnce`, which also writes) in a single fail-open try: a thrown storage error is logged and the read-only status check proceeds, while a detected replay still returns 401. The parent-contact POST stays fail-closed (explicit 500 `Replay protection unavailable` on a storage throw), since a state-changing action should not proceed without recorded replay protection. Added tests for both throw paths (GET returns 200, POST returns 500), and merged main to resolve the `ensureSchema` conflict from #203.

dcadenas (approved):
🔒 single-use replay protection is ready to ship

PR #203, opened 2026-07-23
fix(db): self-heal zendesk_preauth_nonces in ensureSchema (fresh-D1 footgun)

mbradley:
## What this does

This worker builds its D1 schema at runtime via `ensureSchema`; the `migrations/*.sql` files are legacy and are not applied to fresh databases. `zendesk_preauth_nonces` was defined only in a migration, so it was never created on a freshly-provisioned D1. The Zendesk pre-auth flow calls `ensureSchema` expecting that table and then inserts into it, so on a fresh D1 the insert throws "no such table" and Zendesk support-auth token issuance fails outright.

This adds `zendesk_preauth_nonces` and its index to `ensureSchema` so a fresh environment works. The DDL matches migration 0003 exactly, and `CREATE TABLE IF NOT EXISTS` makes it a no-op on existing (migrated) databases.

## Scope

Only `zendesk_preauth_nonces`. `zendesk_tickets` is also migration-only, but it is already self-healed by `ensureZendeskTable()` at its usage sites, so it is intentionally not duplicated here (recorded in a code comment).

## Testing

A real-D1 test runs `ensureSchema` on a fresh Miniflare D1 and asserts the table exists, is usable (insert and read back), and has its index, and is mutation-checked (removing the CREATE turns the test red). Full worker suite green, typecheck clean.

## Impact

Latent today (prod and staging were built via migrations), but it strikes any new environment, a D1 recreation, or local-dev-from-scratch.

### DIFF
--- a/worker/src/db.ts
+++ b/worker/src/db.ts
@@ -1,5 +1,6 @@
 // ABOUTME: Shared D1 schema initialization for worker and Durable Objects
-// ABOUTME: Single source of truth for moderation_decisions and moderation_targets DDL
+// ABOUTME: Creates moderation_decisions, moderation_targets, age_review_cases,
+// ABOUTME: age_review_config, and zendesk_preauth_nonces DDL on a fresh D1
 
 /**
  * Ensure all moderation tables and indexes exist.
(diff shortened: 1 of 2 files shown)


dcadenas (approved):
✅ fresh-D1 schema healing is covered, ship it

PR #205, opened 2026-07-25
feat(age-review): view reported and account content in the case pane (#200)

mbradley:
## Summary

Moderators reviewing an age case could not see the content they were deciding on. This surfaces it in the case pane using only the admin-authorized read paths that already exist, and when content cannot be shown it says why, from a verified cause rather than a guess.

- **Reported content**: `report_id` holds the kind-1984 report's own id, not the id of what was reported, so resolution is two hops. Fetch the report, read its target from the tags, then fetch that event, falling back to `getbannedevent` so a banned post still shows (badged "removed (banned)").
- **Account's recent posts**: rendered through the existing `MediaPreview` (click-to-reveal; its media-proxy fallback covers Blossom-blocked blobs). The reported event is deduped out of the list.
- **Honest labelling**: every "not visible" message names a verified cause (hidden by suspension, removed by ban, relay error, confirmed absent, or status unknown). A failed or unfinished read is never presented as absence.

## Motivation

Closes #200. A moderator clearing or denying an age case needs to see the reported content and the account's other posts to make the call. The pane previously showed account and enforcement metadata but none of the content itself.

## What this does and does not make visible

Read this before assuming it solves the suspended-account case. It does not.

**Becomes visible:**
- The reported post when it was **banned**, retrieved with `getbannedevent`.
- **Blossom-blocked or restricted media**, through the existing authenticated media-proxy.
- Ordinary relay-visible content for a target not yet under enforcement.

**Stays invisible:**
- A **suspended** account's posts. The relay exposes `suspendpubkey`, `unsuspendpubkey` and `listsuspendedpubkeys`, but nothing that returns a suspended pubkey's events, so there is no server-side path to that content and no client change can surface it. Tracked in funnelcake#680.

Age-review enforcement suspends at the account level rather than banning individual posts, so for a target who is already suspended this shows the labelled reason rather than content. The viewing benefit lands on cases before suspension, and on cases where the specific reported post was banned or its media blocked. That is an improvement over a blank pane, but it is not "moderators can now see suspended content".

## Two correctness properties this had to get right

**Never state an absence you have not established.** `NPool.query` returns partial results instead of throwing: it swallows errors, breaks on a relay `CLOSED`, and returns immediately when no relay is routed to. Every one of those yields an empty array that is indistinguishable from a genuinely empty account, and funnelcake sends `CLOSED` when its query layer degrades. Reads that feed an absence claim therefore go through `queryStrict`, which drives `req` directly and treats anything short of `EOSE` as an error. Separately, `getbannedevent` answers "not banned" as HTTP 400, which is distinguished from a transport or auth failure so an outage is never reported as a deletion.

**Never attribute content to the wrong account.** NIP-56 does not require the reported event to be authored by the reported pubkey, and case creation does not check it, so a mis-filed report can name a third party's post (a report about a repost legitimately names the original). Such a post is still shown, because hiding it would lose evidence, but it carries an explicit "authored by X, not by this case's subject" warning and every content card shows its author, so attribution is never implied by position alone. The same normalized pubkey drives the reads and the enforcement actions, so the account being judged and the account being acted on cannot diverge.

A third property is worth naming because it shapes the copy: **a claim resting on a status we could not refresh says so.** The account status is data-first, matching the panel rendered directly above this one so the two cannot contradict each other, but when a refetch has failed the message adds that the answer could not be refreshed and offers a retry.

## Architecture

- `src/lib/relayRead.ts`: `queryStrict` (an EOSE-or-error relay read) and `isDefinitiveRpcNegative` (tells the relay's "no" apart from our failure to ask).
- `src/lib/contentVisibility.ts`: pure derivation of the visibility reason. No I/O.
- `src/hooks/useReportedEvent.ts`: two-hop resolution with the banned fallback and the author check. The tagged-target list comes from an untrusted event, so it is deduped and bounded, and the lookup stops when the query is cancelled.
- `src/components/AgeReviewContent.tsx`: presentation only.
- Wired into `AgeReviewDetail.tsx` below `AccountTypeIndicator`.

## Where a reviewer's attention is best spent

The risk in this change is concentrated in two places, both about claims made to a moderator rather than about rendering:

- `useReportedEvent` and `relayRead`, where a read failure must never resolve into a statement about what exists.
- `contentVisibility`, where the account status and the relay read are combined into a single "why not visible" claim.

## Validation

- `npx tsc -p tsconfig.app.json --noEmit`: clean. `npx eslint .`: clean.
- `npx vitest run`: 498 passing, 1 skipped.
- Tests cover the pure derivations, the two-hop lookup and its failure modes, and the wiring that feeds them, so a dropped prop surfaces as a wrong claim on screen rather than as silence.
- Guards are mutation-verified: each was broken deliberately, confirmed to turn a test red, and restored.

## Screenshots

Not attached. The states that most need a look (relay error, hidden-by-suspension, foreign-author) cannot be staged against a live environment: the relay has no verb for reading a suspended account's events, and staging rejects the seeded events needed to reproduce the rest. A live capture would only show the happy path. All states are covered by unit and wiring tests instead.

### DIFF
--- a/src/components/AgeReview.handoff-mutation.test.tsx
+++ b/src/components/AgeReview.handoff-mutation.test.tsx
@@ -15,10 +15,10 @@ const updateAgeReviewCase = vi.fn();
 const getAgeReviewConfig = vi.fn();
 const getAccountStatus = vi.fn();
 
-vi.mock('@/hooks/useReportedEvent', () => ({ useReportedEvent: () => ({ data: null, isLoading: false }) }));
+vi.mock('@/hooks/useReportedEvent', () => ({ useReportedEvent: () => ({ data: undefined, isFetching: false, isError: false, refetch: () => Promise.resolve() }) }));
 vi.mock('@/hooks/useUserStats', () => ({
   useUserStats: () => ({
-    data: { postCount: 0, reportCount: 0, labelCount: 0, recentPosts: [], existingLabels: [], previousReports: [] },
+    data: { postCount: 0, reportCount: 0, labelCount: 0, recentPosts: [], existingLabels: [], previousReports: [], relayIncomplete: false },
     isError: false,
   }),
 }));
--- a/src/components/AgeReview.handoff-mutation.test.tsx
+++ b/src/components/AgeReview.handoff-mutation.test.tsx
@@ -15,6 +15,7 @@ const updateAgeReviewCase = vi.fn();
 const getAgeReviewConfig = vi.fn();
 const getAccountStatus = vi.fn();
 
+vi.mock('@/hooks/useReportedEvent', () => ({ useReportedEvent: () => ({ data: null, isLoading: false }) }));
 vi.mock('@/hooks/useUserStats', () => ({
   useUserStats: () => ({
     data: { postCount: 0, reportCount: 0, labelCount: 0, recentPosts: [], existingLabels: [], previousReports: [] },
--- a/src/components/AgeReview.list-shadow.test.tsx
+++ b/src/components/AgeReview.list-shadow.test.tsx
@@ -15,10 +15,10 @@ const updateAgeReviewCase = vi.fn();
 const getAgeReviewConfig = vi.fn();
 const getAccountStatus = vi.fn();
 
-vi.mock('@/hooks/useReportedEvent', () => ({ useReportedEvent: () => ({ data: null, isLoading: false }) }));
+vi.mock('@/hooks/useReportedEvent', () => ({ useReportedEvent: () => ({ data: undefined, isFetching: false, isError: false, refetch: () => Promise.resolve() }) }));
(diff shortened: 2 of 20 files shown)


dcadenas (approved):
### What was wrong

The hook used one completeness flag for authored content, labels, and prior reports. A failure in an unrelated labels or reports read could therefore make the case pane claim that account content could not be loaded, even when the authored-content read had completed and found no posts.

### What I changed

I changed `useUserStats` to track completeness separately for authored content, labels, and reports, and updated the age-review pane to use only authored-content completeness for content visibility and enforcement decisions. I also added regression coverage for mixed read results. The change is in `fix(age-review): separate content read completeness` (`96f7801`), and CI passed on that commit.

I made these changes and verified them. Please review them and merge the PR if they look good.

PR #207, opened 2026-07-28
revert: nip98 single-use replay nonce for the mobile minor-review endpoints (#202)

mbradley:
## Summary

Reverts #202 (`dff6205`), the single-use NIP-98 replay nonce on the mobile minor-review endpoints. The nonce is incompatible with the mobile client that is live in the App Store today, and deploying it would break the parent-contact flow for minors in review.

This restores `main` to the tree that is currently running in production.

## How this surfaced

This was not caught in review of #202. It came out of local environment testing while I was designing further improvements in this area, and the pre-deploy safety checks independently flagged it before anything reached production. Nothing shipped, so there is no user impact to remediate. Worth noting for our own process: the incompatibility is only visible when you look at the worker and the mobile client together, which is not what a single-repo review does by default.

## Why it breaks

The nonce rejects any repeat presentation of a signed NIP-98 event. The shipped mobile client (1.0.17) caches signed tokens for 45 seconds and reuses them for identical requests. The worker allows 60 seconds of clock skew, so a cached token is still time-valid and gets rejected as a replay rather than as stale. The windows overlap completely, so there is no safe gap: any repeat status read within 45 seconds returns 401.

The worst case is right after a successful parent-contact submission. The POST succeeds, the app immediately refetches status using the cached token, and that refetch 401s. The client only degrades gracefully on 404 and 501, so a submission that actually worked surfaces to the parent as an error. Retrying reuses the same cached token, so retry cannot succeed until the cache expires.

This is production-only. The public host allowlist is empty on staging, which is why staging was safe to exercise it on.

## Current state

Production runs `ab1b913`, which is this payload without #202. This revert makes `main` identical to that tree, so `main` is deployable again. That is the main reason to land it rather than leave the tip unshipped: otherwise every future merge stacks on a tip that must not be deployed.

## What comes next

#195 should be reopened. The replacement needs to reject *tampered* replays rather than *repeated* ones, so that it coexists with client-side token caching and does not require an app release to be safe. I will follow up with that separately, along with a note in the repo capturing the client-caching constraint so this is not reintroduced.

## Validation

- Reverted tree verified byte-identical to `ab1b913`, the commit running in production.
- Worker tests: 367 passing. Worker D1 tests: 21 passing (these are excluded from the default worker run and have to be invoked explicitly).
- `tsc -p tsconfig.app.json` clean, eslint clean.
- No frontend files touched.


dcadenas (approved):
✅ mobile NIP-98 auth is back to the intended flow

PR #208, opened 2026-07-28
docs: flag the worker D1/e2e test exclusion as a false green

mbradley:
## Summary

AGENTS.md told contributors to verify worker changes with `cd worker && npx vitest run` without saying what that command skips. It excludes `test/**/*.d1.test.ts` and `test/**/*.e2e.test.ts`, so a change whose tests all live in one of those files reports success having run none of them.

This documents the gap in the three places someone would actually look: the command list, the copy-paste bash block, and the "Always do first" checklist.

## Motivation

This is the same class of trap as the bare `tsc --noEmit` false green already documented in this file, and it is worth writing down for the same reason: the documented command exits 0 while verifying nothing.

It is not hypothetical. It came up while reviewing a recent worker change whose entire new test suite lived in excluded files, so the documented worker test command was green throughout.

## What this does not change

The exclusion itself is deliberate and stays as is. `worker/vitest.config.ts` says so directly: it keeps the default suite fast and free of a Miniflare/workerd or local-relay dependency, which is the right default for CI and contributors. The problem was the documentation, not the config.

A stronger fix would be to run `test:d1` in CI on worker-touching PRs rather than relying on contributors reading the docs. That is a separate change and a separate decision, so it is not in scope here.

## Validation

Both newly documented commands were run and behave as described:

- `npm run test:d1`: 5 files, 25 tests, none of which the default run executes.
- `npm run test:e2e`: fails on the WebSocket connection without a local relay, confirming the stated dependency.

Docs only. No code, config, or workflow changes.

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -85,7 +85,9 @@ worker/
 - `npx tsc -p tsconfig.app.json --noEmit`: frontend type-check. **Bare `npx tsc --noEmit` is a false green here** — the root `tsconfig.json` has `"files": []` with project references, so in non-build mode it type-checks nothing and exits 0. Always pass `-p tsconfig.app.json` (or run `npm run test`).
 - `npx vite build`: frontend production build.
 - `npx vite --port 8080`: frontend dev server.
-- `cd worker && npx vitest run`: worker tests when touching worker code.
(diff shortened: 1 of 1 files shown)


dcadenas (approved):
### What was wrong

The checklist only named schema and route changes, so a contributor changing scheduled D1 behavior could run the default worker suite and miss real-D1 tests that Vitest excludes. That left the false-green path this PR is meant to prevent.

### What I changed

I broadened the checklist to require `cd worker && npm run test:d1` for any worker behavior that reads or writes D1, while keeping schema, migrations, and D1-backed routes as examples. I made this in `docs: require D1 tests for all D1-backed worker changes` (`54f0a6a`), and CI passed on that commit.

I made these changes and verified them. Please review them and merge the PR if they look good.

PR #209, opened 2026-07-28
fix(auto-hide): match trusted client names case-insensitively

mbradley:
## What

The client name drifted: `divine-mobile` sends `Divine` while `TRUSTED_CLIENTS` carries the stylized `diVine`. The comparison was exact, so mobile reports stopped matching, and we missed the opportunity to make this case-insensitive when the gate went in.

Per NIP-89 that tag element is a display name rather than a stable identifier, so its casing is each app's choice and shouldn't decide enforcement. Both sides are now trimmed and lowercased.

Preventative fix for interim stability: affected reports were still actioned through other paths, so this restores the fast path rather than fixing an outage.

## Tests

Added coverage for casing variants being accepted (`Divine`, `DIVINE`, ` divine `, `Divine-Web`) and for genuinely untrusted clients still being rejected. Confirmed the new test fails without the change.

## Notes for the reviewer

- Merging doesn't ship this; relay-manager deploys by manual `wrangler`.
- `npm ci` in `worker/` needs `--legacy-peer-deps`, and 8 tests in `age-review.test.ts` / `bulk-moderate.test.ts` fail on `main` independently of this change. Both pre-existing.
- The `client` tag is self-declared and unverifiable: a provenance hint, not a trust boundary. Worth revisiting what this tier should actually gate on.
- Longer term the stable identifier is the NIP-89 handler address (element 3 of the tag, `31990:<pubkey>:<d>`). Matching on that would need a config change, so it's out of scope here.

### DIFF
--- a/worker/src/ReportWatcher.test.ts
+++ b/worker/src/ReportWatcher.test.ts
@@ -802,6 +802,61 @@ describe('ReportWatcher', () => {
       expect(mockDbRun).toHaveBeenCalled();
     });
 
+    // divine-mobile sends `Divine`; TRUSTED_CLIENTS carries the stylized `diVine`.
+    it('should accept a trusted client whose casing differs (divine-mobile sends "Divine")', async () => {
+      for (const clientName of ['Divine', 'DIVINE', ' divine ', 'Divine-Web']) {
+        mockFetch.mockClear();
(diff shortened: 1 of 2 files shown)


dcadenas (approved):
✅ trusted client names now match case-insensitively

PR #217, opened 2026-07-30
fix(age-review): refuse an un-ban on an account under age review

mbradley:
## What

An un-ban of a pubkey with an open age-review case is now refused, the same way an un-suspend already was, and the moderator is routed to the case instead of a dead-end error.

Closes #224.

## Why

`unbanpubkey` calls `unsuspendUser`, which sets Keycast status to `active`. That lifts a *suspension* as well as a ban. So un-banning an account under age review restored login and signing while skipping the case entirely, which is what the guard exists to prevent.

The guard's own comment stated the assumption that made this safe: *"internal moderation/bulk callers use ban*/unban* only, so neither reaches this guard."* That is no longer true. Two internal callers already reach it (the admin UI's un-ban, and `allow_pubkey` via `/api/moderate`, which re-enters `handleRelayRpc`), and two services outside this repo reach it: `divine-moderation-service`, which has mapped `allow_pubkey` -> `unbanpubkey` onto `/api/relay-rpc` since 2026-06-08, and the Coop enforcement adapter, whose reversal routes followed on 2026-06-17.

## What changes for a moderator

Four things they will notice:

1. **Un-ban and Allow are refused on an account under age review**, with a redirect to the case rather than a red error toast. Resolve the case and the un-ban proceeds normally.
2. **If the age-review check itself cannot run, a reversal is refused** rather than going through unchecked. Previously a database problem meant the check silently did not happen and the un-ban succeeded, with nobody learning the hold had been lifted without review. The cost is accepted: during a real outage, reversals are blocked. An outage long enough to matter needs fixing rather than papering over, and `banpubkey` stays unguarded so severe enforcement still works throughout.
3. **`allow_pubkey` via `/api/moderate` now returns the refusal it was given** instead of flattening everything to a 500, so callers can tell a permanent refusal from a transient failure. Only `allow_pubkey`: `delete_event` and `ban_pubkey` still flatten, because neither is guarded and so neither can produce a code a caller would need.
4. **A malformed pubkey is rejected outright on the ENFORCE direction.** `banpubkey` and `suspendpubkey` now answer a non-canonical `params[0]` with `400 Invalid pubkey` before the guard runs, matching what `handleGetActiveAgeReviewCase` already does for the same input. Previously it was forwarded: the relay matches these bytes exactly, so it enforced on nobody while reporting success, and the guard's lookup would have missed a real open case for the same reason. The two REVERSALS deliberately still accept one, so a row banned with a non-canonical value stays removable -- pinned by "lets the REVERSE direction carry a non-canonical pubkey, so a bad row stays removable". The Add User dialog now canonicalises what you paste, so an uppercase pubkey works rather than silently doing nothing.

## Deliberately unchanged

- **`banpubkey` is not guarded.** It is the severe-action escape hatch, pinned by an existing test: a moderator who finds something severe must be able to act without resolving the case first. The reason the asymmetry is coherent is mechanical rather than a matter of taste, and is now written next to the code: `banpubkey` performs a destructive content purge that `unbanpubkey` cannot reverse, so guarding the un-ban removes no recovery path that ever existed for content. It defers only restoring login and signing, which is the hold the case owns.
- **`suspendpubkey` still fails open** when the check cannot run. Failing open there over-enforces, which is visible and undone by the moderator who did it, and blocking it would stop moderation during an outage.
- **Bulk moderation still fails open.** Its own test states that as a deliberate choice. Fail-closed is opt-in per call site rather than a change to the shared guard, so this decision did not silently change a path it was not about.

## Notes for review

The redirect logic had been hand-copied to four call sites across **two** components (`UserActions` suspend/unsuspend/bulk, and `UserManagement` unsuspend as its own inline copy). Extending the guard to `unbanpubkey` gives **four new** sites the ability to receive this code, so rather than copy it a fifth through eighth time it now lives in one hook, `useAgeReviewGuardRedirect`, tested directly. (An earlier version of this note said three components and a missed fifth site; `EventDetail` had no guard handling at all, and the `allow_user` dead-end it cited was not reachable on `main`, since the guard there covered only suspend/unsuspend.)

Two call sites are covered only through that shared hook rather than at component level: `UserManagement`'s un-suspend, whose routing predates this change, and `EventDetail`'s un-ban, which has no test file and pulls in several unrelated hooks. Stated as a conscious drop rather than left implicit. Deleting `EventDetail`'s call alone still survives the suite.

## Test plan

- [x] `tsc -p tsconfig.app.json --noEmit` clean
- [x] `eslint` clean
- [x] UI suite: 514 passing, 1 skipped
- [x] Worker suite: 385 passing, plus 21 D1 tests (`npm run test:d1`)
- [x] Mutation-checked in both directions: removing the guard reds the refusal tests; extending it to `suspendpubkey` reds the fail-open test; neutering the missing-DB refusal reds; dropping the `code` argument from `callRelayRpc` reds the parsing test; reverting the schema-bootstrap catch reds the 3 cold-isolate tests, which are the **only** detectors among 384; deleting the guard-ordering assertion lets a reordered guard survive the entire suite; reverting the pubkey canonicalisation reds the UI test
- [x] Order-independence: `worker/src/index.test.ts` is 59/59 across shuffled seeds, previously 57/1
- [x] **Both refusal paths exercised end to end against a local stack**, with the real Coop enforcement adapter in the loop (recipe: `support-trust-safety/docs/moderation/local-validation-harness.md`)
- [x] `cd worker && npx vitest run src/index.test.ts` - 59 passing
- [x] Takeover validation: `git diff --check`
- [ ] Staging validation

### What the local run showed

The adapter is the consumer that matters here, since it is what makes `unbanpubkey` reachable from Coop in the first place. Both new responses behave correctly downstream:

| This PR returns | Adapter does | Coop sees |
| --- | --- | --- |
| 409 `age_review_active` | passes it through, alert **names the case id and state** and says to clear it from the Age Review flow | terminal |
| 503 `age_review_check_failed` | **502** | terminal, silently (see below) |

**The 503 is not retried, and the row above is why that matters.** The adapter maps it to 502, but Coop drops it there: `publishAction` throws on a non-2xx, and that throw is caught by `publishAction`'s own catch, which logs and returns `false`. `withRetries` wraps `publishAction`, so it only ever sees a resolved value and never re-runs -- the five-attempt budget never engages (coop 1.0.2, `ActionPublisher.ts`: throw `:537`, catch `:731`, return false `:750`).

So the accepted cost, stated accurately: during a D1 outage reversals are refused, and those refusals are dropped silently rather than deferred until the database recovers. The safety direction is unchanged -- a dropped un-ban leaves the hold on -- and `banpubkey` stays unguarded so severe enforcement works throughout.

**One edge worth knowing rather than discovering.** The adapter does not alert on 5xx, which was justified on the assumption that something retries. Since nothing does, a fail-closed reversal during an outage is dropped with no alert at all -- immediately, not after a budget runs out. Tracked in divinevideo/coop-webhook-adapter#6, refiled from #223 into the repo that owns `reversalAlert`, which is where the decision on whether to alert belongs.

**Correction, twice over.** I first recorded 8 worker tests as failing on `main`, then corrected that to a Node version difference. Both were wrong. `worker/` has its own package and lockfile, and in a fresh worktree `worker/node_modules` is empty, so the suite resolves against the root `node_modules` and 8 tests fail spuriously. After `npm ci --legacy-peer-deps` inside `worker/` -- what CI does, for a known peer conflict -- **all 384 worker tests pass on Node 26.5.1**, along with 21 D1 tests. Nothing was ever wrong with `main` or this branch, but neither explanation of why was right.

## Since review

@NotThatKindOfDrLiz took the review blocker and pushed `398a484`, bootstrapping the D1 schema before the relay-rpc guard. Kept: the fresh-D1 gap was real and its regression test is load-bearing.

@realmeylisdev then found that the bootstrap escaped as an unhandled rejection, because this route returns its promise without awaiting, so the top-level catch never saw it. That cost `suspendpubkey` the fail-open behaviour it has on `main`, during exactly the outage it exists for. Their trace is what made it findable, and their fix is what landed. They also found the outage tests were order-dependent: the failing-D1 mock never failed the DDL, and a module-level `schemaReady` meant whichever test reached it first made the bootstrap a no-op for the rest of the file.

Four follow-up commits, all with tests mutation-checked before being kept:

| | |
| --- | --- |
| `99e216f` | A DDL failure reaches the guard's fail-open/fail-closed decision instead of escaping. Cold-isolate tests via `vi.resetModules()` so a freshly started worker is actually exercised. Pins that a refusal lands *before* the relay call -- the previous tests asserted only Keycast, which is `waitUntil` work skipped on any early return, so a guard moved after `callNip86Rpc` reddened nothing. |
| `301130b` | Hook docblock corrected. |
| `e190436` | Pubkeys canonicalised at the Add User dialog, which validates case-insensitively and passed the raw value through. Also pins guard ordering on suspend/unsuspend, and moves the last order-dependent test onto the cold-isolate helper. |
| `73efec2` | A malformed pubkey answered with `400` at the handler rather than the guard's retryable `503`. |
| `53e2279` | Pre-push review of my own diff: dedupe the target expression. |
| `380b733` | Pins the 503 half of the `/api/moderate` contract, which had no detector: swapping `!rpcResponse.ok` for `status === 409` passed all 384 tests while stripping `code` from the 503. Plus two comments that described behaviour the code does not have. |
| `4ce69d8` | Canonical hex required to ENFORCE, not to REVERSE. Validating the reversals but not `banpubkey` made cleanup stricter than the thing that creates the mess: an uppercase ban was accepted and then could not be lifted, leaving a row stuck in the ban list with no way out of the UI. |

## Follow-ups, not in scope here

- divinevideo/coop-webhook-adapter#6: Coop does not retry a 5xx at all, so a fail-closed reversal is dropped silently. Whether the adapter should alert on 5xx is the open decision there.
- divinevideo/divine-moderation-service#191: that service reaches `unbanpubkey` too and throws away `code`/`caseId`/`state`, so a moderator un-banning from its dashboard gets a generic failure with no case to open. A consequence of landing this, not a defect in it.
- The age-review clear path calls `unsuspendPubkey` rather than `unbanPubkey`, so an account banned through the escape hatch and later cleared needs a second manual step.
- `UserActions` closes over the `pubkey` prop rather than mutation variables, so switching selected user mid-flight redirects to the *newly selected* user's case, not merely a stale one: the request went out against the old pubkey, but React Query invokes `onError` from the current render's options. Not a one-line fix either, since those mutationFns take no variables and close over the same prop. Pre-existing, and the refactor confines it to `UserActions` rather than spreading it, since all four new sites pass `variables.pubkey`.

### DIFF
--- a/src/hooks/useAgeReviewGuardRedirect.ts
+++ b/src/hooks/useAgeReviewGuardRedirect.ts
@@ -1,0 +1,42 @@
+// ABOUTME: Routes a moderator to the Age Review case when the worker's age-review
+// guard refuses an enforcement action, instead of surfacing a raw error.
+
+import { useCallback } from 'react';
+import { useNavigate } from 'react-router-dom';
+import { ApiError } from '@/lib/adminApi';
+import { useToast } from '@/hooks/useToast';
+
+/**
+ * The worker refuses `suspendpubkey`, `unsuspendpubkey` and `unbanpubkey` on a
+ * pubkey with an open age-review case, answering with a 409 carrying
+ * `code: 'age_review_active'`. Enforcement must not drift from the case out of
+ * band, so a refused action sends the moderator to the case rather than showing
+ * a dead-end error toast.
+ *
+ * `banpubkey` is deliberately not guarded and so never produces this code: it is
+ * the severe-action escape hatch.
+ *
+ * Returns a predicate. Call it first in a mutation's `onError`; if it returns
+ * true it has already handled the error and the caller should return.
+ *
+ * This is shared rather than copied because it is needed at every call site of a
+ * guarded RPC, and a copied block is easy to omit on a new one. It was, on the
+ * `allow_user` path, which reached `unbanpubkey` and dead-ended.
+ */
+export function useAgeReviewGuardRedirect() {
+  const navigate = useNavigate();
+  const { toast } = useToast();
+
+  return useCallback(
+    (error: unknown, pubkey: string): boolean => {
+      if (!(error instanceof ApiError) || error.code !== 'age_review_active') return false;
+      toast({
+        title: 'This account is under age review',
+        description: 'Opening it in the Age Review flow.',
+      });
+      navigate(`/age-review?pubkey=${encodeURIComponent(pubkey)}`);
+      return true;
+    },
+    [navigate, toast],
+  );
+}
--- a/src/hooks/useAgeReviewGuardRedirect.ts
+++ b/src/hooks/useAgeReviewGuardRedirect.ts
@@ -22,13 +22,21 @@ import { useToast } from '@/hooks/useToast';
  * Shared rather than copied because extending the guard to `unbanpubkey` makes
  * four new call sites able to receive this code — `UserActions` un-ban,
  * `UserManagement` allow_user and un-ban, and `EventDetail` un-ban — on top of
- * the two that already handled it. Six near-identical copies is how one gets
- * missed, so the handling lives here and each site calls it.
+ * the four that already handled it (`UserActions` suspend, unsuspend and bulk;
+ * `UserManagement` unsuspend, which was its own inline copy). Eight
+ * near-identical copies is how one gets missed, so the handling lives here.
  *
- * Not every caller of a guarded RPC needs it: `DebugPanel` issues arbitrary RPC
- * methods and records the error into an action log rather than surfacing a
- * moderator-facing toast, so it deliberately does not use this. The rule is
- * whether a moderator is waiting on the outcome, not whether the RPC is guarded.
+ * Only covers the 409. The same three RPCs also answer 503
+ * `age_review_check_failed` when the check cannot run at all, which is not a
+ * routable case — there is no case id to route to — so this returns false and
+ * each site shows its own error toast. That message is already actionable.
+ *
+ * `DebugPanel` deliberately does not use this. It issues arbitrary RPC methods
+ * as a raw diagnostic and reports whatever came back, rather than running a
+ * moderator workflow with somewhere to be sent. The test is whether a moderator
+ * is waiting on the outcome, not whether the RPC is guarded. (It does toast, and
+ * currently reports a guard refusal as "Partial success", which is misleading
+ * but is a DebugPanel problem rather than one this hook should solve.)
  */
 export function useAgeReviewGuardRedirect() {
   const navigate = useNavigate();
--- a/src/hooks/useAgeReviewGuardRedirect.ts
+++ b/src/hooks/useAgeReviewGuardRedirect.ts
@@ -26,10 +26,12 @@ import { useToast } from '@/hooks/useToast';
  * `UserManagement` unsuspend, which was its own inline copy). Eight
  * near-identical copies is how one gets missed, so the handling lives here.
  *
- * Only covers the 409. The same three RPCs also answer 503
- * `age_review_check_failed` when the check cannot run at all, which is not a
- * routable case — there is no case id to route to — so this returns false and
- * each site shows its own error toast. That message is already actionable.
+ * Only covers the 409. The two REVERSALS, `unsuspendpubkey` and `unbanpubkey`,
+ * also answer 503 `age_review_check_failed` when the check cannot run at all.
+ * `suspendpubkey` never does: fail-closed is passed only for reversals, so a
+ * suspend whose lookup fails proceeds and enforces. The 503 is not a routable
+ * case either way, since there is no case id to route to, so this returns false
+ * and each site shows its own error toast. That message is already actionable.
  *
  * `DebugPanel` deliberately does not use this. It issues arbitrary RPC methods
  * as a raw diagnostic and reports whatever came back, rather than running a
--- a/src/hooks/useAgeReviewGuardRedirect.ts
+++ b/src/hooks/useAgeReviewGuardRedirect.ts
@@ -19,9 +19,16 @@ import { useToast } from '@/hooks/useToast';
  * Returns a predicate. Call it first in a mutation's `onError`; if it returns
  * true it has already handled the error and the caller should return.
  *
- * This is shared rather than copied because it is needed at every call site of a
- * guarded RPC, and a copied block is easy to omit on a new one. It was, on the
- * `allow_user` path, which reached `unbanpubkey` and dead-ended.
+ * Shared rather than copied because extending the guard to `unbanpubkey` makes
+ * four new call sites able to receive this code — `UserActions` un-ban,
+ * `UserManagement` allow_user and un-ban, and `EventDetail` un-ban — on top of
+ * the two that already handled it. Six near-identical copies is how one gets
+ * missed, so the handling lives here and each site calls it.
+ *
+ * Not every caller of a guarded RPC needs it: `DebugPanel` issues arbitrary RPC
+ * methods and records the error into an action log rather than surfacing a
+ * moderator-facing toast, so it deliberately does not use this. The rule is
+ * whether a moderator is waiting on the outcome, not whether the RPC is guarded.
  */
 export function useAgeReviewGuardRedirect() {
   const navigate = useNavigate();
--- a/worker/src/index.test.ts
+++ b/worker/src/index.test.ts
@@ -547,6 +547,39 @@ describe('relay-rpc account-state side effects', () => {
     fetchSpy.mockRestore();
   });
 
+  it('allow_pubkey via /api/moderate forwards the 503 too, not just the 409', async () => {
+    // The other half of the same contract, and the half a caller most needs: a
+    // permanent refusal and an unrunnable check must stay distinguishable across
+    // the hop. Flattening this one to 500 strips `code`, so the caller cannot tell
+    // "there is a case" from "we could not look" from "the relay broke".
+    const fetchSpy = makeFetchSpy();
+    const waitUntil = vi.fn();
+    const testCtx = { waitUntil } as unknown as ExecutionContext;
+
+    const response = await worker.fetch(
+      new Request('https://api-relay-prod.divine.video/api/moderate', {
+        method: 'POST',
+        headers: {
+          'Content-Type': 'application/json',
+          'X-Admin-Key': 'test-admin-key',
+          Origin: 'https://app.divine.video',
+        },
+        body: JSON.stringify({ action: 'allow_pubkey', pubkey: VALID_PUBKEY }),
+      }),
+      makeAccountStateEnvWithFailingCaseLookup(),
+      testCtx,
+    );
+
+    expect(response.status).toBe(503);
+    const body = await response.json() as { code: string };
+    expect(body.code).toBe('age_review_check_failed');
+
+    await drain(waitUntil);
+    expect(keycastCalls(fetchSpy)).toHaveLength(0);
+
+    fetchSpy.mockRestore();
+  });
+
   it('refuses a reversal when the case lookup itself fails, rather than lifting the hold', async () => {
     // Fail closed on the reversal direction: an unchecked unban silently lifts a
     // minor-safety hold and reports success, so nobody learns the check never
--- a/worker/src/index.test.ts
+++ b/worker/src/index.test.ts
@@ -412,6 +412,41 @@ describe('relay-rpc account-state side effects', () => {
     fetchSpy.mockRestore();
   });
 
+  it('allow_pubkey via /api/moderate forwards the guard 409 instead of flattening it to 500', async () => {
+    // allow_pubkey re-enters handleRelayRpc with unbanpubkey, so it inherits the
+    // guard. Re-wrapping at 500 would drop code/caseId/state that callers route
+    // on, and would label a permanent refusal as transient, which retrying
+    // clients treat as "try again".
+    const fetchSpy = makeFetchSpy();
+    const waitUntil = vi.fn();
+    const testCtx = { waitUntil } as unknown as ExecutionContext;
+    const env = makeAccountStateEnvWithDb({ id: 'case-allow', state: 'restricted_pending_user_response' });
+
+    const response = await worker.fetch(
+      new Request('https://api-relay-prod.divine.video/api/moderate', {
+        method: 'POST',
+        headers: {
+          'Content-Type': 'application/json',
+          'X-Admin-Key': 'test-admin-key',
+          Origin: 'https://app.divine.video',
+        },
+        body: JSON.stringify({ action: 'allow_pubkey', pubkey: VALID_PUBKEY }),
+      }),
+      env,
+      testCtx,
+    );
+
+    expect(response.status).toBe(409);
+    const body = await response.json() as { code: string; caseId: string };
+    expect(body.code).toBe('age_review_active');
+    expect(body.caseId).toBe('case-allow');
+
+    await drain(waitUntil);
+    expect(keycastCalls(fetchSpy)).toHaveLength(0);
+
+    fetchSpy.mockRestore();
+  });
+
   it('suspendpubkey is refused when the target has an active age-review case', async () => {
     const fetchSpy = makeFetchSpy();
     const waitUntil = vi.fn();
--- a/worker/src/index.test.ts
+++ b/worker/src/index.test.ts
@@ -684,13 +684,13 @@ describe('relay-rpc account-state side effects', () => {
     fetchSpy.mockRestore();
   });
 
-  it('refuses a reversal whose pubkey is not canonical hex, rather than skipping the check', async () => {
+  it('rejects a non-canonical pubkey with a 400, before the guard runs', async () => {
     // The guard's lookup is byte-exact, so a non-canonical pubkey cannot match a
     // real case even when one exists -- it would report "no case" for an account
-    // that may well be under review. That is the same "the check cannot happen"
-    // the missing-binding and failed-lookup paths already refuse on, so it
-    // refuses identically instead of relying on a downstream format check in
-    // another module to stop the mutation.
+    // that may well be under review. Answer it here rather than leaving it to
+    // the guard: this is the same 400 handleGetActiveAgeReviewCase already gives
+    // for the same regex, and a malformed pubkey never becomes valid on a retry,
+    // so the retryable 5xx class is the wrong answer for it.
     const fetchSpy = makeFetchSpy();
     const waitUntil = vi.fn();
     const testCtx = { waitUntil } as unknown as ExecutionContext;
@@ -698,17 +698,19 @@ describe('relay-rpc account-state side effects', () => {
     const response = await callRelayRpc(
       'unbanpubkey', [VALID_PUBKEY.toUpperCase()], makeAccountStateEnvWithDb(null), testCtx,
     );
-    expect(response.status).toBe(503);
-    const body = await response.json() as { code: string };
-    expect(body.code).toBe('age_review_check_failed');
+    expect(response.status).toBe(400);
+    const body = await response.json() as { error: string };
+    expect(body.error).toBe('Invalid pubkey');
 
     await drain(waitUntil);
-    expect(keycastCalls(fetchSpy)).toHaveLength(0);
+    expect(fetchSpy).not.toHaveBeenCalled();
 
     fetchSpy.mockRestore();
   });
 
-  it('still lets a suspend through with a non-canonical pubkey (enforce direction stays open)', async () => {
+  it('rejects a non-canonical pubkey on the enforce direction too', async () => {
+    // Not a fail-open/fail-closed question: the value is unusable for suspend as
+    // well, since the relay stores these bytes exactly.
     const fetchSpy = makeFetchSpy();
     const waitUntil = vi.fn();
     const testCtx = { waitUntil } as unknown as ExecutionContext;
@@ -716,7 +718,10 @@ describe('relay-rpc account-state side effects', () => {
     const response = await callRelayRpc(
       'suspendpubkey', [VALID_PUBKEY.toUpperCase(), 'policy'], makeAccountStateEnvWithDb(null), testCtx,
     );
-    expect(response.status).toBe(200);
+    expect(response.status).toBe(400);
+
+    await drain(waitUntil);
+    expect(fetchSpy).not.toHaveBeenCalled();
 
     fetchSpy.mockRestore();
   });
--- a/worker/src/index.test.ts
+++ b/worker/src/index.test.ts
@@ -248,6 +248,22 @@ describe('relay-rpc account-state side effects', () => {
     } as never;
   }
 
+  // D1 present but throwing, to exercise the guard's fail-open / fail-closed split.
+  function makeAccountStateEnvWithFailingCaseLookup() {
+    return {
+      ...(makeAccountStateEnvWithDb(null) as unknown as Record<string, unknown>),
+      DB: {
+        prepare: () => ({
+          bind: () => ({
+            first: async () => {
+              throw new Error('D1 unavailable');
+            },
+          }),
+        }),
+      },
+    } as never;
+  }
+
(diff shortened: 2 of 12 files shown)


NotThatKindOfDrLiz (approved):
Final pass is clean after the takeover fix. The fail-closed relay-rpc guard now bootstraps the on-demand D1 schema before querying age_review_cases, and the fresh-D1 regression is covered. CI is green on head 398a484.

hm21 (approved):
LGTM

worker/src/index.ts
```diff
@@ -1005,16 +1023,46 @@ async function handleRelayRpc(
...
+  // terminal state getActiveAgeReviewCase returns null and the unban proceeds.
+  //
+  // Age-review's own enforcement calls the nip86 helpers directly, so it does not
+  // reach this guard and can still act on its own cases.
+  if (
+    body.method === 'suspendpubkey' ||
+    body.method === 'unsuspendpubkey' ||
+    body.method === 'unbanpubkey'
+  ) {
+    if (env.DB) await ensureSchemaOnce(env.DB);
```
realmeylisdev: `ensureSchemaOnce` throws when D1 is unavailable, and nothing catches it here. The route is `return handleRelayRpc(...)` at `index.ts:464` with no `await`, so the rejection sails past the `catch` at `:631` and out of `fetch`. The bulk path at `:547` avoids this only because it awaits inline.

Cold isolate against a failing D1 binding:

- `suspendpubkey` → unhandled rejection. On `main` the same request returns 200 and reaches the relay, so the documented fail-open behaviour is gone during the outage it was written for.
- `unbanpubkey` / `unsuspendpubkey` → unhandled rejection rather than the 503 `age_review_check_failed`, with no CORS headers, so the UI redirect cannot run.
- `/api/moderate allow_pubkey` → `500 {"success":false,"error":"D1_ERROR: ..."}`, no `code`, and the raw D1 message reaches an external caller.

Bootstrapping is part of the check, not a precondition for it, so its failure should reach the same fail-open/fail-closed decision the lookup already makes instead of bypassing it. Swallowing it does that — the lookup below then fails the same way and the guard returns the right answer:

```suggestion
    if (env.DB) {
      // Bootstrapping is part of the check, not a precondition for it: a DDL
      // failure must reach the same fail-open/fail-closed decision as a failed
      // lookup rather than escape as an unhandled rejection (the route returns
      // this promise without awaiting it, so the top-level catch never sees it).
      // Swallow it and let the guard decide — the lookup below fails the same
      // way, refusing a reversal and proceeding for a suspend.
      try {
        await ensureSchemaOnce(env.DB);
      } catch (err) {
        console.error('[handleRelayRpc] age-review schema bootstrap failed:', err);
      }
    }
```

I ran this: suspend returns 200 with the relay called and CORS intact, both reversals return 503 `age_review_check_failed` with the relay untouched, `/api/moderate allow_pubkey` forwards that same 503, and the fresh-D1 bootstrap this commit added still works. Full worker suite stays green.

worker/src/index.test.ts
```diff
@@ -390,6 +443,155 @@ describe('relay-rpc account-state side effects', () => {
...
+      'unsuspendpubkey', [VALID_PUBKEY], makeAccountStateEnvWithFailingCaseLookup(), testCtx,
+    );
+    expect(response.status).toBe(503);
+    await drain(waitUntil);
+    expect(keycastCalls(fetchSpy)).toHaveLength(0);
+
+    fetchSpy.mockRestore();
+  });
+
+  it('still lets a suspend through when the lookup fails (over-enforcing is the safe side)', async () => {
```
realmeylisdev, line 655: This one passes only because `schemaReady` in `index.ts` is module-level and an earlier test in this file already flipped it, so `ensureSchemaOnce` is a no-op by the time it runs. Alone it fails:

```
cd worker && npx vitest run src/index.test.ts -t "still lets a suspend through when the lookup fails"

TypeError: db.prepare(...).run is not a function
 ❯ ensureSchema src/db.ts:22:6
 ❯ ensureSchemaOnce src/index.ts:42:9
 ❯ handleRelayRpc src/index.ts:1057:23
```

`makeAccountStateEnvWithFailingCaseLookup` only makes `.first()` throw and defines no `.run()`, but a real D1 outage fails the DDL too — which is the gap behind the other comment.

The same shared flag makes the new bootstrap test at `:387` order-dependent: it exercises the bootstrap only while it happens to be the first test in the file to reach `ensureSchemaOnce`. A `vi.resetModules()` plus dynamic re-import in the outage tests, and a `.run()` that rejects on the failing mock, would pin both.

realmeylisdev (changes requested):
Reviewed the guard end to end, plus the downstream consumers that reach it (the enforcement adapter, the moderation service, the relay, and Keycast) and NIP-86.

The design is sound and the ban/un-ban asymmetry argument holds up. Two things before merge.

## 1. The schema bootstrap is unguarded

`worker/src/index.ts:1057`, from `398a484`. `ensureSchemaOnce` runs ahead of the guard with no error handling, and the route returns `handleRelayRpc(...)` without `await` (`index.ts:464`), so a D1 failure escapes the top-level catch at `:631` entirely.

Cold isolate, failing D1 binding:

| | `main` | this branch |
| --- | --- | --- |
| `suspendpubkey` | 200, relay called | unhandled rejection |
| `unbanpubkey` | 200 (the bug being fixed) | unhandled rejection, not 503 |
| `/api/moderate` `allow_pubkey` | 500 | 500, no `code`, raw D1 error in the body |

So the fail-open guarantee for suspend is lost in exactly the outage it was written for, and a reversal returns neither `age_review_check_failed` nor CORS headers, which means `useAgeReviewGuardRedirect` never fires and Coop sees a platform error instead of the documented body. One-line fix suggested inline; I ran it against all five cases.

## 2. Coop does not retry the 503

`publishAction` throws on `!response.ok` (`coop/server/rule_engine/ActionPublisher.ts:390`), but that throw sits inside the `try` at `:333`, whose `catch` at `:576` returns `false`. `withRetries` (`server/utils/misc.ts:320`) only re-runs on rejection, so the five-attempt budget never engages. Combined with the adapter suppressing alerts on 5xx, a fail-closed reversal is dropped immediately and silently — not "retried until the database recovers", and not bounded by an exhausted budget.

The safety direction is still right: a dropped un-ban leaves the hold on. But the accepted cost is larger than the description and #223 state, and the adapter comment citing #217 rests on the same premise. I traced only that one wrapper, so there may be an outer retry I did not see.

## Confirmed as described

- `unbanpubkey` → `unsuspendUser` → Keycast `active` (`index.ts:1094`, `keycast/api/src/api/http/nostr_rpc.rs:362`); signing is restored. `verified_minor` is a separate column and survives, so the blast radius is the suspension only.
- `banpubkey` really does purge — confirmed in the relay's management handler, which hard-deletes the pubkey's events after recording the ban. So guarding the un-ban removes no recovery path that existed for content.
- Clear is unaffected — age review calls the `nip86` helpers directly (`age-review.ts:440`).
- Bulk still fails open (`index.ts:566` passes no opts).
- NIP-86 defines `banpubkey`/`unbanpubkey`; `suspendpubkey`/`unsuspendpubkey` are Funnelcake extensions, which the spec permits. The spec names only 401, so 409 and 503 conflict with nothing.
- Hook, redirect, and `code` parsing (`adminApi.ts:270-283`) line up. `variables.pubkey` is correct at both new call sites.

## Ran

`tsc -p tsconfig.app.json --noEmit` and `eslint` clean. UI 511 passed, 1 skipped. Worker 377 passed. `npm run test:d1` 21 passed — needed `npm i -D miniflare` first, it is not in `worker/package.json`. Plus the cold-isolate D1-outage repro on both branches.

Happy to push the inline fix myself if you would rather not hand-apply it.

worker/src/index.ts
```diff
@@ -1005,16 +1036,73 @@ async function handleRelayRpc(
...
+
+    // Canonical hex or nothing, decided before the guard. The guard's lookup is
+    // byte-exact, so a non-canonical pubkey could never match a real case, and
+    // the relay stores these bytes exactly, so it could never enforce either.
+    // 400 rather than the guard's 503: this is the same answer
+    // handleGetActiveAgeReviewCase gives for the same regex, and unlike a D1
+    // outage a malformed pubkey never becomes valid on a retry, so the retryable
+    // class would be a lie. The guard keeps its own check for anything that
+    // reaches it by another route.
+    if (!/^[0-9a-f]{64}$/.test(target)) {
```
realmeylisdev: An uppercase ban can still be created, but can no longer be lifted.

`banpubkey` isn't in the method list above, so it skips this check and forwards the value as-is. `unbanpubkey` is, so it refuses the identical value. I ran both against the worker on a clean tree:

- `banpubkey ["ABCD1234"x8, "spam"]` -> 200, relay called, `params[0]` sent uppercase
- `unbanpubkey` with that same value -> 400, relay never called

Funnelcake stores and deletes by exact bytes (`crates/clickhouse/src/management.rs`: `INSERT INTO banned_pubkeys`, then `ALTER TABLE banned_pubkeys DELETE WHERE pubkey = ?`), and `UserManagement` hands the listed value straight back (`:461` -> `:266` -> `:206`). So a row like that now sits in the ban list with no way to remove it from the UI, including DebugPanel's raw RPC console. On `main` the same click deleted it.

Still reachable: DebugPanel gates its Ban button on `testPubkey.length !== 64` with no hex or case check, and rows written by the Add User dialog before `e190436` may already exist.

Three ways out that I can see — reject the ban too, lowercase on the way through, or let the unban carry a non-canonical value so the row stays removable. Which one is yours to pick, but I don't think it should stay as is.

worker/src/index.ts
```diff
@@ -966,6 +977,26 @@ async function handleModerate(
         });
         // Pass ctx so the unbanpubkey Keycast restore (non-critical) is kept alive.
         const rpcResponse = await handleRelayRpc(rpcRequest, env, corsHeaders, ctx);
+        if (!rpcResponse.ok) {
```
realmeylisdev, line 980: The 503 half of this contract has no detector.

I replaced `!rpcResponse.ok` with `rpcResponse.status === 409` and ran the full worker suite: **384 passed**. Nothing caught it, even though it changes `/api/moderate` `allow_pubkey` under a failing D1 from `503 {code:'age_review_check_failed'}` to a `500` with `code` stripped — exactly the flattening the comment below argues against.

The 409 is pinned (`index.test.ts:515`), but `makeAccountStateEnvWithFailingCaseLookup` is never driven through `/api/moderate`. A sibling test using it would close this. Worth having given how much of the reasoning here rests on that code surviving the hop.

src/hooks/useAgeReviewGuardRedirect.ts
```diff
@@ -0,0 +1,57 @@
...
+ * true it has already handled the error and the caller should return.
+ *
+ * Shared rather than copied because extending the guard to `unbanpubkey` makes
+ * four new call sites able to receive this code — `UserActions` un-ban,
+ * `UserManagement` allow_user and un-ban, and `EventDetail` un-ban — on top of
+ * the four that already handled it (`UserActions` suspend, unsuspend and bulk;
+ * `UserManagement` unsuspend, which was its own inline copy). Eight
+ * near-identical copies is how one gets missed, so the handling lives here.
+ *
+ * Only covers the 409. The same three RPCs also answer 503
```
realmeylisdev: `suspendpubkey` can't produce this 503. `failClosed` is passed as `isReversal` (`index.ts:1102`), and every `cannotCheck()` sits inside `if (opts.failClosed)`. A suspend whose lookup fails returns 200 and enforces — pinned by "still lets a suspend through when the lookup fails" (`index.test.ts:620`).

As written it reads as though suspend is protected against an unrunnable check, which is the opposite of the design you describe. Naming the two reversals instead of "the same three RPCs" would do it.

worker/src/index.ts
```diff
@@ -966,6 +977,26 @@ async function handleModerate(
...
+          // Forward the inner response rather than re-wrapping at 500. The
+          // age-review guard answers a refused unban with a structured 409
+          // (code/caseId/state) that callers route on, and flattening it to 500
+          // both destroys that contract and mislabels a permanent refusal as
+          // transient, which is the shape retrying clients treat as "try again".
+          //
+          // Note this is not 409-only. handleRelayRpc has three reachable
```
realmeylisdev: Four now, not three. The `400 Invalid pubkey` you added at `:1080` is the fourth, and it is reachable from right here — `allow_pubkey` only checks `body.pubkey` for truthiness at `:966`.

Ran it: `POST /api/moderate {action:'allow_pubkey', pubkey:'A'x64}` -> `400 {"success":false,"error":"Invalid pubkey"}`, forwarded verbatim.

Worth naming because the two 400s are indistinguishable by status, so a caller has to read the body to tell malformed input from a relay failure — which is the exact distinction this comment is teaching.

realmeylisdev (changes requested):
Re-reviewed after the five commits. My three earlier findings are all properly fixed, and I re-checked each rather than taking the summary for it:

- The bootstrap no longer escapes. A DDL failure reaches the guard's decision now, so suspend keeps failing open and a reversal is refused.
- `index.test.ts` is 58/58 across five shuffled seeds here.
- The Coop trace is exact at tag `1.0.2`: throw `:537`, caught by `publishAction`'s own catch `:731`, `return false` `:750`, and `withRetries` (`misc.ts:320`) only re-runs on a rejection.

I also went at the premise the change rests on instead of assuming it. `unbanpubkey` does lift an age-review hold: `index.ts:1134` routes to `unsuspendUser`, which is `callKeycast(pubkey, { status: 'active' })` (`keycast-client.ts:95`). Age review can still act on its own cases, since the clear path uses the nip86 helpers directly (`age-review.ts:465`). The 400 is right by NIP-01 and breaks no caller I could find. Your numbers all reproduce: 384 worker, 21 D1, 512 + 1 skipped UI, 58 for `index.test.ts`, tsc clean.

Four things below. One is behaviour: an uppercase ban can still be created but can no longer be lifted. The others are smaller but the kind of thing the next person acts on — two comments that describe something the code doesn't do, and one untested half of the contract this PR argues for at length.

One downstream note, not a change request. `divine-moderation-service` reaches `unbanpubkey` too (`src/index.mjs:765`, live at `:2081`), so this PR newly exposes it to the 409/503. Its `callRelayAdminAction` throws a bare `Error(data.error)` at `:851`, so `code`/`caseId`/`state` are gone before anything can route on them — a moderator un-banning an uploader under age review from that dashboard gets a generic failure. #223 covers the adapter, not this path.

realmeylisdev (approved):
Re-reviewed at `4ce69d8`. My four earlier findings are fixed, and I re-checked each rather than taking the summary for it. One of them was fixed and then re-broken four minutes later by the next commit, which is the first thing below.

I've pushed four commits. Each is one finding and can be reverted alone.

## Pushed

**`ead3f34` — the `/api/moderate` forwarding note, again.** `380b733` correctly changed it from three reachable non-ok returns to four, on the grounds that the canonical-hex 400 was "reachable from right here". `4ce69d8` then narrowed that check to the ENFORCE direction, and `allow_pubkey` issues `unbanpubkey` — a reversal, deliberately exempt. So it is three again, and the "two 400s are indistinguishable by status" sentence describes something that cannot happen. Driven through `worker.fetch`: `POST /api/moderate {action:'allow_pubkey', pubkey:'<uppercase 64-hex>'}` returns `200 {"success":true}` and forwards the uppercase bytes. `index.test.ts` already pins the opposite of what the comment said.

**`9ca2b26` — `divine-moderation-service` reached `unbanpubkey` before Coop did.** The docblock said "Nothing reached it from outside until the Coop enforcement adapter made unbanpubkey callable". That service has mapped `allow_pubkey` -> `{method:'unbanpubkey'}` onto `/api/relay-rpc` since `63cdac3` (2026-06-08), served live at `POST /admin/api/uploader/{pubkey}/enforcement` and pinned by its own test. The adapter's first `unbanpubkey` is `a3d53ec` (2026-06-17); its reversal routes went live 2026-07-27. Worth correcting rather than trimming, because the omitted consumer is the one this worker authorises by name (`allowModRelayAdminKey`) and it is affected — `callRelayAdminAction` throws on any non-ok, so both the 409 and the new fail-closed 503 reach a moderator as an opaque 502. That is divine-moderation-service#191, already linked from the description.

**`1ce5b93` — the guard-409 test comment.** Three claims, none of them true. It is not "the only link": `apiRequest` parses the same structured body and passes `code` into `ApiError` exactly as `callRelayRpc` does, and the pre-existing 409 `version_conflict` test above already pins it. Not "every guarded call": `/api/bulk-moderate` is guarded by the same predicate but routes through `apiRequest`, and `UserActions` branches on it. And "all five" against a list of four.

**`2533470` — pin the two un-suspend guard redirects.** `UserManagement.tsx:255` is the one site this change rewrote from its own inline copy on `main` into the shared hook, which is exactly the edit a hoist can drop; eslint would not catch it either, since the identifier stays in use by the other two mutations in the file. `UserActions.tsx:141` is a pre-existing gap, but a two-line close in a file this change already edits. Both survived deletion before this. Mutation-checked in both directions: with the redirects removed each new test reds and nothing else in either file does; restored, 33 pass in those two files (was 31).

## Description

Corrected six places where it disagreed with the code at head: item 4 named "the three guarded methods" when the 400 covers `{banpubkey, suspendpubkey}` and `banpubkey` is the one the same description says is *not* guarded; the "Why" section credited Coop as the only external caller; and the test-plan counts had drifted (worker 385 not 384, `index.test.ts` 59 not 58, UI 514 not 512). I left the "Since review" table alone — adding rows for my own commits is your call, not mine.

## Still open, and yours to decide

`EventDetail`'s un-ban redirect is the one new call site with no detector: reverting the file wholesale to its `main` shape passes eslint, `tsc`, and all 52 test files. You disclosed exactly this in "Notes for review" and the reasoning holds — standing up the repo's first test file for a 1200-line component with several untested mutations is a scope call, so I did not push it. If you want it later, the minimal version mocks `useAdminApi` so `unbanPubkey` rejects with `new ApiError('under age review', 409, 'Conflict', 'age_review_active')`, clicks Unban, and asserts the navigate, plus the negative case. Note the fallback toast still carries the worker's own actionable text, so what is lost is the click-through, not the information.

## Verified correct

I went at the load-bearing claims rather than assuming them.

- **Guard ordering is genuinely pinned.** I moved the whole guard block to sit after `callNip86Rpc` and re-ran `index.test.ts`: 3 tests fail, each on `expected "fetch" to not be called`. The claim in `99e216f` holds.
- **The ENFORCE/REVERSE split is safe on both legs.** An uppercase `unbanpubkey` is a total no-op: the relay matches bytes exactly, and `callKeycast` rejects it on `HEX_64` before any status change, so nothing lifts a hold. Case rows cannot hide behind it either.
- **`unbanpubkey` really does lift an age-review hold** — routes to `unsuspendUser`, which is `callKeycast(pubkey, { status: 'active' })`. Age review can still act on its own cases, since the clear path uses the nip86 helpers directly.
- **Suites at head:** worker 385, D1 21, UI 514 + 1 skipped, `index.test.ts` 59, `tsc -p tsconfig.app.json --noEmit` and eslint clean.

## Checked and dropped

So you don't re-investigate them: `DebugPanel`'s "Partial success" on a refused un-ban and its audit row — real, but byte-identical to `main` and untouched here. `/api/moderate ban_pubkey` flattening the new 400 to a 500 — the flattening line is outside all three hunks of this diff. `ageReviewActiveGuard`'s "the enforce direction does not reach this" — bulk-moderate cannot enqueue a suspend or ban, so the sentence is fine. osprey's `RelayManagerSink` is a real consumer of this endpoint, but it sends canonical lowercase hex. The `denied_closed` terminal state remaining un-guarded against un-ban is pre-existing, and this change strictly narrows that gap.

One thing that may deserve its own issue rather than a change here: `useReportContext` passes the report's raw `p` tag through to `UserActions` with no normalisation, so a mixed-case tag now gets the new 400 instead of `main`'s 200-that-enforced-on-nobody. That is an improvement in this PR — the failure became visible — but the un-normalised path is still there.

Approving. The worker-side enforcement, which is what this PR is actually about, holds up.

PR #218, opened 2026-07-31
fix(auth): relay management surfaces no longer require a divine-login session

mbradley:
## Summary

Restores relay management for any moderator CF Access admits, with no divine-login session required.

## Motivation

#181 added moderator sign-in for **attribution only** — CF Access stays the access gate (#178). Four components still read `useCurrentUser().user` as an authorization gate, so an admitted moderator got a login wall instead of relay data. Reported from the field: *"I no longer have relay management access."*

The gate was never meaningful: everything behind it goes through `callRelayRpc`, which takes only `apiUrl` and is signed server-side with the shared admin key. No moderator identity was ever involved.

Live impact: the **Settings** tab (NIP-86 section replaced with "Log in to view relay management data") and the **Events** tab (missing banned / needs-moderation markers). `RelaySettings` and `RelayStats` have the same defect but are rendered nowhere; fixed for consistency, not deleted (out of scope).

## Changes

- Removed the `user` gates from `SettingsDashboard`, `EventsList`, `RelaySettings`, `RelayStats`.
- Added `AttributionNotice`: a non-blocking "actions are being recorded without attribution" cue with a sign-in button. Removing the walls also removed the only prompt to sign in; without it every decision silently writes `moderator_pubkey = null`. Hidden for a resolved moderator and during the boot resolve.
- Fixed `<Skeleton>` (a div) nested in `<p>` in the Moderation Overview tiles — latent until this section started rendering by default.

Attribution logic is untouched. `EditProfileForm` keeps its `!user` guard: client-side signing genuinely needs a key.

## Validation

Verified signed-out in a browser against both builds:

| Signed-out page | Before | After |
|---|---|---|
| `/settings` | login wall, **0** relay-rpc calls | sections render, **56** calls |
| `/events` | **0** calls | **16** calls |

New tests confirmed red against `origin/main` before the fix. `tsc -p tsconfig.app.json` clean, `eslint` clean, `vitest run` 52 files / 510 passed, `vite build` succeeds. No worker changes.

**Review tip:** use `git diff -w` on `SettingsDashboard.tsx` — removing the ternary re-indented ~120 lines.

Relates to #178, follows #181.

### DIFF
--- a/src/components/RelayManager.tsx
+++ b/src/components/RelayManager.tsx
@@ -17,6 +17,7 @@ import { Server, FileText, Users, Settings, Flag, Tag, Bug, GripVertical, Clock 
 import { useAppContext } from "@/hooks/useAppContext";
 import AdminBar from "@/components/AdminBar";
 import { DivineLoginButton } from "@/components/auth/DivineLoginButton";
+import { AttributionNotice } from "@/components/auth/AttributionNotice";
 
 // Tab definitions in default order (Reports first for moderation workflow)
 const TAB_DEFINITIONS = [
@@ -173,6 +174,7 @@ export function RelayManager() {
       </header>
 
       <div className="flex-1 min-h-0 overflow-hidden container mx-auto px-4 py-4">
+        <AttributionNotice />
         <Tabs value={currentTab} onValueChange={handleTabChange} className="h-full flex flex-col">
           <TabsList className="shrink-0 grid w-full grid-cols-7">
             {orderedTabs.map((tab) => {
--- a/src/components/RelayManager.tsx
+++ b/src/components/RelayManager.tsx
@@ -173,9 +173,13 @@ export function RelayManager() {
         </div>
       </header>
 
-      <div className="flex-1 min-h-0 overflow-hidden container mx-auto px-4 py-4">
+      {/* Column layout so anything above the tabs (the attribution notice) takes
+          its height out of the tab area instead of pushing the bottom of every
+          tab past `overflow-hidden`. `h-full` on Tabs would claim the full
+          container height regardless of the notice. */}
+      <div className="flex-1 min-h-0 overflow-hidden container mx-auto px-4 py-4 flex flex-col">
         <AttributionNotice />
-        <Tabs value={currentTab} onValueChange={handleTabChange} className="h-full flex flex-col">
+        <Tabs value={currentTab} onValueChange={handleTabChange} className="flex-1 min-h-0 flex flex-col">
           <TabsList className="shrink-0 grid w-full grid-cols-7">
             {orderedTabs.map((tab) => {
               const Icon = tab.icon;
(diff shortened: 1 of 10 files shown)


src/components/RelayManager.tsx
```diff
@@ -172,8 +173,13 @@ export function RelayManager() {
...
 
-      <div className="flex-1 min-h-0 overflow-hidden container mx-auto px-4 py-4">
-        <Tabs value={currentTab} onValueChange={handleTabChange} className="h-full flex flex-col">
+      {/* Column layout so anything above the tabs (the attribution notice) takes
+          its height out of the tab area instead of pushing the bottom of every
+          tab past `overflow-hidden`. `h-full` on Tabs would claim the full
+          container height regardless of the notice. */}
+      <div className="flex-1 min-h-0 overflow-hidden container mx-auto px-4 py-4 flex flex-col">
+        <AttributionNotice />
+        <Tabs value={currentTab} onValueChange={handleTabChange} className="flex-1 min-h-0 flex flex-col">
```
dcadenas, line 182: This is what the two-line layout change is for.

The wrapper is a block box with a definite height (`flex-1 min-h-0 overflow-hidden`), and `Tabs` claimed `h-full` of it. Putting the notice above a sibling that still asks for 100% pushes the content past the wrapper, and `overflow-hidden` quietly cuts about the notice's height off the bottom of every tab — signed out only, which is the state the change is restoring.

Making the wrapper a flex column and giving `Tabs` `flex-1 min-h-0` keeps the signed-in box model identical to the old `h-full`, since the notice renders nothing then. I checked this in a browser against a standalone reproduction of the same box model rather than trusting the CSS reading: a marker at the bottom of the tab area is invisible before the change and visible after. No test came with it — jsdom has no layout engine, so the only assertion available is an echo of the class names.

src/hooks/useStartSignIn.ts
```diff
@@ -0,0 +1,23 @@
+// ABOUTME: Starts the divine-login redirect and reports a failed start as a
+// ABOUTME: toast. Shared by every sign-in affordance so they cannot drift.
+import { useCallback } from 'react';
+
+import { useDivineSession } from '@/hooks/useDivineSession';
+import { useToast } from '@/hooks/useToast';
+
+export function useStartSignIn(): () => void {
```
dcadenas, line 8: The notice copied the header button's sign-in handler verbatim, including the return path and the toast copy. Two affordances that must behave identically now go through one hook, so they cannot drift. Behaviour is unchanged and both components' existing tests still pass.

dcadenas (approved):
✅ the gate removal is the right call — `callRelayRpc` never carried moderator identity, so `user` was only ever attribution.

Pushed two commits on top. The attribution notice was taking its height out of the tab area and clipping the bottom of every tab while signed out, and the sign-in handler it copied from the header button is now shared. Both are explained inline.

Validated on the new head: type-check clean, lint clean, 510 tests passing, production build succeeds, CI green.

PR #219, opened 2026-07-31
fix(auth): surface a signed-in session whose moderator identity never resolves

mbradley:
## Summary

Makes a signed-in session whose moderator identity never resolves visible and escapable, instead of a silent dead end.

## Motivation

Reported from the field: *"there's no logout option and it encourages you to log in and then you log in and then ... you cannot execute anything."*

`user` needs **both** a signer and a pubkey. OAuth yields the signer; the pubkey needs a separate `getPublicKey()` call to login.divine.video that can fail on its own (no Keycast-managed key, RPC error, token without signing scope). The provider swallowed that failure, so `user` stayed undefined — and since sign-out renders only `if (user)`, the header fell back to "Sign in". The moderator was signed in, couldn't tell, couldn't leave, and every action recorded null attribution. Clicking "Sign in" looped them through the same failure.

That is the worst shape for an attribution feature: it puts false confidence into the audit trail.

## Changes

- The session distinguishes "no session" from "a session with no identity" (`identityUnavailable`), and the header renders it: "Identity unavailable, actions unattributed" plus the missing sign-out.
- Logs the `getPublicKey` rejection, previously swallowed by an empty `catch` — the only way to tell a missing Keycast key from a login-server error.
- Keeps the state honest across transitions, not just at boot: a rotated token whose identity fails no longer keeps the old pubkey (the header would otherwise assert a moderator while writes went null); a token that blips and returns re-resolves without flashing the error; a bunker-only session (`accessToken` is optional) is treated as the same dead end rather than a plain "Sign in".

Added during review:

- The session hands out a pubkey only while the token it was resolved for is still live, so a rotated token cannot pair the previous moderator's pubkey with the new session's signer.
- Relay management is gated on session presence rather than on the resolved pubkey. Those calls are authorized by the worker and never read the moderator identity, so a session with no identity was being told to log in while already signed in.
- Sign-out is reachable while the identity is still resolving. A single `getPublicKey` can retry for minutes, and that window previously rendered nothing but a loading placeholder.
- The banner label is hedged to match its tooltip, and widened so the consequence is not truncated away.

Attribution behaviour is unchanged — `getModeratorPubkey()` is byte-identical to `main`. This only makes its absence visible.

## Validation

Test-first; every behavioural guard is mutation-checked (removed individually, each reds a specific test). Two review rounds; the second caught that one guard had been dropped on a mutation result that measured settled state rather than the intermediate render it protects — restored, now pinned by a test that records every render.

`tsc -p tsconfig.app.json` clean, `eslint` clean, `vitest run` 52 files / 522 passed, `vite build` succeeds. No worker changes.

**UI:** the header in its failure and loading states, plus the relay-management panes, which no longer ask an already signed-in moderator to log in. Working and signed-out sessions render exactly as before, both pinned by regression tests.

## Known limitation

`identityUnavailable` is a boolean, so the copy can't distinguish "no Keycast-managed key" (where signing out won't help) from a transient failure. The console warning carries the reason; a typed reason is a follow-up.

Relates to #178, follows #181.


src/components/DivineSessionProvider.tsx
```diff
@@ -76,6 +88,15 @@ export function DivineSessionProvider({ children }: { children: ReactNode }) {
...
   const accessToken = credentials?.accessToken;
+  // Derived during render, like identityResolved below and for the same reason:
+  // the signer is rebuilt from the live token immediately, so an effect that
+  // cleared a stale pubkey would land one render too late, and for that render
+  // the session would hand out the previous token's pubkey paired with the new
+  // token's signer. isResolving hides that window in the header, but
+  // useCurrentUser does not expose isResolving, so a consumer reading
+  // user.pubkey (EditProfileForm loads a profile from it) would see one
+  // session's identity attached to another's.
+  const pubkey = resolvedIdentity.token === accessToken ? resolvedIdentity.pubkey : undefined;
```
dcadenas, line 99: **A rotated token used to hand out the previous token's pubkey for one committed render.**

`identityResolved` was derived during render so it could not lag the token, but `pubkey` was plain state cleared from an effect. The signer is rebuilt from the new token synchronously, so between the rotation and the effect the session exposed one session's pubkey with another session's signer. I probed it before changing anything: exactly one render, with `isResolving` already true.

The header was safe because it checks `isResolving` first, but `useCurrentUser` does not expose `isResolving` and none of its consumers check it, and `EditProfileForm` loads a profile from `user.pubkey`. Storing the pubkey with the token it came from makes the drift unrepresentable rather than corrected after the fact. I tried the effect-based reset first and it failed the test — one render late, the same shape you had already fixed once for `resolvedForToken`.

src/contexts/DivineSessionContext.ts
```diff
@@ -12,8 +12,25 @@ export interface DivineSessionValue {
...
   signer: NostrSigner | null;
+  /**
+   * A session exists, whether or not its moderator identity resolved. Gate UI on
+   * this rather than on `useCurrentUser().user`, which additionally requires the
+   * pubkey: relay management authenticates through the worker (CF Access service
+   * token / X-Admin-Key), so it does not need the moderator's identity, and
+   * gating it on the pubkey locks out exactly the session `identityUnavailable`
+   * describes -- which would still be told to "log in" while already signed in.
+   */
+  isSignedIn: boolean;
```
dcadenas, line 23: **Relay management was gated on the pubkey, which the worker never reads.**

`RelaySettings`, `RelayStats`, `SettingsDashboard` and `EventsList` gated on `useCurrentUser().user`, so a session with no identity was told "Please log in to manage relay settings" while the new banner directly above it said they were signed in. Those calls go through `adminApi.callRelayRpc`, which sends only the CF Access service token and `X-Admin-Key` — no pubkey, no signer — so the gate was vestigial.

This is the widest thing I changed and the one I would most understand you pushing back on, since it reaches outside the header. I took it because without it the banner's "Moderation itself still works" is not true for those panes. Behaviour is unchanged for signed-out and fully-resolved sessions; only the identity-unavailable case moves. It is its own commit if you would rather drop it.

src/components/auth/DivineLoginButton.tsx
```diff
@@ -29,7 +29,27 @@ export function DivineLoginButton() {
...
+            four attempts and 401 once, each with its own 30s timeout and an
+            unclamped server-supplied Retry-After, so a single getPublicKey can
+            stall for minutes. Without this the moderator spends that window
+            looking at an aria-hidden placeholder with no sign-out and nothing
+            for a screen reader -- the same dead end this component exists to
+            remove, reached through a slower door. No timer and no new state:
+            the escape hatch is simply always available once a session exists.
+            Gated on isSignedIn so a signed-out visitor, who is also briefly
+            "resolving" at boot, is never offered a sign-out. */}
+        {isSignedIn && (
```
dcadenas, line 45: **A stalled resolve reached the same dead end by a slower route.**

A `getPublicKey` that rejects is handled well. One that stalls rendered only the `aria-hidden` skeleton: no sign-out, and nothing at all for a screen reader. I had assumed that window was capped at 30s, and that was wrong — `AbortSignal.timeout(3e4)` bounds each request, while `DivineRpc.call` retries 429 up to four attempts and 401 once, with an unclamped server-supplied `Retry-After`. So it is minutes wide.

Rendering the escape hatch unconditionally avoids adding a second timeout that could disagree with the SDK's. The `isSignedIn` gate is load-bearing in the other direction: a signed-out visitor is also briefly resolving at boot and must not be offered a sign-out. Both halves are pinned separately.

This window exists on `main` too, so it is not something the branch introduced.

src/components/auth/DivineLoginButton.tsx
```diff
@@ -46,6 +66,50 @@ export function DivineLoginButton() {
...
+              because the header's outer flex row (RelayManager's gap-3
+              container) keeps min-width:auto, so a very narrow header overflows
+              instead of truncating. Measured: min-width:0 on this label, on the
+              banner span, and on the gap-2 wrapper all change nothing -- and
+              unblocking the whole chain is not the fix either, since the label
+              then shrinks to nothing at 375px, which is worse than an overflow
+              on a desktop-only tool. That overflow predates this banner and
+              affects the signed-in name the same way. jsdom performs no layout,
+              so no test can pin any of this. */}
+          <span className="max-w-[24rem] truncate">Identity unavailable, actions may be unattributed</span>
```
dcadenas, line 103: **The label now hedges, and is wide enough to say so.**

`DivineRpcSigner.getPublicKey` caches only on success, so a rejected resolve is retried on every later `getModeratorPubkey()` call. After a transient failure the banner stays latched for the life of the token while attribution has already recovered — so "actions unattributed" stated as fact was wrong for exactly that window. This matches the hedge you had already put in the title.

The longer string then truncated inside the original `max-w-[16rem]`, cutting off the consequence, which is the reason the consequence was moved onto the label in the first place. Measured against the built CSS in a headless browser: 300px natural, truncated against 256px, clear against 384px. Your original shorter label was 250px and never truncated.

No test can pin this — jsdom performs no layout — so a screenshot at a realistic header width is worth thirty seconds before merge.

dcadenas (approved):
The mechanism here is right, and the instrumentation is what I want to credit first: recording state on every render rather than after it settles is the only thing that catches this class of bug, and the self-correction that a mutation measuring settled state proved nothing about the intermediate render it guards is the correct diagnosis. `getModeratorPubkey` really is byte-identical to `main` — I checked rather than took it. I went looking for a path where the header asserts one identity while a write records another, and did not find one.

I pushed five commits instead of leaving you a list. Two are behaviour: the session now hands out a pubkey only while the token it belongs to is live, and sign-out stays reachable while the identity is still resolving. One reaches past the header — relay management was gated on the resolved pubkey, so those panes were telling an already signed-in moderator to log in — and it is a separate commit if you disagree with the scope. The last two hedge the banner label and correct two comments that stated something untrue, one of which was mine. The reasoning is on the lines.

`tsc -p tsconfig.app.json` clean, `eslint` clean, `vitest run` 52 files / 522 passed, `vite build` succeeds, and every guard I added reds a specific test when removed. I also corrected the description, which said "header only" and carried the old test counts. Two things I deliberately left: a bounded retry for a transient identity failure, which needs a backoff policy and a "still failing versus recovered" affordance the header cannot express, and closing the remaining boot-resolve window by seeding credentials from the SDK's synchronous session read. Both are follow-ups rather than gaps in this branch.

PR #222, opened 2026-08-04
feat(age-review): capture account identity so tickets and contacts say who they are about

mbradley:
## Summary

Age-review cases are keyed entirely by pubkey, so nothing we store can say *who* a case is about. When a parent writes in, tracing their ticket back to a case means pivoting manually through their other tickets. The one readable signal — the kind-0 profile — is destroyed by our own enforcement: `suspendpubkey` hides the account's content, so by the time a moderator opens the case there is nothing left to look at.

This captures a human-readable identifier at case creation, while one is still visible, and puts it where the people doing the work will see it.

Closes #213

## What changes

**Capture.** All three case-creation paths now record `account_name`, `account_nip05`, `account_vine_username` and `identity_captured_at`. The report-driven paths fetch kind-0 ahead of enforcement; the operator-created path already knows the identity and just stores it. Capture is best-effort three layers deep — it can never be the reason a case fails to open.

**Surfacing.** The case ticket's internal note now leads with an identity block and a deeplink; it previously carried a bare pubkey and no way to reach the case it described. The parent's contact record gets the same block in `notes`.

**The contact name is staged deliberately.** At attach time it is the email address alone. It becomes `Claimed parent of <handle>` only once the parent replies. Zendesk renders the stored name into the `To:` header of outbound mail, and the address is supplied by the account under review and never verified — so naming the contact after the account any earlier would disclose a real handle to whoever holds a mistyped address. A reply proves the address is live and held by someone engaging with the review. "Claimed" is deliberate: whether they are the parent is precisely what the review exists to establish.

That staging is what makes the queue readable — the Requester column renders the name and nothing else.

## Notable findings from review

Two things worth flagging, since both changed the design rather than just the code:

- **The parent-address guard was checking the wrong thing.** Verifying `parent_contact_email` is non-null in D1 does not establish that the ticket's requester *is* that parent. Zendesk resolves a requester by email to an existing user and permits agents to be requesters, so an account under review naming a Divine staff address would have made that staff member the requester — and since Zendesk display names are global, the rename would have put a minor's handle in the header of every mail that person subsequently sent. Both writes now resolve the contact and require an end user at exactly the address on the case.
- **The operator path was discarding the username** whenever a display name was supplied. Permanently, since those rows stamp `identity_captured_at` and a backfill keyed on `IS NULL` never revisits them. It is now preserved as the NIP-05 it implies, with the identity domain in `NIP05_DOMAIN` config — staging accounts come from a different Keycast instance, so an unconfigured environment records nothing rather than a plausible but wrong address.

A later adversarial round found two more, both in the capture path itself:

- **A failed relay lookup was recorded as a successful one.** `identity_captured_at` was stamped unconditionally, but the lookup returns nothing for four distinct reasons — no kind-0, timeout, socket error, no relay configured. A brief relay blip therefore wrote a null name *plus* a capture timestamp, which the backfill reads as "looked, found nothing" and skips permanently; enforcement then hides the profile and the identifier is gone. The lookup now reports whether it completed, and only a relay-confirmed answer stamps. `queryRelay`'s `complete` flag is honoured too, so a socket that closes before EOSE counts as unconfirmed rather than as an empty result.
- **Stored names were mangled by a display sanitizer.** `parseKind0Profile` strips markdown punctuation and truncates at 80 characters, which is right for a note and wrong for a column — a display name of `*Star*Girl*` persisted as `StarGirl`, matching nothing searchable. Capture now stores raw; both render surfaces already sanitize on the way out, and their tests feed raw newline-bearing names to prove it.

## Deliberately not here

- **The backfill.** Needs a staging dry-run and a reported yield split before prod — its own review cycle. Follow-up PR.
- **Outbound email identity and wording.** Being decided separately. This PR leaves the pre-reply message exactly as handle-free as it is today.

## Zendesk-side dependencies — both resolved

Both were verified against the live instance and neither blocks this.

1. **The parent-reply trigger was never a risk.** `Age Review: notify relay-manager on parent reply` gates on `role is end_user` alongside `comment_is_public` and the `age-review` tag, so an agent's own follow-up cannot fire the webhook, advance a case, or trigger the rename. The staging this PR's privacy constraint rests on is intact.

2. **The macro is fixed**, in `divine-zendesk-tooling`. `Customer not responding` set no comment mode at all, so it inherited the agent's composer state — meaning it either emailed the customer or wrote a private note depending on an incidental UI setting. On an age-review ticket both outcomes do damage: public clears `age-review-parent-replied` and drops the ticket out of the waiting-on-us view, internal keeps it queued but never nudges anyone. It is now pinned public and no longer interpolates the requester name, so it cannot render `Hello Claimed parent of <handle>.` That defect predated this PR.

## Validation

| Check | Result |
|---|---|
| `cd worker && npx vitest run` | 433 passed |
| `cd worker && npm run test:d1` | 25 passed |
| `cd worker && npm run typecheck` | clean |
| `npm run test` (root) | 500 passed, 1 skipped |
| `npx tsc -p tsconfig.app.json --noEmit` | clean |

These are the four gates CI runs. There is no `npm run lint` in this repo, at either level — the plan doc said otherwise and has been corrected.

Rebased onto current `main`; #209's client-name normalization and this branch's capture ordering verified to coexist. Every guard in the diff was mutation-tested — the guard was removed and the test confirmed to go red. Across two mutation passes, thirteen tests that could not fail were replaced. The recurring cause is worth naming for review: where a guard sits upstream of a best-effort `try/catch`, asserting "the write did not happen" passes whether or not the guard exists — remove it and the code crashes into the same catch, producing an identical outcome. Those guards now assert on a call that was never made or a log that never appeared, and each dies to a mutation of that one clause alone.

No UI changes. Zendesk-visible changes are the contact name, contact notes, and the internal ticket note.


worker/src/report-note.ts
```diff
@@ -73,9 +73,51 @@ function sanitizeInline(value: string | undefined, maxLen = 80): string | undefi
...
+ * `JSON.parse` yields whatever the account put in the field, and the declared
+ * `string | undefined` is a lie for `{"display_name": {}}` or `["a","b"]`.
+ * `sanitizeInline` absorbed that by accident (`Array.from` on a plain object
+ * gives `[]`, so it returned undefined); raw mode has no such accident, and a
+ * non-string reaches `.bind()` as `D1_TYPE_ERROR: Type 'object' not supported`.
+ * That throw propagates out of the case INSERT, so an account could suppress
+ * its own age-review case by publishing a kind-0 whose display_name is not a
+ * string. Capture is best-effort by contract and must never be able to do that.
+ */
+function rawForStorage(value: unknown): string | undefined {
```
dcadenas, line 99: **A kind-0 whose `display_name` was not a string stopped the case from opening.**

`raw: true` skipped `sanitizeInline`, which is right for a stored handle. But the declared `string | undefined` is not what `JSON.parse` returns: `{"display_name": {}}` yielded the object itself. The display path narrowed that away by accident — `Array.from` on a plain object gives `[]` — and raw mode had no such accident.

D1 rejects it: `D1_TYPE_ERROR: Type 'object' not supported for value '[object Object]'`, reproduced against Miniflare. The throw escapes the INSERT in `ReportWatcher.createAgeReviewCase`, which is caught and logged one level up, so the case is silently never created — and the dedup check finds no row, so every later report fails the same way. An account could suppress its own age review by editing its own profile.

`rawForStorage` requires a string and caps at 512, well past any real handle; the existing 120-character no-truncation test is untouched. Raw mode had no direct coverage before, which is how this got through the guard mutation pass — the new tests cover object, array, number and boolean, at the parser, through the socket, and against real D1.

worker/src/age-review.ts
```diff
@@ -1093,11 +1421,26 @@ export async function handleAgeReviewReplyWebhook(
...
+  // this handler means at all -- the Zendesk trigger gates on an end-user public
+  // comment on an age-review ticket. Tying it to `changes === 1` made it
+  // one-shot: upgradeParentContactName swallows its errors and the handler
+  // answers 200 regardless, so Zendesk never redelivers, and the next reply
+  // returns early because the case has already advanced. A single transient
+  // Zendesk blip therefore left the Requester column showing a bare email
+  // permanently, with no path back -- and that column is the readable-queue
+  // payoff the staging exists to unlock. Renaming on every delivery is
+  // idempotent, still gated on resolveVerifiedParentContact, and self-heals.
+  let advanced: AgeReviewCase | null = null;
```
dcadenas, line 1436: **The rename was one-shot, so a single Zendesk blip lost it permanently.**

`upgradeParentContactName` swallows its own errors and the handler answers 200 either way, so Zendesk never redelivers. Tied to `changes === 1`, the only delivery that could ever rename was the one that won the CAS. A later reply returned early, because by then the case had already advanced.

The loop now records the outcome and breaks, and the rename runs once before all three exits. Its real precondition is that the parent replied, which is what reaching this handler means at all — the trigger gates on an end-user public comment on an age-review ticket. The write is idempotent and still gated on `resolveVerifiedParentContact`, so nothing about the privacy staging changes.

worker/src/age-review.ts
```diff
@@ -598,11 +614,50 @@ export async function handleCreateMinorAccount(
...
+    //  - account_nip05 records the address the account is expected to claim.
+    //    Divine's NIP-05 is the subdomain form `_@<username>.<domain>`, so the
+    //    username is recoverable from it. Derived from what the operator gave
+    //    us, not read back from Keycast -- the create-minor-account response
+    //    carries no nip05. No fallback domain on purpose: staging accounts do
+    //    not live under the production identity domain, and guessing one would
+    //    write a wrong address into a record agents read to decide who a case
+    //    is about. Unconfigured stores nothing rather than something false.
+    //
+    //  - account_vine_username holds the username itself, unconditionally.
```
dcadenas, line 635: **Deriving the NIP-05 kept the username only where `NIP05_DOMAIN` is set.**

Staging deliberately leaves it unset, so a case created there with a `display_name` stored the display name and nothing else — the username was gone. This path also stamps `identity_captured_at`, and the backfill keys on `IS NULL`, so it never comes back for the row. That is the same permanent loss the derivation was added to close, still open in the one environment shipped without the config.

`account_vine_username` is unused on this path and `resolveHandle` already falls back to it, so the username now goes there unconditionally, alongside the derived NIP-05 where a domain exists. The staging comment is reworded to describe a steady state rather than a pending task, since nothing is lost by leaving the variable unset.

worker/src/relay-profile.ts
```diff
@@ -0,0 +1,172 @@
...
+    // this branch's merge base), which rewrites queryRelay so an early close or
+    // a NIP-01 CLOSED is `success: false` -- while still returning the events
+    // that did arrive. Testing `success` first would then throw away a kind-0
+    // that a relay streamed before closing without EOSE, leave
+    // identity_captured_at null, and let enforcement hide the profile before any
+    // backfill could run: precisely the permanent loss this capture exists to
+    // prevent. Under the current vendored queryRelay the two orders are
+    // equivalent -- its `success: false` path carries no events -- so this is
+    // cover for the rebase, not a behaviour change here.
+    if (res?.events?.length) {
```
dcadenas, line 142: **Presence has to be checked before absence, or the rebase discards a captured profile.**

An event in hand is a confirmed answer however the read ended. Only absence needs the read to have finished cleanly before it can be believed.

Under the vendored `queryRelay` the order does not matter, because its `success: false` path carries no events. Under #186 it does: an early close or a NIP-01 `CLOSED` reports `success: false` while still returning the events that arrived. Testing `success` first would then throw away a kind-0 that a relay streamed before closing without EOSE, leave `identity_captured_at` null, and let enforcement hide the profile before any backfill ran.

`captures a profile that arrived before the relay closed without EOSE` passes under either order at this head, so it is not yet load-bearing. Patching the vendored close handler to #186's shape locally makes it so: presence-first stays 12/12, success-first goes red. That flip after the rebase is the signal the port was done right.

worker/src/relay-profile.ts
```diff
@@ -0,0 +1,172 @@
+// ABOUTME: Shared relay query helpers and best-effort account identity lookup.
+// ABOUTME: Extracted from index.ts so ReportWatcher and age-review can use them without a circular import.
+
+import { parseKind0Profile, type ReportedProfile } from './report-note';
+
+// Query relay for events matching a filter
+export async function queryRelay(
```
dcadenas, line 7: **This copy predates #186, and it is what `/api/reports` runs on this branch.**

`index.ts` no longer defines `queryRelay`; it imports this one. So resolving the rebase by keeping the branch side would revert #186 for its real consumer, not just for identity capture — the timeout and `CLOSED`-as-error change, the `closeQuietly` hang fix, the `if (resolved) return` guard and the NIP-01 `CLOSED` branch all go with it.

#186 merged on 2026-08-07, after this branch was last rebased, and its own commit message says it should land first. The conflict is `worker/src/index.ts`: main rewrote `queryRelay` in place, this branch deletes it there.

Suggested resolution — port main's `queryRelay` body into this file rather than keeping this copy, then drop the now-dead `res.complete ?? res.success` fallback below and reconcile `worker/src/reports-filter.ts`, whose `result.complete` read #186 also removes. Base sync is yours; this is only how I would take it.

dcadenas (approved):
The design holds up. I tried to break the ordering invariant, the privacy staging on the contact name, and the requester verification, and could not — the enforcement-ordering claim in particular is accurate: `banEvent` is the only enforcement leg that runs before capture, and it leaves the author's kind-0 queryable.

Five defects fixed on the branch, one commit each. Notes inline on each, and on how to resolve the rebase.

### Changes I pushed

- Captured identity is coerced to a bounded string before it is stored. A kind-0 whose `display_name` was an object reached `.bind()` and threw, so the case never opened.
- The parent-contact rename now runs on every reply delivery, not only the one that wins the state CAS.
- The operator-created path stores the username outright, so it survives an unset `NIP05_DOMAIN`.
- `identity_captured_at`'s documented contract now matches the stamp behaviour.
- The relay lookup keeps a profile that arrived before an early close, and reads the completeness flag in a way that survives the rebase onto #186.

The last two are cover for #186, which merged after this branch was last rebased. They are correct at this head and correct after the rebase; the rebase itself is yours, and there is a note inline on how the conflict should be resolved.

### Checks

Run on the pushed head, `49609bd`:

- `worker`: `vitest run` 463 passed, `test:d1` 26 passed, `typecheck` clean
- root: `tsc -p tsconfig.app.json` clean, `eslint` clean, `vitest run` 514 passed / 1 skipped, `vite build` clean
- Full CI workflow green, including the root `npm run test` gate

The `test` check does not appear on the pull request itself. That is the merge conflict, not the code: the workflow triggers on `pull_request`, and GitHub cannot build a merge ref while the branch is conflicting. The same conflict was present before I pushed, and none of my commits touch the conflicting file, so I ran the workflow directly against the branch to get a real result.

### Follow-ups

- An identity note on the ticket created by the parent-contact path when no case ticket exists yet. Out of scope here: #213 asks for the block on contact `notes` and the case ticket's internal note, and the design's decoration-points table lists the same surfaces.

PR #226, opened 2026-08-06
fix(reports): stop presenting handled targets as unhandled when resolution state is incomplete

### LINKED ISSUE #221: fix: reports queue presents handled targets as unhandled when resolution state is incomplete
## Summary

Moderators are re-clearing work they already did. Shauna raised this after repeatedly dismissing the same accounts in the reports queue, including accounts that were already suspended or banned.

The queue hides a target when it appears in `resolvedTargets` (`Reports.tsx:487`), a set built from four independent sources: resolution labels, NIP-86 banned pubkeys, NIP-86 banned events, and D1 moderation decisions. Each is a separate polled query with `retry: false`. Nothing distinguishes "this target has no resolution" from "I could not load the resolution state." When any source is missing or partial, its targets silently reappear as unhandled work.

The filter logic itself is correct. Replaying the queue at three moments where a moderator dismissed the same account, using period-correct windows for all sources, the target was resolved and correctly hidden every time. The card could only have surfaced through incomplete resolution state.

This is distinct from #185, which covers `queryRelay` reporting timeouts as success. That is a fourth source of incompleteness, and its fix does not address the three below.

## Steps to Reproduce

1. Dismiss a report. A resolution label is published and a `dismissed` decision row is written. The card disappears.
2. Hard-reload the reports page.
3. Before `/api/decisions` and `/api/resolution-labels` resolve, the queue renders every target it has, including the one just dismissed.

For the cap path: a target whose only resolution signal is older than the newest 1000 decisions and newest 500 labels reappears permanently.

## Expected Result

A handled target never presents as unhandled. If resolution state cannot be loaded or is known to be truncated, the queue says so rather than rendering work as pending.

## Actual Result

1. **First-load race.** `placeholderData: (prev) => prev` protects mid-session refreshes but gives nothing on a fresh mount. `reports` resolves from the relay in roughly 3s while `/api/decisions` returns ~147KB over 1000 rows and `/api/resolution-labels` makes its own relay round trip. Whichever source has not landed contributes nothing, and the queue renders unfiltered meanwhile.

2. **Error with no prior data.** All four queries use `retry: false`. On a fresh mount an error leaves the source empty until a later 15s poll succeeds. The comment at `Reports.tsx:443` already records cold-start timeouts on `/api/decisions` as the reason `retry` was disabled.

3. **Silent row caps.** `/api/decisions` returns the newest 1000 rows of 3399 (`worker/src/index.ts:1348`). `/api/resolution-labels` returns the newest 500 of 1420 (`index.ts:538`). Both windows currently reach back to mid-June so neither binds against the 200-report queue window today, but a target whose resolution ages out of both un-hides permanently. Only 2 of 1420 resolution labels are `ban`, so takedown resolution leans almost entirely on the decisions table and the ban lists, which makes the decisions cap load-bearing.

## Additional Context

Measured against prod: one account accumulated 13 dismissal labels and 14 `dismissed` decision rows against 11 reports. Four dismissals came after it was already suspended, and three landed within 51 seconds of each other, matching the 15s `refetchInterval`. The reasons moderators typed ("duplicate", "duplicate already suspended") show they believed they were looking at separate items rather than one item returning.

Worth separating from the reappearance problem: genuinely new reports on already-enforced targets are rare. Over the last ~60 days, 10 of 800 reports (1.3%, across 5 targets) arrived after an enforcement action on that target.

Suggested acceptance criteria:

- The queue does not render a `resolvedTargets`-filtered list until every resolution source has loaded or is explicitly reported unavailable.
- When a source is unavailable, show a notice with last-updated time instead of presenting its targets as pending.
- `/api/decisions` and `/api/resolution-labels` either return complete data for the queue's report window or signal truncation.

mbradley:
## Summary

Moderators were re-clearing work they had already done. The queue hides a handled target
by subtracting a `resolvedTargets` set built from four independently polled sources:
resolution labels, banned pubkeys, banned events, and the moderation decisions log. That
set is subtractive, so a source that has not loaded, has errored, or has been silently
truncated does not make the queue smaller and safer. It makes it bigger and wrong.
Nothing distinguished "this target has no resolution" from "I could not load the
resolution state."

Three holes are closed:

- **The first-load race.** Only the decisions query gated the render; the other three
  raced. The queue now waits for every source that affects what is filtered.
- **Errors with no prior data.** All four queries used `retry: false`, so one cold-mount
  failure left a source empty until a later poll. A cold failure now blocks with a pane
  naming the unavailable source, offering Retry and an explicit override.
- **Silent row caps.** `/api/decisions` returns the newest 1000 rows and
  `/api/resolution-labels` the newest 500. Both now report whether they truncated and how
  far back they reach, and the queue states that boundary.

## Motivation

Closes #221.

Raised after a moderator repeatedly dismissed the same accounts, including accounts
already suspended. Replaying the queue at three such moments, with period-correct windows
for all four sources, the target was resolved and correctly hidden every time. The filter
logic was never wrong; the card could only have surfaced through incomplete resolution
state. The reasons typed at the time ("duplicate", "duplicate already suspended") show
moderators believed they were seeing separate items rather than one item returning.

## What moderators will see

Unchanged in normal operation, except that a hard reload takes slightly longer to paint
because the queue now waits to know what has already been handled.

- **A stale source**: queue stays up, filtered by the last good data, with a banner naming
  the source and its age.
- **A cold failure**: what is unavailable, with Retry and "Show the queue anyway."
  Overriding renders under a warning that persists while it is in effect.
- **Offline**: says so, and offers the same override. A paused query never resolves on its
  own, so without an escape a false-negative `navigator.onLine` (captive portals, some VM
  and container stacks) could strand a moderator who actually has connectivity.
- **A capped window**: the date resolution history reaches back to, in UTC.

## Design decisions

**Cold blocks, warm does not.** A warm failure still holds previous data, and that stale
set still hides the handled work, so blocking would cost availability for no safety gain.

That rationale has a limit worth stating: it holds only up to the snapshot. If the
decisions refresh starts failing immediately after a moderator acts, their own newest
dismissal is the row missing from the stale set, and that target can reappear. Not a
regression, since `placeholderData` produced the same filtering before with no banner at
all, and blocking there would be worse. But the warm path is not "safe," and the banner
naming the data's age is doing real work.

**Decisions gates unconditionally; the other three do not.** It feeds
`pendingReviewTargets` as well as `resolvedTargets`, and `pendingReviewTargets` applies on
every path. The other three matter only while the resolved filter is applied, so blocking
on them elsewhere would be a lie.

**The override exists because a blocked queue is its own failure.** Blocking inherits the
combined failure probability of four independent queries on every cold load. It bypasses
only the blocked branch; it never disables a filter.

**The coverage boundary renders in UTC.** In local time a 2026-06-14 00:00 UTC cutoff
shows as Jun 13 in US Pacific, claiming history reaches further back than it does. That is
the falsely reassuring direction. The boundary describes the data, not the reader's
calendar day.

## Validation

- `npm run test`: 538 passed, 1 skipped, build clean
- `cd worker && npm run typecheck`: exit 0 (separate CI gate, has been red here while every
  test suite was green)
- `cd worker && npx vitest run`: 375 passed
- `cd worker && npm run test:d1`: 22 passed, including a real-D1 test for the decisions cap
- `npm run test:e2e`: not run, requires a local relay
- Build ships both the staging and prod API domains

A mutation pass removed each guard one at a time. Seven survived, meaning a future edit
could have removed them silently; five were closed with tests proven to die to their own
mutation, two resolved by decision. The headline survivor was the decisions gating
asymmetry, which had no coverage at all because every test ran with hide-resolved at its
default, so that source always gated through the other term.

## Notes for reviewers

- Overlaps #186 (issue #185) in `Reports.tsx` and the resolution-labels route. Cut from
  `main`, not stacked; rebases once #186 lands. #186's labels-only banner fires on the
  cold-error condition this branch blocks, so the unified policy supersedes it and its
  warm-case text lives on in the stale banner.
- Worker and Pages deploy separately. The client treats a missing `truncated` as false, so
  a new frontend against an old worker degrades quietly rather than warning falsely.
- Adds one devDependency, `@testing-library/user-event`, used only by the new test file to
  click the override. Dev-only, so no bundle or cold-start impact, and the standard
  companion to `@testing-library/react` already in the repo.
- Signalling a capped window is not the same as eliminating it. Scoping the resolution
  reads to the queue's own report window is deliberately left out.


hm21 (dismissed):
Reviewed against `origin/main` (b0af9bf). Pushed two test-only commits; four things left for you.

## Pushed

**`69966c5`** — pins `TZ: America/Los_Angeles` in the vitest env. Both UTC guards (`timeZone: 'UTC'` in `TruncatedHistoryBanner`, the `Z` suffix in `parseOldestCovered`) are unfalsifiable on CI: `test.yml` sets no `TZ` and GitHub's Ubuntu images are UTC, so local time and UTC agree and mutating either guard out leaves the suite green there. With the pin, both mutations fail and the six existing straddle-midnight fixtures needed no edits.

**`10fa845`** — killing tests for three guards that survived mutation: decisions gating on *loading* (the descriptor that replaced main's `decisionsLoading` CSAM guard), banned-events gating (the only one of the four sources with no gating test), and `retry: 1`. Each dies to exactly one mutation and no other.

One further mutation survives and I deliberately left it: dropping `&& !showPendingReview` from `resolvedFilterActive` changes nothing observable, because enabling Pending review force-sets `hideResolved = false` (`Reports.tsx:1343`) and the hide-resolved Switch is `disabled={showPendingReview}` (`:1365`). The term is unreachable by construction, so no honest test can kill it.

Validation: `npm run test` 541 passed / 1 skipped plus build; `worker` typecheck, `vitest run`, and `test:d1` all green.

## Left for you

**1. The escape hatch disappears every poll.** `blockingErrors` reads live status, but query-core's `fetchState` resets `{error: null, status: 'pending'}` on any refetch of a query with `data === undefined` (`query.js:296` → `:353`), and `refetchInterval` keeps firing on errored queries. The skeleton branch (`Reports.tsx:1066`) is checked *before* the pane (`:1102`), so a persistently cold-failing source spends ~61s (30s `API_TIMEOUT_MS` + 1s backoff + 30s) back on the bare skeleton per ~15s of pane. Retry and "Show the queue anyway" vanish under the cursor. Latching on `errorUpdateCount`/`failureCount` would hold the pane across the retry.

**2. Cold-load worst case is that same ~61s of skeleton with no affordance**, up from 30s, and now reachable through any of four sources rather than one. Every escape hatch this PR adds sits behind that branch. Either surface Retry/override inside the loading branch after a few seconds, or lower `timeoutMs` for the four 15s-polled reads — a 30s bound is twice the poll interval that would recover them anyway.

**3. The truncation banner is hidden in the one view built entirely from the truncated read.** It renders only when `resolvedFilterActive` (`:1225`), but `pendingReviewTargets` is derived 100% from `allDecisions` (`:574`) — the read capped at 1000 — and enabling Pending review sets `hideResolved = false`. An `auto_hidden` row that ages out drops its target from the CSAM queue and out of the default view's exclusion, with the banner that exists to state that boundary switched off. The `gatesAlways: true` reasoning for decisions applies to the banner too.

**4. `resolutionOverride` never resets** — only ever set true (`:1057`, `:1108`), and not scoped to the sources it was granted for. An override taken for "Banned posts" silently covers a later cold `decisions` failure, so the moderator never sees the auto-hidden consent step. Holding a `Set` of acknowledged source keys would match the pane's own per-source copy.

Nothing here touches the design, which I think is right — including cold-blocks/warm-doesn't, and leaving `queryRelay` partial reads to #185/#186.

rabble (commented):
Review result: the four previous blockers appear addressed by the latest commits (retry escape-hatch persistence, bounded reads, load-bearing decisions truncation messaging, and source-scoped overrides). I cannot approve because the branch now conflicts substantially with main across Reports, admin API, worker, and tests. Rebase and resolve while preserving both sides' newer behavior, rerun checks, then request re-review.

mbradley:
Merged main to clear the conflicts. Went with a merge rather than a rebase since the branch had drifted 14 commits with overlapping changes across Reports, the worker, admin-api, and tests; same end state, resolved once with full context.

Conflicts resolved to keep both sides' newer behavior:
- decisions read: main's `id DESC` tiebreak plus #221's fetch-one-past-cap truncation detection.
- resolution-label and decisions queries: #221's richer descriptors, and main's two-filter label cleanup on reopen.
- one behavior call worth a look: on a reports-query error I kept main's "hold the last good queue and show a stale warning" over blanking the pane, so a slow poll doesn't read as an empty queue.
- unioned both sides' new banners and tests.

Green locally: frontend tsc/eslint/vitest, worker tsc/vitest, and the D1 suite. CI is green.

Unrelated heads-up: a pre-existing worker test (`feedLabelToEachFilter`) flaked once under parallel runs then passed on re-run; it polls a socket via real `setTimeout`. Not from this change, worth hardening separately.

Ready for re-review.

hm21 (approved):
Approving on `e29c190`.

All four items from my earlier review are resolved, and the merge with main preserved both sides in `Reports.tsx` — the warm-refresh error banner is back alongside the reports-failure-first ordering.

I mutation-tested every new guard. Two survived across the full suite, and I pushed a killing test for each:

**`bc920ea`** — `hasColdFailed`'s `!isPaused` term had no coverage at all: removing it left all 679 tests green. It isn't cosmetic. `errorUpdateCount` only increments, so without that term a source that failed cold and then lost connectivity stays latched in `blockingErrors`, `blockingLoadPaused` is empty, the offline branch is never reached — and the moderator is told the block "usually clears on the next automatic refresh" while there is no connection to clear it, with nothing explaining why Retry does nothing.

**`e29c190`** — `/api/decisions` had cases for 0, 3 and 1500 rows but not exactly 1000, so `>` could regress to `>=` and raise the truncation banner on a window that is actually complete. That boundary is the whole reason the query fetches `DECISIONS_LIMIT + 1`.

Everything else I mutated died to exactly one test: `Math.max`, decisions' `gatesAlways`, `errorUpdateCount` vs live `error`, the per-source override scoping, gating the labels bound on `resolvedFilterActive`, the 8s read timeout, `retry: 1`, `parseOldestCovered`'s `Z` check and its seconds→ms conversion, and both truncation flags. The `TZ` pin earns its place: dropping `timeZone: 'UTC'` now fails three tests instead of none.

Validation on the pushed head: `npm run test` 679 passed / 1 skipped, plus eslint, tsc and build; worker typecheck clean, 636 unit, 34 d1. PR checks green.

Both commits are test-only — no production behaviour changed. Nothing left open from my side. @mbradley over to you; @rabble your review request is still pending and the head has moved.

PR #229, opened 2026-08-08
fix(config): backfill persisted app config from current defaults

rabble:
## Summary

Hardens `AppProvider` against a persisted config that is missing a field, and resolves `relayUrl`/`apiUrl` as a single environment pair so a partial config can never straddle two environments.

Not a fix for #227. Two reviewers independently established the reported state is unreachable — see Correction below — so that link is dropped.

## Correction to the original framing

This branch opened as a repair for configs written before `apiUrl` entered `defaultConfig` (8d162b1, 2026-01-23). Review disproved it:

- The live provider reads `nostr:app-config-v2` (`src/App.tsx:37`), a key introduced in 3d74a98 on 2026-02-06 — two weeks *after* `apiUrl` was already in the defaults.
- There is no migration from the older `nostr:app-config` key, and nothing reads it.
- `useLocalStorage` never writes on init, and every writer spreads the full config, so a `-v2` entry missing `apiUrl` cannot be produced.
- `buildEnvironments()` admits an environment only when both URLs are set, `EnvironmentSelector` and `Reports` write both fields from that list, and the custom relay path validates input. No write path blanks either field.

Credit to @NotThatKindOfDrLiz and @realmeylisdev for both catching this against the v2 key.

## What this is now

Forward-compatibility hardening for the next field added to `AppConfig`, plus two defects the review surfaced in the original approach.

**Environment pairing.** Backfilling the two URLs independently was itself a hazard: a stored staging `relayUrl` with a missing `apiUrl` took the production API. The pane would render staging content while bans executed against production, and the environment chip would read "Custom". `resolveEndpointPair` now takes both halves from the environment matching the stored relay, or both from defaults when it matches none. A complete pair is untouched, custom relays included.

**Malformed JSON re-throws.** On `main`, `deserialize` is `JSON.parse`, so a corrupt cross-tab `storage` event throws, `useLocalStorage` catches it, and the live config is left alone. Returning defaults instead would silently drop a moderator back to production mid-session. Re-throwing preserves that; the init path still reaches `defaultValue` through its own try/catch.

## Approach

Parsing goes through `parseStoredConfig`, passed to `useLocalStorage` via its existing `deserialize` seam, so initial load and cross-tab sync share one path. Stored values win; only missing or empty ones are resolved. Non-object stored values fall back to defaults wholesale. No domains are introduced — fallbacks come from `defaultConfig`.

## Tests

10 tests: complete pairs preserved (known and custom), a missing field filled, the missing half taken from the stored relay's environment, both halves from defaults when the relay is unknown, empty strings, non-object values, and malformed JSON re-thrown.

One renders `AppProvider` with the real `nostr:app-config-v2` key and a pre-seeded store. That covers the wiring, not just the parser — `useLocalStorage`'s third argument is optional, so removing the serializer from `AppProvider` previously left the suite and `tsc` green. Verified: dropping it now fails that test.

Local run on `c6f297b`: `tsc -p tsconfig.app.json --noEmit` clean, `eslint` clean, `vitest run` 546 passed / 1 skipped across 55 files.

## Visual change

None. Behaviour only, and only for a config shape no current write path produces.


src/components/AppProvider.tsx
```diff
@@ -20,8 +21,15 @@ export function AppProvider(props: AppProviderProps) {
...
   } = props;
 
-  // App configuration state with localStorage persistence
-  const [config, setConfig] = useLocalStorage<AppConfig>(storageKey, defaultConfig);
+  // App configuration state with localStorage persistence. The stored copy is
+  // reconciled against defaultConfig on read so a config saved before a field
+  // existed picks it up instead of staying undefined for the life of the browser.
+  const serializer = useMemo(() => ({
+    serialize: JSON.stringify,
+    deserialize: (raw: string) => parseStoredConfig(raw, defaultConfig),
```
NotThatKindOfDrLiz, line 29: This read-time backfill does not appear to fix #227 as described. The live provider is wired from `src/App.tsx` with `storageKey="nostr:app-config-v2"`; git history shows that key was introduced in `3d74a98` on 2026-02-06, and `defaultConfig` already had `apiUrl` at that point. The pre-`apiUrl` configs described in the issue were written under the older `nostr:app-config` key, which the app no longer reads. Please either produce a reproduction/localStorage sample against the v2 key or remove `Closes #227` and reframe this as forward-compatibility hardening.

src/lib/appConfig.ts
```diff
@@ -0,0 +1,36 @@
...
+  } catch {
+    return { ...defaults };
+  }
+
+  return {
+    ...defaults,
+    ...stored,
+    // Empty is as unusable as absent: a failed environment switch can leave
+    // these blank, and blank reads back as "no environment selected".
+    relayUrl: stored.relayUrl || defaults.relayUrl,
```
NotThatKindOfDrLiz, line 33: This fallback can silently synthesize a cross-environment config. For example, a stored staging `relayUrl` with a missing/empty `apiUrl` becomes staging relay plus the production default API. That breaks the environment-pair invariant called out in `EnvironmentSelector` and is risky for moderation actions. If `apiUrl` is absent, resolve the matching environment from the stored relay URL, for example via `getEnvironmentByRelayUrl()`, and only fall back coherently when the relay is unknown.

src/lib/appConfig.ts
```diff
@@ -0,0 +1,36 @@
...
+    }
+    stored = parsed as Partial<AppConfig>;
+  } catch {
+    return { ...defaults };
+  }
+
+  return {
+    ...defaults,
+    ...stored,
+    // Empty is as unusable as absent: a failed environment switch can leave
```
NotThatKindOfDrLiz, line 31: I could not find a current write path that blanks these fields: `EnvironmentSelector` writes configured relay/API pairs, `buildEnvironments()` only admits environments with both URLs, and the custom relay path validates input. The defensive empty-string handling is fine only if it preserves the relay/API pair; as written it contributes to the mismatch risk below.

NotThatKindOfDrLiz (changes requested):
Requesting changes. The implementation is small and CI is green, but the PR should not merge as a fix for #227: the diagnosed stale config shape is tied to `nostr:app-config`, while the live app reads `nostr:app-config-v2`, which was introduced only after `apiUrl` already existed. If this reconciliation stays, it needs to be reframed as forward-compatibility hardening, `Closes #227` should be removed unless there is a real reproduction against the v2 key, and the URL fallback must preserve the relay/API environment pair.

src/lib/appConfig.ts
```diff
@@ -0,0 +1,36 @@
...
+import type { AppConfig } from '@/contexts/AppContext';
+
+/**
+ * Parses a persisted AppConfig, backfilling anything the stored copy is missing.
+ *
+ * `useLocalStorage` only applies its default when nothing is stored at all, so a
+ * config written before a field was introduced keeps that field undefined for as
+ * long as the browser holds it. For `apiUrl` that means every worker request
+ * throws "No relay selected" — permanently, and only for the people who have
+ * used the tool the longest.
```
realmeylisdev, line 13: This state isn't reachable. The key `AppProvider` reads is `nostr:app-config-v2` (`src/App.tsx:37`), and that key landed in 3d74a98 (2026-02-06) — after `apiUrl` entered `defaultConfig` in 8d162b1 (2026-01-23). Checking every commit on every ref that touches `App.tsx`, `-v2` never appears without `apiUrl` in the defaults.

There's no migration from the old key, `useLocalStorage` never writes on init, and every writer spreads the full config, so a `-v2` entry missing `apiUrl` can't be produced. The stale pre-`apiUrl` blob is under `nostr:app-config`, which nothing reads.

If the function stays as hardening, this paragraph should say that rather than describe a live outage.

src/lib/appConfig.ts
```diff
@@ -0,0 +1,36 @@
...
+    return { ...defaults };
+  }
+
+  return {
+    ...defaults,
+    ...stored,
+    // Empty is as unusable as absent: a failed environment switch can leave
+    // these blank, and blank reads back as "no environment selected".
+    relayUrl: stored.relayUrl || defaults.relayUrl,
+    apiUrl: stored.apiUrl || defaults.apiUrl,
```
realmeylisdev, line 34: I couldn't find the failed environment switch this describes. `buildEnvironments()` only pushes an environment when both `VITE_*` URLs are set (`src/lib/environments.ts:21-54`); `EnvironmentSelector.tsx:36` and `Reports.tsx:759` write both fields from that list; `useTheme.ts:14` touches only `theme`. No path persists a blank.

Also worth deciding what should happen if it ever does fire on one field only: a stored staging `relayUrl` with a backfilled prod `apiUrl` is a pair that belongs to no environment. The worker takes its relay from its own `RELAY_URL` binding (`worker/wrangler.prod.toml:79`), so the pane would show staging content while bans executed against prod, and the env chip would read "Custom". Today a blank is a loud "No relay selected" instead.

Either drop the branch, or fall back to the whole environment pair so the two can't drift apart.

src/lib/appConfig.ts
```diff
@@ -0,0 +1,36 @@
...
+  let stored: Partial<AppConfig>;
+
+  try {
+    const parsed: unknown = JSON.parse(raw);
+    if (typeof parsed !== 'object' || parsed === null || Array.isArray(parsed)) {
+      return { ...defaults };
+    }
+    stored = parsed as Partial<AppConfig>;
+  } catch {
+    return { ...defaults };
```
realmeylisdev, line 25: Small behaviour change worth a look. On `main` `deserialize` is `JSON.parse`, so a corrupt cross-tab `storage` event throws, gets caught in `useLocalStorage`'s handler, and the live config is left alone. Returning defaults here means a bad write from another tab silently drops the moderator back to production mid-session.

Re-throwing keeps the old ignore-it behaviour — the init path still falls back to `defaultValue` through its own try/catch, so nothing else changes.

src/lib/appConfig.test.ts
```diff
@@ -0,0 +1,60 @@
...
+import { parseStoredConfig } from './appConfig';
+import type { AppConfig } from '@/contexts/AppContext';
+
+const defaults: AppConfig = {
+  theme: 'light',
+  relayUrl: 'wss://relay.example.com',
+  apiUrl: 'https://api.example.com',
+};
+
+describe('parseStoredConfig', () => {
```
realmeylisdev, line 15: All five cases call `parseStoredConfig` directly, so nothing asserts the serializer is actually wired into `useLocalStorage`. The third argument is optional, so removing it from `AppProvider.tsx:32` leaves the suite and the type-check green — and `src/test/setup.ts` stubs `localStorage.getItem` to `null`, so no existing component test reaches this path either.

One test that renders `AppProvider` with the real `storageKey` and a pre-seeded store would cover the wiring, and would have surfaced the `-v2` mismatch straight away.

realmeylisdev (changes requested):
I went looking for the broken state and couldn't produce it — I think the key rename already fixed this back in February.

`AppProvider` reads `nostr:app-config-v2` (`src/App.tsx:37`). That key arrived in #17 (3d74a98, 2026-02-06), two weeks *after* `apiUrl` joined `defaultConfig` (8d162b1, 2026-01-23). Across every commit on every ref that touches `App.tsx`, `-v2` and `apiUrl` only ever appear together — no build has ever read `-v2` with an `apiUrl`-less default. There is no v1→v2 migration, `useLocalStorage` doesn't write on init, and all three writers (`useTheme`, `EnvironmentSelector:36`, `Reports:759`) spread the whole config. So a `-v2` blob missing `apiUrl` can't exist. The pre-`apiUrl` blob is still parked under `nostr:app-config`, which nothing reads.

Two things in the description follow from that:

- The repro sets `nostr:app-config`. I ran it through `AppProvider` with the real key: `config.apiUrl` is the prod default both before *and* after this change, so the "Before: Failed to load reports" step doesn't happen.
- Nothing writes the backfill back — the repair effect doesn't fire on an `https://` URL — so after a reload `localStorage['nostr:app-config-v2']` still has no `apiUrl`. Healing in memory every load is fine, the described post-state just isn't what occurs.

The change itself isn't harmful; merging defaults on read is a reasonable guard for the next field someone adds. I'd just like the story to match the code before it lands.

**What I'm asking for**

1. Either close #227 as already fixed by the key bump, or re-frame this as hardening for future field additions and point the repro at `nostr:app-config-v2`.
2. One test through `AppProvider` with the real storage key. All five tests call `parseStoredConfig` directly, and `src/test/setup.ts` stubs `localStorage.getItem` to `null`, so nothing in the suite reaches the deserializer — delete the `serializer` argument on `AppProvider.tsx:32` and everything still passes.
3. The empty-string branch needs a real justification or removal — inline note below.

**Checks run locally:** `npx tsc -p tsconfig.app.json --noEmit` clean, `npm run test` 541 passed / 1 skipped across 55 files, `vite build` clean. Nothing here touches the worker or the relay protocol, so no backend verification applies.

realmeylisdev (approved):
All four addressed, and the pairing fix goes further than what I asked for — I'd only said don't let the two halves drift; resolving through `getEnvironmentByRelayUrl` and falling back to the whole default pair closes it properly.

Verified on `c6f297b`:

- **Pairing** — a complete pair passes through untouched (custom relays included), a stored staging relay with no `apiUrl` picks up the staging worker, an unknown relay takes both halves from defaults. No input produces a mixed pair. `.env.production` sets both prod and staging, so `getEnvironmentByRelayUrl` is live in a deployed build rather than a no-op.
- **Malformed JSON re-throws** — a corrupt cross-tab write is ignored again instead of resetting the session to production.
- **The wiring test bites.** I dropped the `serializer` argument from `AppProvider.tsx:32`: that test fails on the `api` assertion while the other nine and `tsc` stay green. That's exactly the check that was missing.
- **Framing** — doc comments and the PR body describe hardening now, not an outage.

Checks I ran locally on a clean `npm ci`: `tsc -p tsconfig.app.json --noEmit` clean, `eslint` clean, `vitest run` 546 passed / 1 skipped across 55 files, `vite build` clean. Matches your numbers.

One non-blocking observation, take it or leave it: `resolveEndpointPair` only looks up by relay, so a config carrying an `apiUrl` and no `relayUrl` discards the stored API and takes both halves from defaults. `getEnvironmentByApiUrl()` already exists if you ever want it symmetric. Safe either way — it can't straddle, and neither shape is producible today.

Agreed on leaving #227 open. If a moderator really hit that error the cause is somewhere else, and the issue is the right place to keep looking.

PR #230, opened 2026-08-08
fix(moderation): stop reporting unverified actions as confirmed

rabble:
## Summary

Post-action verification no longer reports success it did not observe. The failure paths return a distinct "could not confirm" outcome instead of "confirmed".

Closes #228.

## Motivation

Every way the check itself could fail was reported as a completed moderation action:

- `verifyEventDeleted` resolved `true` on its 5s timeout (`// Timeout = probably deleted`), `true` in `ws.onerror` (`// Error = assume deleted`), and `true` in the outer `catch`.
- `verifyPubkeyBanned` and `verifyPubkeyUnbanned` returned `true` from their `catch`, commented `// If we can't check, assume it worked`.

A relay that could not be reached therefore confirmed every ban and deletion — the one case verification exists to catch. The path that actually let unremoved content read as handled is `useModerationStatus`: a failed check resolved `true` into `wsResult.userBanned`, so `ReportDetail`'s `isResolved` flipped on an unreachable relay. `verifyModerationAction.allSuccessful` was computed from the same value and is corrected here too, but nothing in the repo calls it, so it was not a live path.

## Approach

The three verifiers return `VerificationOutcome = boolean | null`, where `null` means the check could not be completed. It is deliberately distinct from `false` ("checked, and it did not apply"). `allSuccessful` now requires an explicit `true`.

Call sites compare against `true` rather than relying on truthiness. Their existing message ternaries already route a falsy result to the warning branch, so an unreachable relay now reads as a warning instead of a confirmation. `result: { verified }` in the debug action log keeps the raw outcome, so `null` stays visible there.

## Manual validation plan

With devtools offline throttling, or pointed at an unreachable relay:

- Ban a pubkey, then run verification. Before: "Ban verified - user is in banned list". After: the warning branch, and the action is not counted as successful.
- Delete an event with the relay unreachable. Before: "Event deletion verified". After: the warning branch.
- With the relay reachable, confirm the positive paths are unchanged: a real ban still verifies, and an event that is genuinely still present still reports still-present.

## Checks run

CI green on this PR: `test` (clean install, then `tsc --noEmit`, `eslint`, `vitest run`, `vite build`) plus worker typecheck, worker tests, and worker d1 tests. `semantic-pull-request` and CLA also pass.

Adds tests covering both the positive paths and every fail-closed path. Confirmed the 5 fail-closed assertions are red against the previous implementation while the 5 positive-path assertions stay green, so the tests discriminate rather than merely passing.

Correction to an earlier revision of this description: it claimed a local baseline of 5 type errors and 11 failing test files "identical to `main`". That was an artifact of a stale local `node_modules` — CI does a clean install and root vitest excludes `worker/**`. `main` is clean and so is this branch. Disregard those numbers.

## Visual change

No layout change. The rendered text changes only in the failure case, where an unreachable relay now shows the existing warning copy instead of the existing confirmation copy.

## Follow-up, deliberately not in this PR

That warning copy still says "may not be banned" when the accurate statement is "could not check". Distinguishing the two means rewriting message ternaries across twelve components, and #218, #219 and #226 are open in exactly those files. Noted in #228 rather than bundled here.


hm21 (approved):
Approving. The fix is right, and I checked it discriminates rather than merely passing: reverting `adminApi.ts` to `main` turns exactly the five fail-closed assertions red while the five positive-path ones stay green.

Worth noting the blast radius is wider than the description claims, in your favour. `verifyModerationAction.allSuccessful` has no consumer in the repo — the path that actually let unremoved content leave the queue is `useModerationStatus`, whose fields were already typed `boolean | null`. A failed check used to resolve `true` there, so `ReportDetail`'s `isResolved` flipped on an unreachable relay. That now falls back to the ban list instead.

Pushed two commits, each revertable on its own:

**9581255** — folded the new cases into `src/lib/adminApi.test.ts`. They cover `adminApi.ts`, and that file already holds a `verifyAgeRestricted` block, so one module had tests in two files with two mocking setups. The move also picks up that file's fake-timer pattern instead of sleeping through the real 500ms/1000ms/5000ms waits — 12.5s down to under a millisecond, the timeout case alone having been 6s. Re-checked the five fail-closed assertions still go red against pre-fix `adminApi.ts` after the move.

**5443f4f** — `EventDetail.handleReVerify` was the one place the unknown outcome still printed unhedged copy: `null` rendered "User is NOT in banned list" / "Event is still accessible on relay", asserting something nothing observed. Both branches now route `null` to the string the `catch` block already used, so no new copy. This is a deliberate narrow deviation from your #228 deferral — that deferral covers the hedged "may not be" strings across twelve components, which I left alone; these two claimed certainty. Added `EventDetail.reverify.test.tsx` pinning all three outcomes; the inconclusive case fails against 8171011.

Validation: `tsc --noEmit`, `eslint .`, full `vitest run` (55 files, 549 passed), `vite build`, and CI green on the pushed head.

One thing I did not touch and you may want to: `useModerationStatus.runCheck` now stores `null` for both "not checked" and "checked, could not confirm", and `isUserBanned: wsResult.userBanned ?? isUserBannedFromList` silently prefers the stale list in the second case. Correct today, but it collapses a distinction this PR just established. Left it for you rather than guessing at the intent.

src/hooks/useModerationStatus.ts
```diff
@@ -146,18 +155,30 @@ export function useModerationStatus(
...
 
   return {
-    // User ban: WebSocket check result takes precedence over ban list
-    isUserBanned: wsResult.userBanned ?? isUserBannedFromList,
+    // User ban: a completed check wins over the ban list. When the check could
+    // not answer, the list may still settle it — but only positively. A list
+    // `false` is untrustworthy here, because a failed list query resolves to
+    // `[]` above and so looks identical to "not in the list". A list `true`
+    // cannot be produced that way: membership is evidence either way.
+    isUserBanned: wsResult.userBanned === undefined
```
dcadenas, line 168: This is the split that made the rest work.

`undefined` now means no check has run, and `null` means a check ran and could not answer. Only the first consults the ban lists, because a failed list query resolves to `[]` a few lines up — so a list "absent" is indistinguishable from "the query died".

The carve-out matters in the other direction. That same failure mode cannot invent a *membership*, so a list `true` is still good evidence and is kept. Dropping it cost the banned badge, `isResolved`, and the Unban affordance whenever a fresh check hiccupped on an already-banned pubkey.

Worth knowing for the pubkey case specifically: `verifyPubkeyBanned` calls `listBannedPubkeys` itself, so this is a failed fresh call falling back to a cached one within `staleTime`, not two independent sources. The argument still holds, but it is a cache, and an unban followed by a failed verify can briefly re-assert "banned" before the refetch settles it.

src/components/ThreadContext.tsx
```diff
@@ -300,32 +300,50 @@ export function ThreadContext({
...
 
+    // Each half of the summary line is stated only for the outcomes it was
+    // actually told. A check that could not answer says so rather than
+    // borrowing the negative wording.
+    // `false` here is not "the normal queries missed it" — it is the direct
+    // verification REQ getting the event back (adminApi.ts resolves false only
+    // on an EVENT match). Reporting that as "not found" states the inverse of
+    // what the check observed.
+    const eventLine = isEventDeleted === false
+      ? 'Event is still on the relay.'
```
dcadenas, line 311: `verifyEventDeleted` resolves `false` only inside the `EVENT` match for the requested id, and `isEventGone` cannot reach `false` any other way — the pre-check branch yields `true` or `null`. So this state means the verification request got the event back, and the old "Event not found on relay." stated the inverse of the observation.

It is also the more useful sentence here: this block only renders when the normal queries missed the event, so a direct request finding it is a real discrepancy rather than something to flatten.

That wording predates this PR. It only became visible once the outcomes were separated.

src/components/ReportDetail.tsx
```diff
@@ -835,20 +835,29 @@ export function ReportDetail({ report, allReportsForTarget, allReports = [], onD
...
                     User is not banned
                   </span>
                 </div>
+              ) : moderationStatus.checkedAt ? (
+                // The check ran and could not answer. Saying "not banned" here
+                // would state a fact nothing observed.
+                <div className="flex items-center gap-2 p-2 rounded bg-muted">
+                  <RefreshCw className="h-4 w-4 text-muted-foreground shrink-0" />
+                  <span className="text-sm font-medium text-muted-foreground">
+                    Could not check ban status
```
dcadenas, line 858: This branch is the one that motivated the rest.

The panel previously rendered "User is not banned" whenever a check had run and the value was falsy, alongside a "Checked:" timestamp. Once a failed check resolved `null` instead of the old fail-open `true`, that became a timestamped claim about a relay nothing had reached.

The negative branch keeps its `checkedAt` guard, so nothing that was reachable before renders differently.

dcadenas (approved):
✅ The fail-closed change is right, and the three-state outcome is the correct shape for it.

The verifiers no longer report success they did not observe. What the new `null` exposed
is that several consumers stated the *opposite* outcome just as confidently, so I pushed
fixes for those rather than hand them back.

### Changes I pushed

- An inconclusive check no longer falls through to the ban list for a negative answer, so
  an unreachable relay stops reading as "not banned".
- It still keeps a positive ban-list membership, which a failed list query cannot fabricate.
- The report panel and the thread summary gained an explicit "could not check" state
  instead of borrowing the negative wording.
- The post-action ban, delete, and unban results route the unknown outcome to the existing
  inconclusive message, so none of them cite a banned list they never read.
- A verified-present event now reads as present. That copy said "not found" for the one
  state where the check had actually received the event back.

Each is in its own commit. The widest one is the thread summary: its line is now composed
from each half separately, so a known outcome and an unchecked one can appear together.
That changes wording in states that were previously reachable, and it is isolated in
`d144c15`.

### Checks

- `tsc -p tsconfig.app.json --noEmit`
- `eslint`
- `vitest run`: 58 files, 571 passed, 1 skipped
- `vite build`
- Same four on a throwaway merge with current `main`, and on CI at the pushed head
- Every new assertion was run against its own parent commit and fails there

### Record correction

The description attributed the queue-hiding consequence to
`verifyModerationAction.allSuccessful`. That function has no caller, so it was not a live
path — the behavior came through `useModerationStatus` into `isResolved`. I corrected that
paragraph in the description; the fix itself was never in question.

### Follow-ups

- `Closes #228` will close that issue on merge, but #228's own follow-up paragraph — the
  hedged copy across the remaining components — is still open. Worth changing to a
  non-closing reference, or moving the follow-up somewhere it survives. Your call, since
  the deferral is yours.
- The remaining hedged wording in `EventModeration`, `EventsList`, `Labels`, and
  `DebugPanel` is untouched, as you scoped it.

Neither blocks this branch.

PR #231, opened 2026-08-08
chore(repo): remove stale worker public artifacts

NotThatKindOfDrLiz:
## Summary
- remove stale checked-in frontend build output from worker/public and delete the dead GitLab Pages config
- document dist/ as the single Cloudflare Pages artifact path and ignore any future worker/public output
- tighten per-test mock isolation in two worker suites without changing behavior
- finish stale worker config, relay host, and API host cleanup in touched docs/scripts

Closes #53.
Refs #54.

## Validation
- npm run test
- cd worker && npm run test:run
- cd worker && npm run typecheck
- npm run build
- confirmed dist/ contains index.html, 404.html, manifest.json, and _redirects

## Notes
- Wrangler Cloudflare-side preflight could not run in this non-interactive shell because CLOUDFLARE_API_TOKEN is not set.


hm21 (approved):
LGTM. Verified `worker/public/` is dead: no `[assets]`/`[site]` binding in any of the three worker wrangler configs, no `ASSETS` reference in `worker/src`, and nothing in CI or docs points at it. `.gitlab-ci.yml` is mkstack template residue — the only GitLab reference left is `stack.json`'s upstream URL. Built locally and confirmed `dist/` has `index.html`, `404.html`, `manifest.json`, `_redirects`. The `tsc -p tsconfig.app.json` README fix is a real catch — root `tsconfig.json` is `"files": []` plus references, so the old command type-checked nothing.

One note on the worker test-mock changes: I could not reproduce a failure they fix. `main` is green at the lockfile's vitest 4.0.18, at 4.1.10 (newest in the `^4.0.18` range, in case resolver drift was the trigger), and across three `--sequence.shuffle` runs. The existing comment at `worker/src/bulk-moderate.test.ts:528` ("call history accumulates across tests") also suggests `vi.restoreAllMocks()` isn't wiping those module mocks at this version. They're strictly more robust so I left them in — but if you hit a real failure somewhere, worth capturing where, since it'd point at something the pinned config doesn't cover.

Unrequested change, pushed as its own commit (62d3cc8) so you can drop it: `DEPLOYMENT.md` told readers to configure worker vars in `worker/wrangler.toml`, which doesn't exist — retargeted at the three per-environment configs. Same stale reference also lives in `.env.example` and `docs/api-webhooks.md`; left those for a separate pass.

worker/src/bulk-moderate.test.ts
```diff
@@ -210,6 +211,8 @@ describe('runBulkModeration', () => {
 
   beforeEach(() => {
     vi.restoreAllMocks();
+    vi.mocked(getAdminPubkey).mockResolvedValue('moderator-pubkey');
+    vi.mocked(banEvent).mockResolvedValue({ success: true });
```
realmeylisdev, line 215: These two lines are a no-op, and the reason given for them ("stabilize existing worker test mocks so the required worker suite passes") doesn't hold — the suite already passes without them.

On vitest 4.0.18 `vi.restoreAllMocks()` only restores `vi.spyOn` spies; it does not clear implementations set on `vi.fn()`, whether module-scope or returned from a `vi.mock` factory. I checked with a probe test rather than going from the changelog. Concretely:

- Checked out main's version of this file and `age-review.test.ts` and ran the worker suite: 411/411 pass.
- main's CI has been green on the last 8 runs of the test workflow.
- Line 531 in this same file already carries the comment "call history accumulates across tests" next to an explicit `mockClear()` — which is the same conclusion from the other direction.

So: harmless, and mildly useful as a guard if someone later adds a non-`Once` override above these tests. But it's unrelated to removing `worker/public`, and the PR body describes it as a fix for a failing suite that wasn't failing. Either drop the two test files from this PR or reword that bullet — your call, not blocking.

worker/src/age-review.test.ts
```diff
@@ -2071,17 +2071,19 @@ describe('handleGetAgeReviewFunnel', () => {
     { state: 'denied_closed', created_via: 'report', c: 1 },
     { state: 'submitted_for_review', created_via: 'report', c: 4 },
   ];
-  const mockDb = {
-    prepare: vi.fn().mockReturnValue({
-      bind: vi.fn().mockReturnValue({ all: vi.fn().mockResolvedValue({ results: groupRows }) }),
-    }),
+  const makeMockDb = () => {
```
realmeylisdev: Same as the note on `bulk-moderate.test.ts`: nothing was resetting the shared `mockDb`, so this isn't fixing a failure. It does buy genuine per-test isolation of call history, which is a fine thing to have — it just isn't what the PR body claims.

While you're here, the factory body can be a concise arrow (`const makeMockDb = () => ({ prepare: ... })`) — the `return` wrapper isn't doing anything.

realmeylisdev (approved):
Approving. The removal is safe and I verified it end-to-end rather than taking the premise on trust. I also pushed four doc commits for gaps the accuracy pass left behind — details below.

## Why the deletion is safe

- No worker config binds static assets: no `[site]`, no `assets`, no `bucket` in `wrangler.local/staging/prod.toml`.
- No static serving in the worker either — no `ASSETS`, `__STATIC_*`, or `getAssetFromKV` anywhere in `worker/src/`. It really is API-only.
- Nothing references the path. After this PR the only mention of `worker/public` in the tree is the new `.gitignore` line.
- Nothing regenerated it. `worker/public` came in with the original UI commit and was last hand-updated in fe88582; the deleted GitLab job built root `public/` (`mv dist public`), not `worker/public/`. GitLab is dead here anyway — the repo runs on Actions, and the only remaining `gitlab` strings are the mkstack template origin in `stack.json`.
- Pages still deploys `dist` via the root `wrangler.toml` (`pages_build_output_dir = "dist"`).

I also confirmed the new DEPLOYMENT.md paragraph is true: `npm run build` emits `dist/` with `404.html`, `_redirects`, `index.html`, `manifest.json`. (Note `npm run test` ends in a bare `vite build`, so *that* dist has no `404.html` — harmless, nothing deploys from it.)

## What I pushed (4 commits, one per finding)

The doc pass fixed one instance of "there is no `worker/wrangler.toml`" and left four more:

1. **3973c91** — the Environment Configurations section called production `relay.dvines.org` and staging `relay.divine.video`, and labelled both blocks `wrangler.toml`. Per the configs, prod is `relay.divine.video`, staging is `relay.staging.divine.video`, and `MANAGEMENT_PATH = "/"`. Also fixed the `RELAY_URL` / `MODERATION_SERVICE_URL` examples in the table right under your edit.
2. **29fff44** — `cd worker && wrangler deploy` cannot work with no default config. Deploy, secret, and dev examples now pass `--config`, per AGENTS.md.
3. **87db289** — `.env.example` and `docs/api-webhooks.md` still pointed at `worker/wrangler.toml`.
4. **6933139** — the 404 troubleshooting entry said "Funnelcake uses `/management`". It doesn't: NIP-86 serves the management API on the same URI as the relay websocket, and funnelcake dispatches on `path == "/"` (`crates/relay/src/websocket.rs:506`). Both deployed configs already set `"/"`, so that entry was pointing anyone debugging a 404 at the exact value that causes one.

## Validation

`npm run test` (536 passed, 1 skipped, build clean) · worker `typecheck` · worker `test:run` (411) · worker `test:d1` (21). All green locally.

## Two things left for you

- **The test-mock commit** — see the inline note. Not a problem, but the stated reason doesn't hold up and it's unrelated to this PR.
- **DEPLOYMENT.md is stale beyond what I touched.** `Verifying Deployment` curls `https://relay.admin.divine.video/api/info`, but that host is the Pages frontend origin — the worker routes are `api-relay-{prod,staging}.divine.video`. I left it alone because a dashboard-side route could exist that isn't in the wrangler config; you'd know. Separately, `MANAGEMENT_PATH`'s built-in default in `nip86.ts:70` is `/management`, which 404s against Funnelcake — it only works today because both configs override it. Both feel like a follow-up issue rather than scope creep here.

PR #232, opened 2026-08-12
feat(moderate-media): let a caller declare the state it expects to change

mbradley:
## What is wrong today

This endpoint sets the moderation state of a piece of media. It does not record what you meant by the change.

So "make this visible again" and "clear the hold on a minor's video" are the same request. There is no way for the endpoint to tell them apart, because both arrive as `action: SAFE` and nothing else.

That matters because of how age review works. When we hold a minor's video, we hide it with `QUARANTINE` and later clear it with `SAFE`. COOP's "Un-Restrict-Media" button also sends `SAFE`. A moderator clicking that button on a reported video could therefore clear an age-review hold without meaning to, and without anything warning them.

We already refuse this kind of accidental reversal for accounts. The existing guard covers un-banning and un-suspending a person. It works off the account key, and this endpoint works off a file hash, so it never applied here.

## What this adds

A caller can now say what state it expects to be changing. If the media is in some other state, the request is refused and nothing happens.

The COOP un-restrict button says it expects to be undoing an age restriction. If the video is actually on an age-review hold, the request is refused, the hold stays, and the response says both what the caller expected and what is really there.

Refusals are:

| Status | Meaning |
|--------|---------|
| 409 | The media is in a different state than you said. Says which. |
| 503 | We could not read the current state. Try again. |
| 400 | The request itself was malformed. |

The 503 is deliberate. If we cannot check, we refuse rather than proceed, because "I could not look" must never come back as "done".

## What this does not change

Nothing, for anyone not using it. The field is optional. Every caller that exists today omits it, behaves exactly as before, and does not pay for the extra lookup.

## What this does not do

It protects a caller from its own mistake. It is not a security control. The caller supplies the expected state, and leaving it out gets the old behaviour, so it does not stop anyone determined to skip it. The endpoint is already behind admin authentication, and an honest caller getting it wrong is the real risk here.

There is a small gap between reading the current state and changing it. If something else changes the media in that instant, we will not catch it. Closing that properly needs a change in moderation-service, which is not in this PR. The gap is one round trip, and this is still better than today, where there is no check at all.

What we compare against is moderation-service's record. Actions taken directly in Blossom's own admin screens do not write there, so those are not visible to this check. The age-review path that motivates the change does write there, so that one is covered.

## What review turned up

The first version of this had four real problems, three of which passed all tests. Worth listing, because they are the reason this took more than one pass:

**It could check one record and change a different one.** Hashes are matched case-sensitively when written but lowercased when read. So a hash sent in capitals looked at one record and wrote to another. In practice: a hold placed using a capitalised hash would leave the original record still saying "age restricted", so the un-restrict button would check that stale record, see what it expected, and lift the hold anyway. The file would become visible again while our age-review screens kept showing it as hidden. Two reviewers found this separately.

**An unchecked hash could redirect the lookup.** A caller sending path characters instead of a hash could point the state lookup at a different internal endpoint, and that endpoint's response could satisfy the check. It required admin access already, so this was not an open door, but it meant a caller could be told its state was verified when nothing relevant had been read.

**An empty expected-state was ignored instead of refused.** If a caller sent the field but it came out blank, the check was skipped silently and the caller got a success. That is the worst possible outcome: it asked for the protection and was told it got it.

**A malformed response was read as "nothing has been moderated".** That reported a fact about the file that had never been established. It now reports that the state could not be read, which is what actually happened.

## How it was checked

Fifteen tests. Each was verified by breaking the code on purpose and confirming a test caught it, fifteen different ways. That is how three of the four problems above were found: the tests passed, but they would still have passed with the protection removed.

Also run against the local stack end to end:

```
put a video on an age-review hold           -> held
click un-restrict (expects age-restricted)  -> refused, 409, both states named
check the video afterwards                  -> still held
```

504 tests, type checks and linting clean.

## Sequencing

`divinevideo/coop-webhook-adapter#8` is the first caller and should land after this. Against an older relay-manager it would simply not get the protection, rather than break.

### DIFF
--- a/docs/api-webhooks.md
+++ b/docs/api-webhooks.md
@@ -129,7 +129,7 @@ validated upstream.
 
 | Status | `code` | Meaning |
 |--------|--------|---------|
-| 400 | `invalid_from` | `from` was present but empty, whitespace, or not a string. Omit it to skip the check. |
+| 400 | `invalid_from` | `from` was present but empty, whitespace, null, or not a string. Omit it to skip the check. |
 | 400 | `invalid_sha256` | `sha256` is not 64 hex characters. |
 | 409 | `state_mismatch` | Current state is not the declared one. The body carries `from` and `current`. |
 | 503 | `state_unreadable` | Current state could not be read. Retryable, and deliberately not a success. |
--- a/docs/api-webhooks.md
+++ b/docs/api-webhooks.md
@@ -100,9 +100,46 @@ Moderate media content (images/videos) by SHA-256 hash.
 {
   "sha256": "abc123...",
   "action": "SAFE" | "REVIEW" | "QUARANTINE" | "AGE_RESTRICTED" | "PERMANENT_BAN" | "DELETE",
-  "reason": "CSAM content"
+  "reason": "CSAM content",
+  "from": "AGE_RESTRICTED"
 }
 ```
+
+**`from` (optional): declare the state you expect to be changing.**
+
+This endpoint sets state; it does not describe a transition. `SAFE` means "make
+this Active", not "undo the age-restriction I applied", so a caller reversing its
+own action and a caller clearing an age-review quarantine on a minor's content
+send the identical request.
+
+`from` lets a caller declare the state it believes it is changing. The current
+state is read first and the action is refused unless it matches, so a button can
+only reverse the thing it is named for. Omit `from` and nothing changes: no
+extra read, no new failure mode.
+
+It guards a cooperating caller against its own bug. It is not a security
+boundary, since the caller supplies the value and omitting it restores the
+unguarded path.
+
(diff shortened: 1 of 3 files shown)


dcadenas (approved):
The expected-state check matches the claim: omit `from` and nothing changes; send it and a mismatch or unreadable state refuses instead of writing.

### Changes I pushed

- A non-string `sha256` on this path is now 400 instead of a 500 TypeError.
- JSON `null` for `from` is now 400 `invalid_from`, same as an empty string.

### Checks

- Worker typecheck
- Worker `vitest run`: 504 passed
- GitHub `test` and `semantic-pull-request` are green on 8d879b2

### Follow-ups

- The check and the write are still two requests. A real compare-and-set needs a change in moderation-service. That is outside this branch.

PR #233, opened 2026-08-12
fix(local-dev): stop local dev writing to production

mbradley:
## The problem

Running relay-manager locally wrote to production.

`wrangler.local.toml` pointed three outbound URLs at deployed infrastructure:

```
RELAY_URL              = "wss://relay.divine.video"
MODERATION_SERVICE_URL = "https://moderation-api.divine.video"
MODERATION_ADMIN_URL   = "https://moderation-api.divine.video"
```

Those are not decorative. The worker proxies moderation actions to `MODERATION_ADMIN_URL`, and it signs and publishes events to `RELAY_URL`. So starting the worker the way the file's own header tells you to, opening the admin UI, and clicking a moderation button wrote a real production decision.

### How exposed was it, precisely

Narrower than "everyone, all the time", and worth stating accurately.

`worker/.dev.vars` overrides `[vars]`, and on a configured machine it already sets `RELAY_URL`, `FUNNELCAKE_API_URL` and `MODERATION_ADMIN_URL` to local values. I confirmed that by asking the running worker, which reports `ws://localhost:4444`. So a developer whose local state is fully set up was not publishing to the production relay.

Two gaps remain, and they are the reason this is worth fixing rather than leaving to convention:

- **`MODERATION_SERVICE_URL` is not in `.dev.vars`.** It fell through to the production value even on a fully configured machine. That is the URL the expected-state read in #232 uses.
- **`.dev.vars` is gitignored.** A fresh clone gets none of it. Neither does a git worktree, which is how a lot of work in this repo happens, and which is exactly where someone would start the stack without noticing.

The committed default was a trap for anyone whose untracked local state did not happen to paper over it. Putting the right values in the tracked file removes the dependence on each developer's private file.

## The change

All three now point at the local stack, which I verified is actually running rather than assuming:

- Local funnelcake answers NIP-11 as `divine-local` on 4444
- A local moderation-service answers `/health` as `divine-moderation-api` on 8789

`127.0.0.1` rather than `localhost`, because `localhost` can resolve to `::1` first while `wrangler dev` binds v4.

`REALNESS_API_URL` still names the deployed service. There is nothing local to run in its place. `handleRealnessViaHTTP` POSTs to `/analyze` with CF Access credentials, so exercising that path creates real jobs. That is accepted risk, not an absence of risk.

`FUNNELCAKE_API_URL` is set explicitly to `http://127.0.0.1:3333`. Local funnelcake serves the relay and the REST API on different ports (4444 and 3333). Deriving the API URL from `RELAY_URL` would send `/api/funnelcake/*` at the relay port.

## A second bug this uncovered

Pointing `RELAY_URL` at the local relay did not work on its own, because `getManagementUrl` forced https for both websocket schemes:

```ts
env.RELAY_URL.replace(/^wss?:\/\//, 'https://')
```

So `ws://` was called over TLS. Against a local relay serving plain HTTP that fails on the handshake, taking every NIP-86 management request with it: ban, delete, allow. Without fixing it, this PR would have swapped one broken local setup for another.

The existing test locked the behaviour in while contradicting its own name. It was called `should handle WS (non-secure) URLs` and asserted the result was `https`. It now asserts `http`, with a companion test so the secure scheme cannot be downgraded by a later edit to the same line.

`deriveFunnelcakeApiUrl` already maps the two schemes separately from this same value, so the two helpers now agree rather than disagreeing.

Deployed configs are unaffected. `wrangler.staging.toml` and `wrangler.prod.toml` both set `RELAY_URL` to `wss://`, and that path is unchanged.

## The guard

A test that fails if local config points anywhere deployed.

It is inverted on purpose. Instead of listing the variables that must be local, it finds **every** variable whose value looks like a URL and requires each one to be either local or explicitly excused with a written reason. A new outbound URL added later fails the test until someone decides which it is, rather than silently inheriting a production default. That is the failure mode this PR is fixing, so the guard should not be able to miss the next instance of it.

A third check keeps the excuse list honest: every excused variable must still exist and must carry a real reason, so the list cannot rot into a blanket exemption.

## How it was checked

The guard was mutation-checked both ways: reverting a URL to production fails it, and adding a new unclassified URL variable fails it.

The config change was verified end to end, not by reading it. I started a worker on a spare port from the committed config with **no** URL overrides, and asked it about a hash that exists only in the local database:

```
hash dd44...be8b6

  local worker (committed config, no overrides)  -> age_restricted
  production                                     -> unknown
```

It reaches the local service on the strength of the file alone. Used a read-only route for this deliberately, so nothing was written anywhere during verification.

507 tests, type checks and linting clean.

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -252,6 +252,26 @@ This starts:
 - **Caddy HTTPS proxy** on `https://localhost:8788` (terminates TLS, forwards to worker)
 - **Vite frontend** on `https://localhost:5173`
 
+### Backing services
+
+`dev-local.sh` does **not** start these, but the worker calls them. It warns at
+startup for any that are not answering, and continues.
+
+| Service | Where | How to start |
+|---------|-------|--------------|
+| Funnelcake relay | `ws://127.0.0.1:4444` | `cd ~/code/divine-relay-test && ./scripts/relays/setup-funnelcake.sh` |
+| Funnelcake REST API | `http://127.0.0.1:3333` | Same cluster as the relay |
+| moderation-service | `http://127.0.0.1:8789` | `cd ~/code/divine-moderation-service && npx wrangler dev --port 8789 --local` |
+
+Without them, moderation actions and relay reads fail. They do **not** fall
+through to production.
+
+That is a deliberate change. `wrangler.local.toml` used to point these at the
+deployed services, so local dev appeared to work with none of the above running,
+by writing real production moderation decisions and publishing to the production
+relay. `worker/test/local-config-targets-local-stack.test.ts` now fails if any
+local URL points at deployed infrastructure.
+
 Select "Local" in the environment selector. The frontend sends `X-Admin-Key` header (from `VITE_ADMIN_API_KEY`) for admin auth since CF Access isn't available locally.
 
 ### Why HTTPS locally
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -266,11 +266,16 @@ startup for any that are not answering, and continues.
 Without them, moderation actions and relay reads fail. They do **not** fall
 through to production.
 
-That is a deliberate change. `wrangler.local.toml` used to point these at the
-deployed services, so local dev appeared to work with none of the above running,
-by writing real production moderation decisions and publishing to the production
-relay. `worker/test/local-config-targets-local-stack.test.ts` now fails if any
-local URL points at deployed infrastructure.
+That is a deliberate change. The committed `wrangler.local.toml` used to name
+the deployed services. A machine with a populated `worker/.dev.vars` already
+overrode `RELAY_URL`, `FUNNELCAKE_API_URL` and `MODERATION_ADMIN_URL`, but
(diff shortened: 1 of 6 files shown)


dcadenas (approved):
Local defaults no longer depend on an untracked `.dev.vars` file to stay off production.

### Changes I pushed

- Corrected leftover comments that still claimed every local run wrote to production.
- Named `CDN_DOMAIN` in the local-stack guard as accepted risk (no local blossom).
- Pointed the Funnelcake API URL at kind port 3333. 3000 on that cluster is Grafana.

Pairing the kind relay on 4444 with native funnel on 3000 mixes two stacks. Native funnelcake uses 3000 with relay 7777.

### Checks

- worker `npm run typecheck`
- worker `npm run test:run`: 494 passed
- frontend `npx tsc -p tsconfig.app.json --noEmit`
- GitHub `test` and semantic-title green on this head

Not run here: D1 suite (no schema change), e2e (needs a local relay). Root `npm run test` is a multi-step script; CI ran the equivalent.

### Follow-up

@mbradley the branch is one commit behind `main` (`#232`, expected-state). No overlapping files. Leaving the base sync to you.

PR #235, opened 2026-08-13
fix(settings): tolerate object-shaped listallowedkinds from the relay

mbradley:
## Summary
Stops the Settings tab from crashing the admin UI when the relay returns its allowed event kinds as objects rather than bare numbers. The dashboard now normalizes both shapes before rendering, so an object can never reach JSX. The same normalizer is also wired on RelaySettings and RelayStats, which are not currently mounted.

## Motivation
The Settings dashboard rendered the relay's `listallowedkinds` response assuming `number[]`. Funnelcake returns `Array<{kind, added_at}>`, so rendering an entry threw React #31 and the app-level error boundary took over the whole UI. The render was previously reached only by signed-in moderators; #218 dropped the sign-in gate on this query, so the common signed-out case now hits it reliably. This also closes the latent signed-in crash on current prod.

## Related Issue
- Closes #234

## Validation
- [x] `npx tsc -p tsconfig.app.json --noEmit`
- [x] `npx vitest run src/lib/allowedKinds.test.ts` (5 passing; mutation-verified the object-shape assertion fails on the old behavior)
- [x] `npx vitest run src/components/relay-management-access.test.tsx` (includes a Settings render of the Funnelcake object shape)
- [x] Re-smoked on staging: Settings renders signed-out, no crash, all 40 kind badges correct (`Kind 0 — Metadata` … `Kind 1984 — Reporting`)

## Manual Test Plan / Notes
Deployed the branch to the staging Pages alias against the staging worker; opened Settings while signed out. Before: full-page "Something went wrong". After: Event Kind Configuration renders normally.

## Visuals / API Examples
Visual change (crash → working Settings tab), verified in-browser on staging; screenshots omitted by request.

## Public Artifact Review
- [x] Titles, descriptions, branch names, and screenshots avoid partner, customer, brand, campaign, or other sensitive external names.

### DIFF
--- a/src/components/RelaySettings.tsx
+++ b/src/components/RelaySettings.tsx
@@ -13,6 +13,7 @@ import { Tabs, TabsContent, TabsList, TabsTrigger } from "@/components/ui/tabs";
 import { Dialog, DialogContent, DialogDescription, DialogFooter, DialogHeader, DialogTitle, DialogTrigger } from "@/components/ui/dialog";
 import { useToast } from "@/hooks/useToast";
 import { Globe, FileText, Image, Plus, Trash2, Shield, Network } from "lucide-react";
+import { normalizeAllowedKinds, type AllowedKindEntry } from "@/lib/allowedKinds";
 
 interface RelaySettingsProps {
   relayUrl: string;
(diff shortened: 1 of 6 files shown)


dcadenas (approved):
Object-shaped `listallowedkinds` entries no longer reach the Settings kind badges.

The original helper and Settings `select` were the right fix. I pushed the same normalizer onto the other two kind-badge queries so they cannot regress if they are wired up again.

### Changes I pushed

- Applied the same `select` normalizer on RelaySettings and RelayStats.
- Drop non-integer kind values, not only non-finite ones.
- Added a Settings render test for the Funnelcake `{kind, added_at}` shape.

### Checks

- `npx eslint`
- `npx vitest run`: 63 files, 609 passed, 1 skipped
- CI Test and semantic-pull-request are green on `7b2127b`

Did not run `vite build` locally. Did not re-smoke staging after the takeover commit.

PR #236, opened 2026-08-13
feat(age-review): lead the parent outreach email with what we need

mbradley:
## Summary

- The first message a parent receives now leads with the verification video and its requirements, instead of asking them to confirm they are the parent.
- It names the account under review, so a parent knows which account this concerns.
- Sent as HTML, so the do-not-send warning reads as a warning and the family and kids pages are links.
- One template serves both age bands this path reaches; the age-band label is gone.

## Motivation

The opening message asked a parent only to confirm they were the parent. Parents replied "yes", which establishes nothing, and the real requirements arrived a round trip later. Anyone who is not the parent cannot produce the video regardless, so the confirmation step bought nothing and cost a step. Copy is signed off by the Trust & Safety leads, including the legal-permission qualifier, the emphasis on what not to send, and the family-resources framing.

Two things fell out of writing it down:

- The age-band label rendered as "may belong to a minor in the 16+ (claimed) age range" for the band where nobody is claiming the holder is a minor. One template with no label is correct for both.
- The video list omitted "a clear statement that the teen is between 13 and 15", which both divine.video/kids and divine.video/age-review require. A parent following the email exactly produced a video missing a required element, costing the round trip this change exists to remove.

## This reverses a decision from #222 - please look here first

#222 established that this pre-reply message stays handle-free: the address is supplied by the account under review and is unverified until someone replies from it, so naming the account discloses a minor's handle to whoever holds a mistyped address.

This names the account anyway. The tradeoff was weighed and taken: a parent cannot act on a review without knowing which account it concerns, and the earlier approach of asking them to supply an identifier produced display names that identify nothing.

The residual risk is real and unchanged. The reversal is scoped as narrowly as it can be:

- The Zendesk requester name stays address-only, so no handle reaches the To: header of outbound mail.
- The "Claimed parent of &lt;handle&gt;" rename still waits for a reply.
- Agents keep the fuller block - case deeplink, band, claimed-and-unverified qualifier.

Only the message body changed. Two of #222's assertions flipped to match, deliberately and visibly rather than quietly.

## Hardening that came out of review

Naming the account puts attacker-chosen text into outbound mail, and the same account also chooses the recipient. Escaping stops markup but not mail-client auto-linking, so:

- A display name that looks like a link is dropped rather than rendered. Otherwise a name of "verify at http://not-divine.example/claim" would have Divine Trust & Safety deliver an attacker's link, under Divine branding, to an address the attacker picked.
- The username renders only on a domain we issue, read from `NIP05_DOMAIN`. A NIP-05 is a bare hostname by construction, so it cannot be screened the same way. Unset means no username row, matching the call #222 made for capture - staging has no identity domain of its own.
- Unicode dot equivalents, zero-width characters and control characters are normalized or stripped before the check, so a hostname cannot hide from us while remaining legible to the mail client.
- Truncation is by code point at 80 with an ellipsis, so an emoji is never cut into a lone surrogate.

## Validation

| Check | Result |
|---|---|
| `cd worker && npx vitest run` | 533 passed |
| `cd worker && npm run test:d1` | 26 passed |
| `cd worker && npm run typecheck` | clean |
| `npx tsc -p tsconfig.app.json --noEmit` | clean |
| `npx eslint` | clean |
| root `npx vitest run` | 609 passed, 1 skipped |

HTML delivery was verified against the live Zendesk instance, not inferred from settings: a throwaway ticket confirmed bullets render as a list, the warning renders bold, and both links are live anchors. Ticket and its throwaway end user were deleted afterwards. The script that does it is in `divine-zendesk-tooling`.

Every guard here is mutation-tested - each was removed and the test confirmed to go red. Two rounds of that found tests of mine that passed for the wrong reason; both are fixed and the reasoning is recorded where it happened.

## Out of scope

The email says "reply within 15 days". The clock's actual behaviour on pause and resume is being handled separately and is not touched here.

### DIFF
--- a/worker/src/age-review.test.ts
+++ b/worker/src/age-review.test.ts
@@ -2220,16 +2220,41 @@ describe('buildParentOutreachBody', () => {
   });
 
   it('drops a NIP-05 that is not actually a NIP-05', () => {
+    // The issuing domain must be supplied, or this asserts nothing: without it
+    // displayNip05 returns early and the shape check never runs.
     const html = buildParentOutreachBody({
       pubkey: 'a'.repeat(64),
       account_nip05: 'http://not-divine.example/claim',
-    });
+    }, 'divine.video');
     expect(html).not.toContain('Username:');
     expect(html).not.toContain('not-divine.example');
   });
 
+  it('drops a NIP-05 smuggling a URL in front of a domain we do issue', () => {
+    // The host really is divine.video, so the domain gate passes it. Only the
+    // shape check stands between this and a live link to evil.example in mail
+    // sent under Divine branding.
+    const html = buildParentOutreachBody({
+      pubkey: 'a'.repeat(64),
+      account_nip05: 'http://evil.example/claim@divine.video',
+    }, 'divine.video');
+    expect(html).not.toContain('Username:');
+    expect(html).not.toContain('evil.example');
+  });
+
+  it('drops a NIP-05 carrying markup in front of a domain we do issue', () => {
+    const html = buildParentOutreachBody({
+      pubkey: 'a'.repeat(64),
+      account_nip05: '<script>alert(1)</script>@divine.video',
+    }, 'divine.video');
+    expect(html).not.toContain('Username:');
+    expect(html).not.toContain('script');
+  });
+
   it('prints no empty row for a NIP-05 that is only the root prefix', () => {
-    const html = buildParentOutreachBody({ pubkey: 'a'.repeat(64), account_nip05: '_@' });
+    const html = buildParentOutreachBody(
+      { pubkey: 'a'.repeat(64), account_nip05: '_@' }, 'divine.video',
+    );
     expect(html).not.toContain('Username:');
   });
 
@@ -2251,6 +2276,16 @@ describe('buildParentOutreachBody', () => {
     });
     expect(html).toContain(`${'x'.repeat(79)}\u{1F600}\u2026`);
     expect(/[\uD800-\uDBFF](?![\uDC00-\uDFFF])/.test(html)).toBe(false);
+  });
+
+  it('drops a display name carrying a bare IP address', () => {
+    // No letters after the final dot, so the hostname rule alone does not see
+    // this one.
+    const html = buildParentOutreachBody({
+      pubkey: 'a'.repeat(64),
+      account_name: 'Star Girl go to 93.184.216.34/claim',
+    });
+    expect(html).not.toContain('Display name:');
   });
 
   it('drops a display name using a unicode dot to hide a hostname', () => {
--- a/worker/src/age-review.test.ts
+++ b/worker/src/age-review.test.ts
@@ -2143,7 +2143,7 @@ describe('buildParentOutreachBody', () => {
       pubkey: 'a'.repeat(64),
       account_name: 'Star Girl',
       account_nip05: '_@stargirl.divine.video',
-    });
+    }, 'divine.video');
     expect(html).toContain('The account under review:');
     expect(html).toContain('Display name: Star Girl');
     // NIP-05 root identifiers display as the bare domain, so the stored
@@ -2156,9 +2156,42 @@ describe('buildParentOutreachBody', () => {
   it('leaves a non-root NIP-05 local part alone', () => {
     const html = buildParentOutreachBody({
       pubkey: 'a'.repeat(64),
-      account_nip05: 'alice@example.com',
+      account_nip05: 'alice@divine.video',
+    }, 'divine.video');
+    expect(html).toContain('Username: alice@divine.video');
+  });
+
+  it('drops a NIP-05 on a domain we do not issue', () => {
+    // A NIP-05 is a bare hostname, so looksLinkish cannot screen it without
+    // rejecting every real one. The issuing domain is the screen instead: this
+    // value is unverified kind-0, and rendering it would mail an
+    // attacker-chosen, highly linkifiable hostname under Divine branding.
+    const html = buildParentOutreachBody({
+      pubkey: 'a'.repeat(64),
+      account_nip05: '_@claim-your-teen-account.example',
+    }, 'divine.video');
+    expect(html).not.toContain('Username:');
+    expect(html).not.toContain('claim-your-teen-account');
+  });
+
+  it('renders no username at all when no issuing domain is configured', () => {
+    // Staging deliberately leaves NIP05_DOMAIN unset. No username beats one we
+    // cannot vouch for.
+    const html = buildParentOutreachBody({
+      pubkey: 'a'.repeat(64),
+      account_nip05: '_@stargirl.divine.video',
     });
-    expect(html).toContain('Username: alice@example.com');
+    expect(html).not.toContain('Username:');
+  });
+
+  it('accepts a subdomain of the issuing domain but not a lookalike', () => {
+    expect(buildParentOutreachBody(
+      { pubkey: 'a'.repeat(64), account_nip05: '_@kid.divine.video' }, 'divine.video',
+    )).toContain('Username: kid.divine.video');
+    // `notdivine.video` ends with the domain as a substring but is not ours.
+    expect(buildParentOutreachBody(
+      { pubkey: 'a'.repeat(64), account_nip05: '_@notdivine.video' }, 'divine.video',
+    )).not.toContain('Username:');
   });
 
   it('escapes an ID that could not be encoded as an npub', () => {
@@ -2178,7 +2211,7 @@ describe('buildParentOutreachBody', () => {
       pubkey: 'a'.repeat(64),
       account_name: 'Star Girl verify at http://not-divine.example/claim',
       account_nip05: '_@stargirl.divine.video',
-    });
+    }, 'divine.video');
     expect(html).not.toContain('Display name:');
     expect(html).not.toContain('not-divine.example');
     // Failing safe still leaves the parent able to identify the account.
@@ -2200,13 +2233,42 @@ describe('buildParentOutreachBody', () => {
     expect(html).not.toContain('Username:');
   });
 
-  it('caps an overlong display name rather than mailing it whole', () => {
+  it('caps an overlong display name and marks it as cut', () => {
     const html = buildParentOutreachBody({
       pubkey: 'a'.repeat(64),
       account_name: 'Q'.repeat(300),
     });
-    expect(html).toContain(`Display name: ${'Q'.repeat(80)}<`);
+    expect(html).toContain(`Display name: ${'Q'.repeat(80)}\u2026`);
     expect(html).not.toContain('Q'.repeat(81));
+  });
+
+  it('truncates by code point, so an emoji is never cut in half', () => {
+    // A lone surrogate in html_body is malformed and Zendesk may reject the
+    // whole comment, which would drop the outreach entirely.
+    const html = buildParentOutreachBody({
+      pubkey: 'a'.repeat(64),
+      account_name: `${'x'.repeat(79)}\u{1F600}extra`,
+    });
+    expect(html).toContain(`${'x'.repeat(79)}\u{1F600}\u2026`);
+    expect(/[\uD800-\uDBFF](?![\uDC00-\uDFFF])/.test(html)).toBe(false);
+  });
+
+  it('drops a display name using a unicode dot to hide a hostname', () => {
+    // UTS-46 maps U+3002 to '.', so a mail client resolves this as a hostname
+    // even though an ASCII-only check does not see one.
+    const html = buildParentOutreachBody({
+      pubkey: 'a'.repeat(64),
+      account_name: 'Star Girl go to evil\u3002example',
+    });
+    expect(html).not.toContain('Display name:');
+  });
+
+  it('drops a display name hiding a dot behind a zero-width character', () => {
+    const html = buildParentOutreachBody({
+      pubkey: 'a'.repeat(64),
+      account_name: 'Star Girl evil\u200b.example',
+    });
+    expect(html).not.toContain('Display name:');
   });
 
   it('omits the rows it has no value for rather than printing empties', () => {
--- a/worker/src/age-review.test.ts
+++ b/worker/src/age-review.test.ts
@@ -1593,8 +1593,8 @@ describe('parent contact record', () => {
     expect(payload.ticket.requester.name).toBe('parent@example.com');
     expect(payload.ticket.requester.name).not.toContain('Claimed parent');
     expect(payload.ticket.requester.name).not.toBe('Parent/Guardian');
-    // The outreach is sent as html_body; assert on the field that actually ships,
-    // or this passes vacuously against an undefined `body`.
+    // The outreach ships as html_body now; asserted against `body` this errors
+    // on an undefined receiver rather than checking anything.
     expect(payload.ticket.comment.html_body).not.toContain('Some One');
   });
 
@@ -2037,6 +2037,11 @@ describe('handleParentContact Zendesk integration', () => {
 
     // Nor may the message body carry it: same unverified address.
     expect(ticketPut.ticket.comment.html_body).not.toContain('Some One');
+
+    // Zendesk sends `body` when both are present, so a plain body reappearing
+    // here would silently revert the message to text. This is the branch a
+    // moderator-restricted case takes, so it needs the guard as much as create.
+    expect(ticketPut.ticket.comment.body).toBeUndefined();
 
     // Agent-only: the block reached the contact, with the right case.
     const contactPut = mockFetch.mock.calls.find(
--- a/worker/src/age-review.test.ts
+++ b/worker/src/age-review.test.ts
@@ -1593,9 +1593,10 @@ describe('parent contact record', () => {
     expect(payload.ticket.requester.name).toBe('parent@example.com');
     expect(payload.ticket.requester.name).not.toContain('Claimed parent');
     expect(payload.ticket.requester.name).not.toBe('Parent/Guardian');
-    // The outreach ships as html_body now; asserted against `body` this errors
-    // on an undefined receiver rather than checking anything.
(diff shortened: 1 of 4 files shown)


worker/src/age-review.ts
```diff
@@ -867,20 +867,238 @@ async function getZendeskClientConfig(env: AgeReviewEnv): Promise<ZendeskClientC
...
+  // class cannot be exempted from no-control-regex inline. report-note.ts walks
+  // the string the same way.
+  const stripped = Array.from(value)
+    .filter((ch) => {
+      const c = ch.codePointAt(0) ?? 0;
+      return !(
+        c <= 0x1f || c === 0x7f || c === 0xad ||
+        (c >= 0x200b && c <= 0x200f) || c === 0x2060 || c === 0xfeff
+      );
+    })
```
realmeylisdev: This list covers 8 code points. The set that defeats `looksLinkish` is much larger.

Running the shipped `buildParentOutreachBody`, 25 of 32 format / default-ignorable characters I tried get a hostname through. Your test at line 2301 with one character swapped:

```js
account_name: 'Star Girl evil\u200b.example'  // dropped, as the test asserts
account_name: 'Star Girl evil\u2062.example'  // "Display name: Star Girl evil.example"
```

U+2062 is INVISIBLE TIMES. Also through: U+2061/2063/2064, U+206A-206F, U+061C, U+180E, U+FFF9-FFFB, U+E0001 and U+E0020-E007F, the variation selectors U+FE0E/FE0F/E0100, and the bidi controls U+202A-202E / U+2066-2069. That last group is the one I'd weigh most — an unterminated U+202E reorders the remainder of the `<p>`, and the npub row is inside it.

Stripping by Unicode property instead of by enumeration closes all 32. I ran this exact change: 533/533 worker tests pass, `npm run typecheck` clean, `npx eslint` clean. Property escapes don't trip `no-control-regex`, so the `noInlineConfig` constraint in your comment doesn't apply to this form.

```suggestion
  // Control characters, invisible/format characters, and unpaired surrogates,
  // tested by Unicode property rather than an enumerated list: the set that can
  // hide a hostname from `looksLinkish` while a mail client still parses one is
  // far wider than the well-known zero-width characters, and includes the bidi
  // controls. Property escapes do not trip no-control-regex, so noInlineConfig
  // is not a constraint here.
  const stripped = Array.from(value)
    .filter((ch) => !/[\p{Cc}\p{Cf}\p{Cs}\p{Co}\p{Zl}\p{Zp}\p{Default_Ignorable_Code_Point}]/u.test(ch))
```

`\p{Cs}` also closes the input side of the lone-surrogate case your truncation test guards against — `JSON.parse('"\\ud800"')` is reachable from kind-0, so one can arrive already in `account_name` rather than being created by the slice.

One cost worth naming: variation selectors go too, so an emoji loses its presentation selector. ZWJ sequences already split today, since U+200D is in the current range.

worker/src/age-review.ts
```diff
@@ -867,20 +867,238 @@ async function getZendeskClientConfig(env: AgeReviewEnv): Promise<ZendeskClientC
...
+ * Keycast instance, and no username beats a wrong or unvouched one.
+ *
+ * NIP-05 calls `_@domain` the "root" identifier and says to display it as the
+ * bare domain (nips/05.md, "Showing just the domain as an identifier"). Divine
+ * issues `_@<username>.<NIP05_DOMAIN>`, so the prefix is noise to a parent. Any
+ * other local part is a real name and is left intact.
+ */
+function displayNip05(nip05: string, issuingDomain?: string): string | undefined {
+  if (!issuingDomain) return undefined;
+  if (!/^[a-z0-9._-]+@[a-z0-9-]+(\.[a-z0-9-]+)+$/i.test(nip05)) return undefined;
```
realmeylisdev, line 981: A question rather than a finding — I couldn't prove impact, so I'm not asking for a change.

The shape check and the domain gate both constrain the **host**. The local part is only charset-limited, and NIP-05 permits `.` there, so it can hold something hostname-shaped:

```
account_nip05: 'www.evil.example@divine.video'
  ->  Username: www.evil.example@divine.video
```

The same string as a display name is dropped by `looksLinkish`. So "renders only on a domain we issue" holds for the host, but the full rendered string still contains attacker-chosen text that reads like a hostname.

Whether that matters comes down to whether any real client linkifies the local part separately from the address, which I couldn't establish either way — most would produce a single `mailto:`. Raising it because `_@<username>.<NIP05_DOMAIN>` is the only form Divine issues, so the leniency that keeps `alice@divine.video` working is carrying kind-0 values it wasn't aimed at. If that's deliberate, a line saying so would settle it.

realmeylisdev (changes requested):
Pulled the branch and re-ran the table: 533, 26, 609 + 1 skipped, both typechecks and eslint all reproduce exactly. The `_@` stripping matches what NIP-05 actually says under "Showing just the domain as an identifier", and `html_body` carrying `<p>`/`<ul>`/`<strong>` is the shape Zendesk documents. Flipping the two #222 assertions in the open rather than quietly is the right call, and the reasoning for the reversal is clearer than most design docs.

One thing to fix first.

**The invisible-character list doesn't cover the case it was written for.** `cleanIdentityText` strips 8 code points. Running the shipped `buildParentOutreachBody`, 25 of the 32 format / default-ignorable characters I tried still get a hostname past `looksLinkish`. Your test on line 2301, with one character changed:

```
'Star Girl evil\u200b.example'  ->  dropped
'Star Girl evil\u2062.example'  ->  Display name rendered
```

Both are invisible, and a mail client's parser can't tell them apart — which is the argument your own comment makes for stripping the first one. Bidi controls get through as well, landing in `html_body` in the same `<p>` as the npub.

Suggestion inline. I ran it: 533/533 worker tests pass, typecheck and eslint clean.

The NIP-05 local part has a related gap, but I couldn't establish that it reaches anything, so I've left it as a question rather than a requested change.

worker/src/age-review.ts
mbradley: Fixed — property-based strip in place of the enumeration. Reproduced it first: U+2062 and an unterminated U+202E both rendered through, the RLO landing in the npub's `<p>`. Both drop now; legit names including emoji still render. Two regression tests added (invisible-format + bidi, the latter asserting no U+202E reaches the body), each mutation-checked red against the old filter. 536 worker tests, typecheck + eslint clean — property escapes don't trip no-control-regex, as you said.

worker/src/age-review.ts
mbradley: Tightened rather than documented. `displayNip05` now drops the row when the local part itself `looksLinkish`, so `www.evil.example@divine.video` is refused even though the host is ours. `_@…` and a plain `alice` survive; only a dotted, hostname-shaped local part drops. Regression test added and mutation-checked; the `alice@divine.video` leniency test still passes, so it doesn't over-catch.

worker/src/age-review.ts
```diff
@@ -867,20 +867,243 @@ async function getZendeskClientConfig(env: AgeReviewEnv): Promise<ZendeskClientC
...
+ */
+function cleanIdentityText(value: string): string {
+  // Control characters, invisible/format characters, and unpaired surrogates,
+  // tested by Unicode property rather than an enumerated list: the set that can
+  // hide a hostname from `looksLinkish` while a mail client still parses one is
+  // far wider than the well-known zero-width characters, and includes the bidi
+  // controls. Property escapes do not trip no-control-regex, so noInlineConfig
+  // is not a constraint here.
+  const stripped = Array.from(value)
+    .filter((ch) => !/[\p{Cc}\p{Cf}\p{Cs}\p{Co}\p{Zl}\p{Zp}\p{Default_Ignorable_Code_Point}]/u.test(ch))
```
realmeylisdev, line 908: `\p{Default_Ignorable_Code_Point}` has no test holding it. Delete it from this class and all 536 still pass.

It is load-bearing. 4032 code points are in it but not in `Cc/Cf/Cs/Co/Zl/Zp`, and the variation selectors among them are ICU-resolvable:

```js
account_name: 'Star Girl evil\uFE0F.example'
//   with the class:  dropped
//   without it:      Display name: Star Girl evil.example

new URL('http://evil\uFE0F.example/').hostname  // 'evil.example'
```

U+FE0F is the emoji presentation selector, so this is a shape that reaches you through ordinary display names rather than only crafted ones. Same for U+FE00–FE0E, U+180B–180F, U+034F and U+E0100+.

Your three invisible-character tests (`\u200b`, `\u2062`, `\u202e`) are all `Cf`, so removing `\p{Cf}` from this class does go red — it is only the `\p{DI}` half that is unheld.

worker/src/age-review.ts
```diff
@@ -867,20 +867,243 @@ async function getZendeskClientConfig(env: AgeReviewEnv): Promise<ZendeskClientC
...
+ * hostname to the client and was not to an ASCII-only check. NFKC alone does
+ * not cover it -- it folds U+FF0E and leaves the other two -- so they are
+ * replaced outright.
+ *
+ * Deliberately blunt. A dotted token with a letter suffix is enough to trip it,
+ * so a legitimate name like "Anna.Marie" is dropped too. Losing a display name
+ * costs recognisability; shipping a live link costs more.
+ */
+function looksLinkish(value: string): boolean {
+  const normalized = value.normalize('NFKC').replace(/[。．｡]/g, '.');
```
realmeylisdev, line 943: Same story for `.normalize('NFKC')` — take it out and the suite stays green, because the U+3002 test is caught by the explicit `.replace(/[。．｡]/g, '.')` on this line rather than by the fold.

What the fold uniquely catches is fullwidth, and that one resolves:

```js
account_name: 'Star Girl go to \uFF19\uFF13\uFF0E\uFF11\uFF18\uFF14\uFF0E\uFF12\uFF11\uFF16\uFF0E\uFF13\uFF14'
//   with NFKC:  dropped
//   without:    rendered

new URL('http://９３．１８４．２１６．３４/').hostname  // '93.184.216.34'
```

The IPv4 branch below uses ASCII `\d`, and the explicit replace only touches the dots, so with the fold gone neither branch sees it.

worker/src/age-review.test.ts
```diff
@@ -2043,6 +2057,314 @@ describe('handleParentContact Zendesk integration', () => {
...
+  it('drops a name carrying a bidi override and never emits one into the body', () => {
+    // U+202E (RIGHT-TO-LEFT OVERRIDE), left unterminated, reorders the rest of
+    // the paragraph it lands in -- and the npub row shares that paragraph.
+    const html = buildParentOutreachBody({
+      pubkey: 'a'.repeat(64),
+      account_name: 'Star Girl evil\u202e.example',
+    });
+    expect(html).not.toContain('Display name:');
+    expect(html).not.toContain('\u202e');
+  });
```
realmeylisdev, line 2341: Four tests, one per uncovered guard. Verified against `0ee65ff`: all four pass on your code, and each goes red against exactly its own mutant and nothing else —

- variation selector → `\p{Default_Ignorable_Code_Point}`
- fullwidth digits → `.normalize('NFKC')`
- scheme with a dotless host → the `://` / `www.` branch
- padded NIP-05 → `.trim()` on `account_nip05`

540 worker, 26 D1, typecheck and eslint clean with these in.

```suggestion
  });

  it('drops a name hiding a dot behind a variation selector', () => {
    // U+FE0F is Default_Ignorable but not Cf, so the \p{Cf} half of the strip
    // never reaches it -- and it is the ordinary emoji presentation selector, so
    // it turns up in real display names rather than only in crafted ones. ICU
    // resolves `evil\uFE0F.example` to `evil.example`, so without
    // \p{Default_Ignorable_Code_Point} this renders a live hostname.
    const html = buildParentOutreachBody({
      pubkey: 'a'.repeat(64),
      account_name: 'Star Girl evil\uFE0F.example',
    });
    expect(html).not.toContain('Display name:');
  });

  it('drops a display name whose IP address is written in fullwidth digits', () => {
    // The IPv4 rule matches ASCII \d only, and the explicit dot replacement does
    // not touch the digits. NFKC is the only thing that folds this to
    // `93.184.216.34` -- which is exactly what a UTS-46 resolver sees.
    const html = buildParentOutreachBody({
      pubkey: 'a'.repeat(64),
      account_name: 'Star Girl go to \uFF19\uFF13\uFF0E\uFF11\uFF18\uFF14\uFF0E\uFF12\uFF11\uFF16\uFF0E\uFF13\uFF14',
    });
    expect(html).not.toContain('Display name:');
  });

  it('drops a display name carrying a scheme even when the host has no dot', () => {
    // The hostname rule needs a dot after the label, so an intranet or localhost
    // target is caught by the scheme check alone.
    const html = buildParentOutreachBody({
      pubkey: 'a'.repeat(64),
      account_name: 'Star Girl go to http://localhost/claim',
    });
    expect(html).not.toContain('Display name:');
  });

  it('accepts a NIP-05 that arrived with surrounding whitespace', () => {
    // Capture stores kind-0 verbatim and never trims, so a padded value reaches
    // the anchored shape check as-is and would be dropped without the trim.
    const html = buildParentOutreachBody({
      pubkey: 'a'.repeat(64),
      account_nip05: '  _@stargirl.divine.video\n',
    }, 'divine.video');
    expect(html).toContain('Username: stargirl.divine.video');
  });
```

realmeylisdev (changes requested):
Re-ran the table on `0ee65ff` — 536 / 26 / 609 + 1 skipped, both typechecks, eslint. All of it reproduces.

The property strip does close what I raised last round, and I checked it properly this time: I swept every Unicode code point against Node's ICU host parser, and there is now **no** character that a UTS-46 resolver treats as a dot and `looksLinkish` misses. The local-part gate on `displayNip05` is right too, and `_@` → bare domain is what 05.md actually sanctions ("Showing just the domain as an identifier").

One thing before merge.

**Four guards in this diff can be deleted with the suite still green.** I mutation-tested the 26 guards in the change; 22 die. These four don't:

| delete | suite | what it was holding back |
| --- | --- | --- |
| `\p{Default_Ignorable_Code_Point}` | 536 green | `evil\uFE0F.example` renders — ICU resolves it to `evil.example` |
| `.normalize('NFKC')` | 536 green | a fullwidth-digit IP renders — ICU resolves it to `93.184.216.34` |
| the `://` / `www.` branch | 536 green | `http://localhost/claim`, `https://intranet/pay` |
| `.trim()` on `account_nip05` | 536 green | a whitespace-padded kind-0 value |

The top two are the ones I'd want held down. Your three invisible-character tests all use `Cf` characters, so they pin the `\p{Cf}` half of the widening and leave the `\p{DI}` half loose — and U+FE0F is the emoji presentation selector, so it arrives in ordinary display names, not only crafted ones. Same shape as last round's finding, one layer in. NFKC's only unique job here is the fullwidth fold; the U+3002 test is caught by the explicit `replace` on the same line, not by NFKC.

Suggestion inline: four tests, one per guard. I ran them — all four pass against your code, and each goes red against exactly its own mutant and nothing else. 540 worker, 26 D1, typecheck and eslint clean.

Everything else I'd merge as-is. Flipping the two #222 assertions in the open, and writing down why, is the right call — and the escaping notes are unusually clear about what they do and don't cover.

worker/src/age-review.test.ts
mbradley: Applied all four. Verified the gap first — on `0ee65ff` each of the four guards deletes to a green 536, confirming they were unpinned. Added the four tests and mutation-checked the 1:1 mapping: `\p{Default_Ignorable_Code_Point}` → the variation-selector test, NFKC → the fullwidth-IP test, the scheme branch → the dotless-host test, `.trim()` on `account_nip05` → the padded-value test. Each mutant reddens exactly its own test and nothing else. 540 worker, typecheck + eslint clean.

worker/src/age-review.ts
mbradley: Pinned — the variation-selector test (U+FE0F) now goes red on deleting `\p{Default_Ignorable_Code_Point}` and on nothing else.

worker/src/age-review.ts
mbradley: Pinned — the fullwidth-IP test goes red on removing `.normalize('NFKC')` and on nothing else. The explicit dot-replace and the ASCII-`\d` IPv4 branch don't fold it, as you noted.

realmeylisdev (approved):
Re-ran the table on `a03821a2` — 540 worker, 26 D1, 609 + 1 skipped at the root, both typechecks, eslint. All of it reproduces.

The four tests landed exactly as suggested. I re-checked the 1:1 mapping at this head rather than carrying last round's result forward — each guard still deletes to exactly one red test and nothing else:

| delete | red test |
| --- | --- |
| `\p{Default_Ignorable_Code_Point}` | variation selector |
| `.normalize('NFKC')` | fullwidth-digit IP |
| the `://` / `www.` branch | scheme with a dotless host |
| `.trim()` on `account_nip05` | padded NIP-05 |

**I re-swept the range with the check I should have used the first time.** My earlier sweep asked whether the row rendered, which is the wrong question — a stripped character renders as `evilexample`, and that is not a hostname. This one takes the *rendered* string and hands each token to ICU's host parser: every code point in `0x0`–`0x10FFFF` minus surrogates, in three shapes (`evil<cp>example`, `evil<cp>.example`, `evil.<cp>example`). Zero cases where something that renders still parses as a hostname. Same result for a fullwidth scheme, a fullwidth host, U+2024, punycode, IPv6-with-scheme and an uppercase scheme.

What does still render, and why I'm not asking for it: a bare decimal or hex IP literal (`3221226219`, `0xC0000201`). Neither is linkified without a scheme, and with one the `://` branch takes it first.

Two things I checked rather than assumed:

- `under_13` is rejected with a 400 before either ticket path, so "one template for both bands this path reaches" holds. An under-13 case cannot receive copy asking for a video that states the teen is between 13 and 15.
- `ageBand` is still load-bearing after the copy change — ticket tags and the agent contact block both use it — so the parameter didn't quietly become dead.

Approving. The reversal of #222 is scoped to the message body, and it is written down where the next person to touch this will actually find it; the escaping comments are unusually honest about what they don't cover.

PR #237, opened 2026-08-14
feat(moderate): record content decisions so ReportWatcher stops undoing them

mbradley:
## Summary

Manual hide and restore actions were not recorded in the state that ReportWatcher uses, so a later report could undo a moderator's decision.
This change coordinates relay visibility and durable moderation state so human intent remains authoritative, including when relay or database work only partially succeeds.

## What Changed

- Added explicit `hide_event` and `allow_event` moderation actions.
- Serialized event visibility operations through ReportWatcher.
- Persisted human direction and auto-hide state atomically where the operation requires both.
- Added recoverable states for uncertain hides and failed restores, and surfaced them in the pending-review UI.
- Added deterministic decision ordering and protected coordinated state actions from generic writes.
- Preserved raw relay-action provenance so later reports cannot silently override a human restore.

## Motivation

The moderation queue must not re-hide content that a moderator deliberately restored.
Partial failures must also remain visible and recoverable instead of being reported as a completed decision.

Linked issue: none.

## Testing

- `npx eslint .`
- `npx vitest run` (625 passed, 1 skipped)
- `npx vite build`
- `cd worker && npm run typecheck`
- `cd worker && npm run test:run` (605 passed)
- `cd worker && npm run test:d1` (33 passed)

Worker end-to-end tests were not run because they require a local relay.
The semantic-title and endpoint-health checks run in CI.

## Risks

Human visibility direction remains persistent by design, so a later automatic report cannot override it.
Uncertain auto-hide outcomes now require explicit moderator confirmation or restoration rather than guessing the relay state.

## Deploy Ordering

The companion webhook-adapter change must deploy after this change because it consumes the new actions and `recorded` result.


dcadenas (approved):
Human moderation intent now survives visibility races.

PR #238, opened 2026-08-15
fix(auth): accept canonical host for mobile pre-auth

rabble:
## Summary

- accept the configured canonical public host for the mobile support pre-auth request
- retain exact NIP-98 scheme, path, query, method, timestamp, and signature validation
- preserve direct-worker compatibility and strict rejection when no host is allowlisted

## Motivation

Mobile should sign and call the canonical public API rather than depend on an internal service hostname. The two minor-review endpoints already support this host-relaxation contract; this completes the same contract for the remaining mobile-facing pre-auth call.

Related to #173. This change is backward-compatible and inert until the canonical edge route is deployed.

## Validation

- npm run test:run (542 passed)
- npm run typecheck
- git diff --check

### DIFF
--- a/worker/src/index.test.ts
+++ b/worker/src/index.test.ts
@@ -1364,13 +1364,12 @@ describe('mobile NIP-98 endpoint host allowlist (#173)', () => {
   });
 });
 
-describe('zendesk pre-auth NIP-98 scope boundary (#173)', () => {
+describe('zendesk pre-auth canonical public host', () => {
   const OWN_HOST_PREAUTH_URL = 'https://api-relay-prod.divine.video/api/zendesk/pre-auth';
   const PUBLIC_HOST_PREAUTH_URL = 'https://api.divine.video/api/zendesk/pre-auth';
 
-  // Tolerates ensureSchemaOnce's DDL (db.ts: a sequence of `.prepare(sql).run()`
-  // calls, no `.bind()`/`.first()` needed for schema setup) as async no-ops. The
-  // pre-auth nonce INSERT is never reached in this test — auth fails first.
+  // Tolerates ensureSchemaOnce's DDL and the pre-auth nonce INSERT as async
+  // no-ops; the response proves canonical-host auth reached token issuance.
   function makeZendeskDb() {
     const statement = (): { bind: () => unknown; run: () => Promise<{ success: boolean }>; first: () => Promise<null> } => ({
       bind: () => statement(),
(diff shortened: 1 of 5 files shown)


NotThatKindOfDrLiz:
@rabble I pushed one scoped takeover commit, `16e5e5d test(auth): clarify NIP-98 allowlist boundary`.

What changed:
- Retargeted the helper-level NIP-98 test so it documents the default empty-allowlist behavior instead of the old Zendesk pre-auth call-site boundary.
- Finished the `hostAllowlistedUrlMatch` comment so it describes the helper contract now that all in-repo callers pass `getNip98AllowedHosts(env)`.

Validation on the pushed head:
- `worker: npm run test:run` — 542 passed
- `worker: npm run typecheck`
- `worker: npm run test:d1` — 26 passed
- root: `npx tsc -p tsconfig.app.json --noEmit`
- `git diff --check`
- GitHub checks: Semantic PR and Test are green

Please review the takeover commit and merge when you are ready.

NotThatKindOfDrLiz (approved):
Approved after takeover cleanup.

I re-reviewed the full diff after `16e5e5d`. The host relaxation remains host-only, exact scheme/path/query/method/timestamp/signature binding is preserved, unset allowlist remains strict, and the stale security-boundary wording is now corrected.

Local validation passed:
- `worker: npm run test:run` — 542 passed
- `worker: npm run typecheck`
- `worker: npm run test:d1` — 26 passed
- root: `npx tsc -p tsconfig.app.json --noEmit`
- `git diff --check`

GitHub checks on the current head are green. Not merging; rabble should review the takeover commit and merge when ready.

realmeylisdev (approved):
Independently re-verified on top of Liz's review — traced the auth logic directly (host is the only thing relaxed; scheme, path, query, method, timestamp, and signature all stay exactly bound) and reran everything myself: 542 unit + 26 D1 tests green, both typechecks clean, no whitespace issues.

Also checked outside the diff: the companion `divine-router#25` routing PR is correctly gated on this landing first, and the two follow-up hardening issues filed against this same auth path either don't apply here or are already resolved — #195 (replay protection) is closed, and Zendesk pre-auth's nonce was the reference implementation it was measuring the other endpoints against; #196 (body-hash binding) is open but scoped to parent-contact, not applicable here since pre-auth never reads a request body. NIP-98 itself is silent on proxies, so the host-only relaxation doesn't violate the spec. Mobile still signs and sends this call straight to `api-relay-prod.divine.video` today, so this is genuinely inert until the router PR ships.

Approving.

PR #239, opened 2026-08-17
Standalone admin media viewer for restricted content, linked from Coop

mbradley:
When content is already restricted, such as by an automated verdict, a moderator cannot view the reported media: the card's `media_url` no longer serves. This adds a minimal viewer so they can check the actual bytes before acting, without dropping them into relay-manager's full UI.

`GET /media/<sha256>` on the worker serves a self-contained page (not the SPA): the restricted media streamed through the existing `/api/media-proxy` admin bypass, a "restricted, admin view" banner, and the sha. It sits inside the same CF Access + `verifyAdminAccess` gate as every other admin route, so only an authenticated moderator reaches it, and it reuses the host that already serves the proxy, so no new config or origin is involved. Chosen over an SPA route because that host is already CF-Access-gated and it sidesteps the shared-frontend runtime-backend selector entirely.

Security posture, since the page can serve CSAM:

- one validity check drives both the HTTP status and the body, so a loosened check cannot 200 the error page or 400 the real one
- the sha is normalized once and only that value is interpolated, so the validated string and the rendered string cannot diverge
- document headers: `no-store` (the URL and body name a content hash), a locked-down CSP, `nosniff`, `no-referrer`
- proxy response (`/api/media-proxy`): `no-store` (restricted bytes never reach the browser HTTP disk cache), `nosniff` + `CSP: sandbox` (a blob stored as `text/html` cannot execute on the admin origin if navigated to directly); the viewer's video controls suppress the browser download/Cast/PiP affordances, and the optional URL extension is validated server-side before interpolation
- unknown blob types are not offered as a download, so possibly-CSAM bytes never touch a moderator's disk; the page shows the type and sends them back to Coop

Tests: `renderMediaPage` unit + route integration including the auth gate; 629 worker tests pass. The gate, the CSP header, the lowercase normalization, and the no-download rule are each mutation-checked.

**Sequencing:** this is the target of the osprey `relay_manager_url` change and must merge and deploy **before** it, or Coop cards would carry links that 404. Acceptance check after deploy: load `/media/<sha>` in a browser as a moderator on staging and confirm the media renders.

### DIFF
--- a/worker/src/index.ts
+++ b/worker/src/index.ts
@@ -475,11 +475,25 @@ export default {
       // authenticated moderator reaches it. Serves a minimal self-contained page,
       // NOT the relay-manager SPA, and streams the blob via /api/media-proxy.
       if (path.startsWith('/media/') && request.method === 'GET') {
-        const raw = path.slice('/media/'.length).replace(/\.[^./]+$/, '').toLowerCase();
-        const valid = /^[a-f0-9]{64}$/.test(raw);
-        return new Response(renderMediaPage(raw), {
-          status: valid ? 200 : 400,
-          headers: { 'content-type': 'text/html; charset=utf-8', ...corsHeaders },
+        const raw = path.slice('/media/'.length).replace(/\.[^./]+$/, '');
+        const { status, html } = renderMediaPage(raw);
+        return new Response(html, {
+          status,
+          headers: {
+            ...corsHeaders,
+            'content-type': 'text/html; charset=utf-8',
+            // The URL and body name a content hash (potentially CSAM); keep it off disk.
+            'cache-control': 'no-store',
+            // Locks the page to exactly what it needs: same-origin fetch to the proxy,
+            // its own inline script/style, blob: media. Also makes the inline-script
+            // dependency explicit rather than resting on no CSP ever being added here.
+            'content-security-policy':
+              "default-src 'none'; script-src 'unsafe-inline'; connect-src 'self'; " +
+              "img-src blob:; media-src blob:; style-src 'unsafe-inline'; " +
+              "frame-ancestors 'none'; base-uri 'none'",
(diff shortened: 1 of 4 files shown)


dcadenas (approved):
The viewer design is sound: the admin gate placement, the one-check-drives-both status/body shape, and the no-download rule for unknown types are all right. I hardened the byte path around it.

### Changes I pushed

- The media proxy now sends `no-store` so restricted bytes never reach the browser HTTP disk cache, plus `nosniff` and a `sandbox` CSP so a blob stored as `text/html` cannot run script on this origin when its proxy URL is navigated to directly.
- The viewer's video controls suppress the browser's download, Cast, and Picture-in-Picture affordances.
- The optional URL extension is kept as a fallback type hint when the proxy reports an unhelpful content type, and is validated server-side before anything reaches the document.
- Media that fails to decode now shows an error line instead of a silent black player.

The `sandbox` choice over `Content-Disposition: attachment` is the one worth naming: `attachment` forces exactly the write-to-disk the viewer exists to avoid, while `sandbox` neutralizes direct navigation and leaves the fetch-into-blob consumers untouched.

### Checks

- Worker typecheck: `tsc -p tsconfig.json --noEmit`, clean
- Worker suite: `vitest run`, 21 files, 629 tests passed
- Worker D1 suite: 5 files, 33 tests passed
- Not run locally: the root `npm run test` (app typecheck, eslint, vite build) and `test:e2e` (needs a local relay). No frontend files changed; CI covers the root job.

dcadenas (approved):
✅ lint-only follow-up on top of the hardening already reviewed: the one-character regex cleanup keeps the extension-hint parsing identical and satisfies no-useless-escape. Root eslint now reproduced locally with the repo-pinned plugin versions and clean repo-wide.

PR #240, opened 2026-08-18
fix(reports): scroll detail pane on short viewports; dismissable attribution banner

mbradley:
## Summary

Fixes a height-driven layout bug that made the Reports tab unusable on short (laptop, full-window) viewports, and makes the attribution banner dismissable.

On a short viewport the report **detail pane** clipped its content with no way to scroll: the action-button footer was a `shrink-0` block sitting *outside* the scroll region, so it squeezed the report context to a sliver and, when the footer alone exceeded the pane height, overflowed the pane's `overflow-hidden` and became unreachable. A moderator on a full-window laptop could not see the reported content or reach the action buttons ("won't let me see anything or even scroll the page").

The fix scrolls the whole detail pane rather than an inner region: the content keeps its natural height, and the footer stays pinned to the bottom when it fits (tall screens, unchanged) and scrolls into view with the content when it doesn't (short screens).

Separately, the "Moderation actions are being recorded without attribution" banner is now dismissable (persisted), since it is informational only and steals vertical space that matters on short viewports.

## Motivation

A moderator was fully blocked. The root cause was viewport **height**, not width or the sign-in flow: the fixed-height shell plus a non-shrinking action footer. Reproduced from the reported screen (1888×1256 @2× = 944×628 CSS).

## Changes

- `ReportDetail.tsx`: detail pane `overflow-hidden` → `overflow-y-auto`; inner `ScrollArea` (`flex-1 min-h-0`) → a plain `flex-1` div (drops the now-unused `ScrollArea` import).
- `AttributionNotice.tsx`: add a dismiss control; persist dismissal in `localStorage`, guarded so disabled storage never crashes the banner.

## Deploy status

Already deployed to CF Pages **Production** (commit `534089e`) as an urgent hotfix for the blocked moderator. This PR lands the same change on `main` so prod and the repo stay in sync — without it, the next `main` frontend deploy would revert the fix.

## Validation

- `npx tsc -p tsconfig.app.json --noEmit`: clean.
- `AttributionNotice.test.tsx`: 4/4 pass; accessibility snapshot confirms both **Sign in** and **Dismiss** render.
- Could not screenshot a live report locally — the Reports queue requires the relay stack / CF Access, so the detail pane can't be populated in local dev. Functional confirmation is a moderator hard-reloading `relay.admin.divine.video/reports` on a short laptop viewport, opening a report, and confirming the detail pane scrolls and the action buttons are reachable.

## No linked issue

Urgent hotfix in response to a blocked moderator; no pre-existing tracking issue. Can file one if we want it tracked.

## Follow-up (not in this PR)

The Age Review detail pane (`AgeReviewDetail.tsx`) uses the same `overflow-hidden` + `shrink-0` footer pattern and likely clips the same way on short viewports. Left out to keep this hotfix tight.

### DIFF
--- a/src/components/ReportDetail.tsx
+++ b/src/components/ReportDetail.tsx
@@ -8,7 +8,6 @@ import { Card, CardContent, CardHeader, CardTitle } from "@/components/ui/card";
 import { Button } from "@/components/ui/button";
 import { Badge } from "@/components/ui/badge";
 import { Separator } from "@/components/ui/separator";
-import { ScrollArea } from "@/components/ui/scroll-area";
 import {
   AlertDialog,
   AlertDialogAction,
@@ -469,8 +468,14 @@ export function ReportDetail({ report, allReportsForTarget, allReports = [], onD
   const categoryLabel = CATEGORY_LABELS[category] || category;
   const isHighPriorityCategory = HIGH_PRIORITY_CATEGORIES.includes(category);
 
+  // Scroll the whole pane, not just an inner region. On a short viewport the
+  // tall action footer used to starve the content area (it was a shrink-0 block
+  // outside the scroll), squeezing the report context to a sliver and making it
+  // unreachable. Scrolling the pane keeps the footer pinned to the bottom when
+  // it fits (tall screens) and lets the content + footer scroll together when it
+  // does not (short laptops).
   return (
-    <div className="h-full flex flex-col overflow-hidden">
+    <div className="h-full flex flex-col overflow-y-auto">
       {/* Dialogs - rendered as portals, don't affect flex layout */}
       {/* Ban Confirmation Dialog */}
       {/* Dismiss Report Dialog */}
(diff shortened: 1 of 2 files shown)


dcadenas (approved):
The short-viewport layout now keeps the report context and moderation actions reachable. Persisting the informational-banner dismissal until site storage is cleared is an accepted tradeoff.

### Changes I pushed

- Added regression coverage for persisted dismissal and unavailable storage.
- Clarified when the action footer scrolls with the report content.

### Checks

- `npx --no-install tsc -p tsconfig.app.json --noEmit`
- `npx --no-install eslint .`
- `npx --no-install vitest run`: 70 files passed; 724 tests passed and 1 skipped
- `npx --no-install vite build`

The aggregate `npm test` wrapper was not runnable because npm 12 rejected its leading `npm i` with `EALLOWSCRIPTS`; the constituent frontend checks above passed.

PR #242, opened 2026-08-20
chore(repo): remove stale worker/public artifacts

NotThatKindOfDrLiz:
## What this does

Removes checked-in build artifacts and dead CI config that no longer match how this project deploys, and corrects the deploy docs and example configs that had drifted from the live setup.

## Why

- `worker/public/` held a checked-in frontend build bundle (~712 KB). Nothing serves it: the API worker has no asset binding, and the frontend deploys to Cloudflare Pages from `dist/`. The stale tree had already caused drift (PR #65 edited the dead CSP copy, not the live one).
- `.gitlab-ci.yml` was a dead scaffold for a GitHub-only repo, and its logic would delete the real `public/` dir if it ever ran.
- The deploy docs and example env files still referenced a nonexistent root `worker/wrangler.toml`, the old `relay.dvines.org` and deprecated `api-relay.divine.video` domains, and `MANAGEMENT_PATH = /management` — none of which match the per-environment wrangler configs the project deploys with.

## Changes

- Delete `worker/public/` (5 files) and add it to `.gitignore`.
- Delete `.gitlab-ci.yml`.
- Correct `DEPLOYMENT.md`, `README.md`, `docs/api-webhooks.md`, `.env.example`, and the PR template so documented relay URLs, API domains, `MANAGEMENT_PATH` (`/`), moderation service URL, `--config` usage, and typecheck commands match `worker/wrangler.{local,staging,prod}.toml`.
- Point `worker/package.json` scripts at the per-environment configs (`deploy:staging`, `deploy:prod`, `dev --config wrangler.local.toml`).
- `scripts/seed-test-events.ts`: require `RELAY_URL` explicitly instead of defaulting to a relay (the script publishes fabricated events and reports).
- Test hygiene, unrelated to the deletions: `age-review.test.ts` uses a per-test fresh-mock factory; `bulk-moderate.test.ts` moves to top-level imports with `beforeEach` mock resets. Behavior unchanged.

## What this does not change

No application or worker runtime behavior — docs, config examples, dead-file deletions, and test-mock structure only.

## Verification

`npm run test` (type-check, eslint, vitest, build). No remaining references to `worker/public`, `relay.dvines.org`, `api-relay.divine.video`, or the removed `npm run deploy`; corrected domains and values match the wrangler configs.

Closes #53 (refs #54).


mbradley (dismissed):
Pushed two commits to this branch (now at f69cd5b):

- `3bd93d0` **fix(seed): require RELAY_URL so the seeder can't default to production.** `scripts/seed-test-events.ts` had its fallback changed to `wss://relay.divine.video`, which is the live production relay. That script publishes fabricated profiles, videos, and kind 1984 reports, so a run with `RELAY_URL` unset would have injected fake content and fake reports into the production moderation queue. It now exits with a clear message when `RELAY_URL` is unset instead of defaulting to any relay.
- `f69cd5b` **docs(readme): match the worker typecheck comment to the actual command.** The worker `npm run typecheck` runs `tsc -p tsconfig.json --noEmit`; the comment said `tsc --noEmit`, inconsistent with the frontend line this branch already corrected.

Validation: `eslint` passes on the seed script (eslint lints `scripts/`; `tsc -p tsconfig.app.json` does not cover that path). I did not re-run the full suite, as these two changes are a standalone script and a comment.

The rest of the change checks out against the configs: no remaining references to `worker/public`, `relay.dvines.org`, the deprecated `api-relay.divine.video`, or the removed `npm run deploy`; the corrected relay URLs, `MANAGEMENT_PATH`, and moderation service URL all match the wrangler files; the `404.html` copy and `public/` contents are described accurately.

One item needs a decision before this merges.

The title and description scope this as removing `worker/public`, but the change also rewrites deploy docs and domains and refactors two test files. `age-review.test.ts` and `bulk-moderate.test.ts` are not "test reference updates" (description item 5). They are mock-isolation refactors: a fresh-mock factory in one, top-level imports with `beforeEach` resets in the other. They pass, but they are unrelated to the removed paths. The description also omits `.env.example`, `.github/pull_request_template.md`, `docs/api-webhooks.md`, `scripts/seed-test-events.ts`, and `worker/package.json`.

Pick one:

- Keep the test refactors here and rewrite the description to cover the full change (the config and domain cleanup, and the test refactors described as what they are), and list the omitted files.
- Or split the two test refactors into a separate PR so this one matches its description.

@NotThatKindOfDrLiz requesting changes on the description/scope call above. The two commits are on the branch for you to review, adjust, or revert.

mbradley:
sorry for the churn here. this should have been one changes-requested with a clean close-out.

the only thing actually blocking was the description not matching the pr (beyond removing worker/public, it also cleans up deploy docs and domains and refactors two test files). i acted on a note to fix the description myself and flipped a verdict that didn't need flipping. editing the description was all this needed, and that's where we landed anyway.

mbradley:
Closing this as superseded. #231 landed the same worker/public and stale-config cleanup on Aug 9, so main already has it — which is why this now conflicts across the board.

The one piece here that wasn't in #231 is the test seeder still defaulting to the production relay. I've pulled that into its own small PR: #243. Everything else here is already on main.

PR #243, opened 2026-08-21
fix(seed): require RELAY_URL so the test seeder can't default to production

mbradley:
## What this fixes

`scripts/seed-test-events.ts` loads fabricated test data (fake profiles, videos, and reports) for exercising the moderation tools. Run without `RELAY_URL` set, it defaulted to the production relay (`wss://relay.divine.video`), so an unconfigured run injects fake content and fake reports straight into the live moderation queue.

## The change

Require `RELAY_URL` explicitly and exit with a clear message when it is unset. No change in behavior when `RELAY_URL` is provided.

A second commit adds a regression guard, `scripts/no-production-endpoint-defaults.test.ts`. Without it nothing enforced the title's claim: `scripts/` is outside `tsconfig.app.json` and had no test, so the next script written with `process.env.RELAY_URL || 'wss://relay.divine.video'` would ship the same way this one did. The guard is inverted, like `worker/test/local-config-targets-local-stack.test.ts` (added by #233 for the same hazard in `wrangler.local.toml`): it finds every hardcoded endpoint a script would silently adopt and requires each to be local or explicitly excused, so a script added later has to be judged rather than inherit a default.

## What it deliberately does not change

The other five relay scripts already default to `ws://localhost:4444` and are untouched. A default carried on a command-line flag is out of the guard's scope: `bulk-delete-videos.mjs` defaults `--relay` to production because deleting production events is its purpose, and it cannot run without `--pubkeys`. The hazard being fixed is specifically the run that looks local and is not.

## Context

This is the one still-useful change from #242. The rest of #242 duplicated #231 (merged Aug 9), so #242 is being closed and this carries the fix forward on its own.

## Validation

`npm test` (`tsc -p tsconfig.app.json --noEmit`, `eslint`, `vitest run`, `vite build`) passes: 638 tests across 65 files. The guard was checked both ways — it fails, naming file, line, variable and value, when the production default is put back on line 15, and passes once it is removed. `RELAY_URL= npx tsx scripts/seed-test-events.ts` exits 1 and publishes nothing.

Closes #none

### DIFF
--- a/scripts/no-production-endpoint-defaults.test.ts
+++ b/scripts/no-production-endpoint-defaults.test.ts
@@ -1,0 +1,145 @@
+// A script under scripts/ must not fall back to a deployed endpoint when its
+// environment variable is unset.
+//
+// This is not hygiene. seed-test-events.ts publishes fabricated profiles,
+// videos and kind 1984 reports, and it read
+// `process.env.RELAY_URL || 'wss://relay.divine.video'`. Running it without
+// setting RELAY_URL — the documented usage at the top of the file said exactly
+// that — injected fake content and fake reports into the live moderation
+// queue. wrangler.local.toml had the same shape and the same consequence
+// (see worker/test/local-config-targets-local-stack.test.ts).
+//
+// The check is deliberately inverted, like the worker one: rather than listing
+// the scripts that must be safe, it finds every hardcoded endpoint that a
+// script would silently adopt and requires each to be local or explicitly
+// excused. A script added later fails this test until someone decides which it
+// is, instead of inheriting a production default unnoticed.
+//
+// Two shapes are covered, because both make an unset environment reach a
+// deployed host with no action from the operator:
+//
+//   const X_URL = process.env.X_URL || 'wss://relay.divine.video';
+//   const X_URL = 'wss://relay.divine.video';
+//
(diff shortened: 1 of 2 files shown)


hm21 (approved):
@mbradley The fix is right, and it is the only script that had the unsafe default — the other five already point at `ws://localhost:4444`. Verified `RELAY_URL= npx tsx scripts/seed-test-events.ts` exits 1 and publishes nothing.

One finding, pushed as `1f655fc` rather than handed back: nothing enforced the title. `scripts/` is outside `tsconfig.app.json` and had no test, so the next script written with `process.env.RELAY_URL || 'wss://relay.divine.video'` ships the same way this one did. `scripts/no-production-endpoint-defaults.test.ts` mirrors the inverted guard #233 added for `wrangler.local.toml`, over `scripts/` instead. I checked it both ways: it fails naming file, line, variable and value when the production default is put back on line 15, and passes once removed. Command-line-flag defaults are deliberately out of scope, so `bulk-delete-videos.mjs` stays green — reason is in the file header.

I also refreshed the PR description, since it said `scripts/` is not in the vitest gate and now it is.

Validation: full `npm test` green (638 tests, 65 files) plus CI. Nothing uncertain left on my side. @rabble the head moved.

PR #245, opened 2026-08-21
feat(minors): preserve protection when hosted accounts are replaced

NotThatKindOfDrLiz:
Protected-minor status currently disappears when its Keycast account row is deleted, so a legitimate delete-and-replace flow can silently remove safeguards from a minor. This change makes Relay Manager retain that classification independently of any replaceable hosted account.

Relay Manager now stores an opaque protected subject, its current and historical account bindings, provisioning operations, and retryable Keycast projection jobs. Onboarding records the case, subject, binding, and completed operation together after retry-safe Keycast provisioning. Authorized denial clears the durable subject first, then converges Keycast; failed projection calls remain queued for the deadline scheduler to retry.

Funnelcake receives dedicated bearer-only resolve and idempotent binding-close endpoints with the frozen determinate response shapes. Replacement provisioning persists and reuses the caller operation ID, rejects stale or cleared subjects, and remains disabled by configuration until Keycast #385 and the retention gate are ready. The registry does not copy parent contact, identity-review evidence, credentials, usernames, or display names.

This deliberately leaves ReportWatcher reading Keycast verified_minor as the enforcement projection. It also does not enable production replacement or resolve the Cloudflare Access service-token policy; rollout ordering and rollback constraints are documented.

Validation:
- npm run test (66 files, 692 passed, 1 skipped; type-check, lint, tests, and production build)
- cd worker && npx vitest run (22 files, 644 tests)
- cd worker && npm run test:d1 (7 files, 47 tests)
- cd worker && npx tsc -p tsconfig.json --noEmit

Closes #244
Refs divinevideo/support-trust-safety#203
Refs divinevideo/divine-funnelcake#1118
Refs divinevideo/keycast#385

### DIFF
--- a/docs/api-webhooks.md
+++ b/docs/api-webhooks.md
@@ -4,6 +4,8 @@
 
 The Divine Relay Manager exposes several API endpoints for integration with external systems like Zendesk (ticket report parsing, decision sync, age-review replies). Note: Zendesk does **not** execute moderation actions. Moderation is performed in Relay Manager; Zendesk only receives decision updates back. The former inbound moderation-execution path has been retired (see the note under `POST /api/zendesk/webhook`).
 
+Protected-minor deletion and replacement service endpoints are documented in [protected-minor-registry.md](protected-minor-registry.md). They use a dedicated bearer credential and are not authorized by moderator/browser credentials.
+
 **Base URL:** `https://api-relay-prod.divine.video` (production) or `https://api-relay-staging.divine.video` (staging)
 
 ## Authentication
--- a/docs/protected-minor-registry.md
+++ b/docs/protected-minor-registry.md
@@ -1,0 +1,26 @@
+# Protected-minor registry
+
+Relay Manager is the durable authority for whether a protected-minor subject is active or cleared. Keycast's `verified_minor` value is an enforcement projection associated with the current hosted account; deleting that account does not clear the subject.
+
+The registry deliberately stores only opaque subject and operation UUIDs, account pubkeys, age-review case provenance, lifecycle state, reasons, and timestamps. It does not copy parent contact, identity-review evidence, usernames, display names, credentials, or recovery material. Subject state and binding history are retained indefinitely until the retention review in support-trust-safety#204 defines a deletion period. Provisioning and projection-operation records are likewise retained so an old operation cannot create a second account after its idempotency record expires.
+
+## Service API
+
+All `/api/internal/protected-minors/*` routes accept only `Authorization: Bearer <PROTECTED_MINOR_SERVICE_TOKEN>`. Cloudflare Access browser assertions and `X-Admin-Key` do not authorize them. Infrastructure must separately allow the caller through Cloudflare Access using a service token or a path-scoped policy; the Worker bearer remains mandatory defense in depth.
+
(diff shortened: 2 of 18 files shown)


rabble (commented):
Reviewing rather than approving — I read the D1 migration, the worker configuration changes, and the deploy-affecting surface, but not the full ~800 lines of `protected-minors.ts`, `age-review.ts`, and the React components. The registry logic and its tests deserve a close read from someone with the minor-protection domain context.

**What I checked, and it is solid.**

The schema is genuinely defensive rather than nominally so. The paired `CHECK` on `protected_minor_subjects` makes a `cleared` row without `cleared_at` / `clear_reason` — and an `active` row carrying them — both unrepresentable, so a partially-written clear cannot be mistaken for a completed one. The pubkey `CHECK (length(pubkey) = 64 AND pubkey = lower(pubkey))` enforces canonical hex at the storage layer instead of trusting every call site. Most importantly, the two partial unique indexes are the right invariant:

```sql
CREATE UNIQUE INDEX ... ON protected_minor_account_bindings(subject_id) WHERE unbound_at IS NULL;
CREATE UNIQUE INDEX ... ON protected_minor_account_bindings(pubkey)     WHERE unbound_at IS NULL;
```

A subject can have at most one live binding and a pubkey can serve at most one subject, while full history is retained — which is exactly what "preserve protection when hosted accounts are replaced" requires. `deletion_attempt_id TEXT UNIQUE` gives replacement idempotency against a retried deletion.

The rollout posture is right too: `PROTECTED_MINOR_REPLACEMENT_ENABLED = "false"` in both `wrangler.prod.toml` and `wrangler.staging.toml`, so this lands dark and gets enabled deliberately, and the migration file is marked a parity copy with the real schema applied by `ensureSchema()`.

**One prerequisite before deploying.** Both wrangler configs add a `secrets_store_secrets` binding for `PROTECTED_MINOR_SERVICE_TOKEN` against store `ef490704b12a4feb91c7feb51229c364`. That secret has to exist in the store *before* the worker deploys — a binding to a missing secret fails the deploy outright, and it takes the whole worker down with it, not just this feature. Please confirm it is provisioned in both the staging and production stores first.

**Cross-repo consistency.** This is the server side of the call that divinevideo/divine-funnelcake#1120 makes, where the client-side credential is named `RELAY_MANAGER_SERVICE_TOKEN`. Different names on each side of one call is fine, but confirm both hold the *same* secret material — the asymmetry makes it easy to provision two unrelated values and only discover it when authentication fails in staging. Note also that nothing in IaC currently gives funnelcake-api `RELAY_MANAGER_URL` or `RELAY_MANAGER_SERVICE_TOKEN` at all, so that side is still unfinished.

dcadenas (approved):
🔒 Durable minor protection now survives account replacement and enforcement retries.

### Changes I pushed

- Enforced replacement ordering and operation-kind idempotency.
- Made denial and projection reconciliation durable.
- Tightened response validation, backfill ordering, and claimed-account UI state.

### Checks

- Frontend typecheck, lint, 692 tests passed with 1 skipped, and production build.
- Worker typecheck, 644 unit tests, and 47 D1 tests passed.
- Relay E2E was unavailable because the required local relay was not running.

